PR #445, opened 2021-05-20
Easier setup

ewels:
My attempt at getting this server to run locally without needing to set up any kind of config.

Hopefully, most things should now work with:

* `git clone ...`
* ~`docker-compose build` (or just `docker-compose run` twice)~ _(no longer needed)_
* `docker-compose run`

Then in theory you can overwrite all of the meaningless defaults with real values locally, mostly via environment variables, and it should run as a production server.

Still to do:

- [ ] Test that overwrite mechanism actually works
- [ ] Try to find places that break horribly without config, and make them break nicely
- [ ] Look into whether we can clean up some of the empty placeholder folders in the repo
- [x] _(Maybe)_ Think about getting the `docker-compose` file to wait for the database to be ready before launching the back end
- [ ] _(Maybe)_ Get Flask to log a functional URL instead of the one internal to the docker container
- [ ] _(Maybe)_ Get rid of config file mechanism and do _everything_ with environment vars instead
- [x] Document how to use an env vars file with the `docker-compose up` command instead of exporting everything.


README.md
```diff
@@ -1,58 +1,102 @@
...
-	```bash
-	docker-compose up
-	```
+For example:
 
-	* To use terminal after starting services, use the `-d` option.
+```bash
+SITE_NAME = "My Custom Data Delivery System"
+SECRET_KEY = "some-mega-random-string"
+SQLALCHEMY_DATABASE_URI = "mysql+pymysql://johannes:test_password@db/DeliverySystem"
```
alneberg, line 90: You could chose a different username if you want to? 😆

alneberg: Well, it wasn't exactly sensitive information... Great!

senthil10 (changes requested):
Thank you again Phil, looks really good. Like I mentioned, lets not remove the database docker file yet, as I am looking into extending it with other SQL related stuff. Can remove later if what I am looking into din't workout

ewels:
Copying over relevant bits from Slack discussion:

https://hub.docker.com/_/mariadb

> ### Initializing a fresh instance
> When a container is started for the first time, a new database with the specified name will be created and initialized with the provided configuration variables. Furthermore, it will execute files with extensions `.sh`, `.sql`, `.sql.gz`, and `.sql.xz` that are found in `/docker-entrypoint-initdb.d`. Files will be executed in alphabetical order. `.sh` files without file execute permission are sourced rather than executed. You can easily populate your `mariadb` services by mounting a SQL dump into that directory and provide custom images with contributed data. SQL files will be imported by default to the database specified by the `MARIADB_DATABASE` / `MYSQL_DATABASE` variable.

A directory of these files could just be mounted as a volume in the correct place. eg. from [this Stack Overflow post](https://stackoverflow.com/a/64471585/713980):

```yaml
services:
  db:
    image: mariadb:latest
    volumes:
       - "./sql-scripts:/docker-entrypoint-initdb.d"
```

So maybe this is a nicer way to do database initialisations instead of having a custom `Dockerfile` for the database...?

talavis:
> Hopefully, most things should now work with:
> 
> * `git clone ...`
> * `docker-compose build` (or just `docker-compose run` twice)
> * `docker-compose run`

`docker-compose up` should merge the final two steps into one, right?

README.md
```diff
@@ -1,58 +1,102 @@
-# Data Delivery System -- **WIP**
-A a single cloud-based system for all SciLifeLab facilities, where data generated throughout each project can be delivered to the research groups in a fast, secure and simple way. 
+# SciLifeLab Data Centre - Data Delivery System
 
-**_This project is supported by EIT Digital, activity number 19390. This deliverable consists of design document and implementation report of application and validation of VEIL.AI technology in SciLifeLab context in Sweden._** 
+**A a single cloud-based system for all SciLifeLab facilities, where data generated throughout each project can be delivered to the research groups in a fast, secure and simple way.**
```
talavis, line 3:
```suggestion
**A single cloud-based system for all SciLifeLab facilities, where data generated throughout each project can be delivered to the research groups in a fast, secure and simple way.**
```

ewels:
> `docker-compose up` should merge the final two steps into one, right?

Right, yes - at the time of writing that didn't work because the second container would crash due to the SQL database not being ready. But since then I added the `healthcheck` stuff and now it waits like a good little container and `docker-compose up` is sufficient 👍🏻

.env
```diff
@@ -0,0 +1,17 @@
+##################################################################
+# This file is used by `docker-compose` and sets default values.
+# You can override any of these in your shell, for example by
+# adding export statements to your .bashrc:
+#    export DDS_CLI_ENV="production"
+##################################################################
+
```
senthil10, line 7: Can you also add `FLASK_ENV="development"`, since this file is intended for development

docker-compose.yml
```diff
@@ -0,0 +1,55 @@
...
+      retries: 5
+
+  backend:
+    container_name: dds_backend
+    image: dds-backend-local:latest
+    build: ./
+    command: python3 app.py
+    environment:
+      - DDS_APP_CONFIG
+      - FLASK_ENV
```
senthil10, line 30: Here can you also add `DDS_CLI_ENV`, atm its needed

senthil10 (commented):
Ok, I have tested this locally and as far as I can tell the basics works as it is without any (modification) which is suffice for frontend development

senthil10 (approved):

.env
ewels: This is exposed in the `docker-compose.yml` file. But yeah I agree that it would be better to put it here and set a default.

docker-compose.yml
ewels: That's already in the `.env` file, set as `development` by default.

ewels:
So all of the required environment variables were there already, but they were split between the `.env` file and the `docker-compose.yml` file. I don't think that there's any reason that these can't all be in the `.env` file so I've moved them there for clarity.

PR #446, opened 2021-05-21
Front end package manager

ewels:
This PR adds a front-end build system using `npm`.

Instead of loading CSS and Javascript assets from online CDN servers, assets are now fetched and built using `npm` and served locally. This has a few advantages:

* We can use SASS / SCSS instead of pure CSS, allowing us to customise variables for Bootstrap instead of having to overwrite styles
* Better stability - production server doesn't rely on anyone else
* We can compile and minify assets to make them more efficient
* People can run the server behind a firewall if they want to (eg. offline)

To do this, I have set up the project to use `npm` just for the front end assets, in `dds_web/static`. The production Dockerfile now uses a multi-stage build, running `npm` to compile the requires CSS and then copying the finished file into the second stage production image.

For those running locally, you need to use `npm` locally (for now anyway, unless anyone can figure out a better way). Make sure that you have NodeJS installed and `npm` is vaguely up to date, then cd into `dds_web/static` and run `npm install` to fetch all required packages. Then run `npm run css` to compile the production CSS file. The `docker-compose` setup mounts this local folder so then the running server should pick up the files.

If you're working on the SASS files, you'll want to be able to test the updated compiled CSS as you go. To do this, run `npm run watch` in a separate terminal - that will run and watch for changes and recompile on changes. As the volume is mounted they should appear in the browser immediately (remember to force-refresh).

The other major change here is to update from Bootstrap 4 to Bootstrap 5. This breaks a load of stuff. That combined with the move away from the `styles.css` and greater use of helper bootstrap classes means that currently everything on the front-end is broken. This PR is WIP whilst I start work on fixing up the front end templates. Feel free to test + review the build system in the mean time to make sure that you are happy with the code and approach 👍🏻


Dockerfile
```diff
@@ -1,9 +1,27 @@
+##############################
+## Compile dist CSS
+##############################
+FROM node:12-alpine AS compile_css
```
talavis, line 4: Maybe use newer Node? Node 12 has EOL on 30 Apr 2022 for security updates (bug fixes ended on 20 Oct 2020).

ewels: Yeah I had 16 originally but that wouldn't run. Downgrading fixed it. Can maybe downgrade a little less aggressively though.

ewels: Ok, checked my local and I have version 15. Updated the Dockerfile to 15 and that worked fine. Also dropped the `apk` install line below which was part of my trial-and-error approach to fixing the problem and turns out to not be needed..

ewels:
@alneberg reminded me that I haven't yet looked at the interface for uploaded files. Will look into that asap - happy to do as a follow-on PR if you want to merge before I get there.

dds_web/templates/base.html
```diff
@@ -3,69 +3,28 @@
...
-        <!-- Custom JS functions -->
+
+        <!-- Javascript -->
+        <script src="{{ url_for('static', filename='node_modules/jquery/dist/jquery.min.js') }}"></script>
+        <script src="{{ url_for('static', filename='node_modules/bootstrap/dist/js/bootstrap.bundle.min.js') }}"></script>
+        <script src="{{ url_for('static', filename='node_modules/datatables.net/js/jquery.dataTables.min.js') }}"></script>
+        <!-- TODO - bs5 version not yet on npm. When it is, install as an npm package. -->
+        <script src="https://cdn.datatables.net/1.10.24/js/dataTables.bootstrap5.min.js"></script>
+        <!-- We could use FontAwesome through npm, but then we need configure secrets to access the Pro kit which is a hassle -->
+        <script src="https://kit.fontawesome.com/8cd1f5d753.js" crossorigin="anonymous"></script>
```
alneberg, line 26: When we added fontawesome to genstat I think we wanted to keep this URL secret?

senthil10 (approved):
Tested and works for me, I did not test extensively but it shouldn't be a problem as basic features work. Any small bugs/fixes can be made along the way. 

P.S: I believe from here the frontend will only get prettier 😄

alneberg (approved):
Looks good to me!

senthil10:
> @alneberg reminded me that I haven't yet looked at the interface for uploaded files. Will look into that asap - happy to do as a follow-on PR if you want to merge before I get there.
> 
> I should also check to see if I can strip out any now-unused javascript.

I would suggest we can merge this PR first, and make separate PR for making things better ^^

dds_web/templates/base.html
ewels: Yeah I did wonder that at first. At the moment it's unlocked so anyone else can use it but we can lock it to a specific domain in the future.

Or if we're worried I can regenerate the ID and we can stick it in a secret. But then that's an additional hassle for anyone wanting to develop the site.

ewels:
let's do this

PR #448, opened 2021-05-25
Changes in database structure

inaod568:
_This will probably break stuff, so @senthil10 can merge and fix the web stuff, I will not have time to look into that before the end of the week at the earliest._

Instead of having a `Roles` table and separate tables for Facility users and research users, the structure is changed to:

- `Facility`: Contains contact information and (for now) the Safespring etc info for each facility. Has a column containing all accounts connected to the specific facility, and projects.
- `project_users`:  Association table to allow for many-many relationship between users and projects. 
- `Project`: 
    - `facility_id` points to the responsible facility, not account but `Facility` table row. 
    - `files`: All files uploaded within the project.
    - `owner` is removed.
- `User`: All DDS users. 
    - `facility_id`: If facility account - linked to `Facility` table. 
    - `permissions`: Column which will regulate which actions the users are allowed to perform in the system. Format: `aglpr` = `add/admin-get-list-put-remove`. The letters are replaced with `-` if the user does not have permission for a specific action. 
    - `projects`: All projects which the account has access to. Currently same permissions for all projects for a specific account.
   
Have changed the rows affected in the `dds_web/api` code for CLI use, so that should work. 

Have **not** changed the `dds_web/dds_web` code - the code including the `Role` table needs to be changed, and some column names may have changed slightly. Make sure to change from `Project.id` to `Project.public_id` in certain queries.


dds_web/database/models.py
```diff
@@ -73,38 +20,25 @@ class Facility(db.Model):
...
-
-    # Table setup
-    __tablename__ = "roles"
-    __table_args__ = {"extend_existing": True}
-
-    # Columns
-    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
-    username = db.Column(db.String(20), unique=True, nullable=False)
-    facility = db.Column(db.Boolean, unique=False, nullable=False)
+project_users = db.Table(
```
senthil10, line 37: Would it be better to name it `user_projects` ?

dds_web/database/models.py
```diff
@@ -115,7 +49,8 @@ class Project(db.Model):
     __table_args__ = {"extend_existing": True}
 
     # Columns
-    id = db.Column(db.String(32), primary_key=True)
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    public_id = db.Column(db.String(32), unique=True, nullable=False)
```
senthil10, line 53: Just occurred, since we will using this field will mostly used elsewhere (for queries and etc) could we name this column as `id` and change the primary key column to `db_id` or `secret_id` or something

dds_web/development/db_init.py
```diff
@@ -78,309 +78,136 @@
...
+            public_id="public_facility_id",
+            username="facility",
+            password="$argon2id$v=19$m=102400,t=2,p=8$0jcemW3Ln+HTPUt/E3xtKQ$aZGqrrBBU5gq5TbWYwUWD62UiQUmTksbKOkmbMJzdhs",
+            role="facility",
+            permissions="--lpr",
+            facility_id=facilities[0],
+        ),
     ]
 
+    # User(
```
senthil10, line 145: Can remove this ? not a important one, just feel like we never use these old codes later (even for ref)

senthil10 (commented):
Looks good to me, just few thoughts. Let me know if you wanna fix it or I could do it too, when I fix the web part ^^

dds_web/api/db_connector.py
```diff
@@ -84,9 +96,15 @@ def project_size(self):
 
         num_proj_files, error = (0, "")
         try:
+            current_project = models.Project.query.filter(
+                models.Project.public_id == func.binary(self.project["id"])
+            ).first()
```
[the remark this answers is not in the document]
inaod568: Yes, but I'm looking into a better way - not getting the current project first, just using `join`. So this was just temporary, so that we have the db structure and then I can fix a better way.

dds_web/database/models.py
inaod568: I mean, we could technically, but since it's a many-to-many table, and the `User` table has a `relationship` and not a column with foreign key, it's more logical I feel to have it as `project_users`. The original idea was to have an association table to keep track of which users are allowed to access which projects. `user_projects` is basically `projects` relationship in the `User` table already.

dds_web/database/models.py
inaod568: I think it's like a standard to have `id` as the primary key column? We don't have to, but personally it feels like this could cause even more confusion, but we can discuss it if you want.

dds_web/development/db_init.py
inaod568: Yes, will remove it.

dds_web/database/models.py
senthil10: Its fine, I was just talking in terms of semantics 👍

inaod568: Me too:

- `User.projects` : The projects a specific user has access to
- `project_users`: Which users have access to specific projects

dds_web/database/models.py
senthil10: Yes, just thinking as DB entity `id` is more suitable for a primary key column. But while working for frontend, during display or to fetch some info using `id` (real project id which is displayed) I first instinctively write `id` then will change to `public_id`. That is why I thought could just have `id` for `id`, maybe `key` or something else for primary key

PR #451, opened 2021-05-27
Start fixing up / rewriting project file tree

ewels:
Follow on from https://github.com/ScilifelabDataCentre/dds_web/pull/446 where I broke the rendering of the project file tree.

Now looks ok again, but seems to mostly download empty zip files which is probably not the intended behaviour 🤔 😅


ewels:
Ah the empty zip files are probably because I haven't set it up with safespring. So probably could do with better checking / error messages but I suspect that this is no worse than the previous behaviour. Needs more testing.

senthil10 (approved):

PR #455, opened 2021-05-28
Rewrite web uploader

ewels:
Work in progress: rewriting the web file upload interface. Dropping the requirement for a plugin and going for a more customised / native feel with additional features.

Rough wish list of features:

* [x] Drag + drop files and folders onto a drop zone, instead of having 2x file inputs
* [x] List files + paths to be uploaded before doing so
* [x] Progress bars in page, not in modal
* [ ] Independent file uploads with their own progress + error messages
* [x] Display file sizes before upload
* [ ] Display nice units of file size uploaded / total on progress bar (in addition to percentage)
* [x] Auto-detection of files that are too large prior to upload
* [ ] Option to retry failed uploads / (clear successful uploads?)
* [x] Confirm box requirement when navigating away from the page during active uploads
* [x] Refresh for file listing / addition of prompt to reload page


ewels:
Ok, I think that this PR has gotten big enough. As far as I can tell it is now functional (more testing welcome!) and hopefully an improvement over the existing functionality. I think now is a good time to stop and put the rest of the to-do list into an issue for future PRs.

senthil10 (approved):
From quick review and testing, looks good to me. Just noticed few things

- [ ] As soon as new project is created it is listed in the project list, so I click the project id to go to the project page, then when I go back (via browser back), the list doesn't contain the newly added project, we have to reload it again then its fine. Guess have to do with cache
- [x] For the very first upload, the file list is not updated (had to refresh), maybe cause the file list doesn't exist 
- [ ] Maybe later we could add remove option for the dropped files (if the user had change of mind on certain files)

All are not crucial stuff, will leave it to you if you wanna fix and merge or merge and fix later :)

ewels:
Good spot - will fix tonight 👍🏻

ewels:
> Maybe later we could add remove option for the dropped files (if the user had change of mind on certain files)

Yup - this along with with independent file uploads, independent file size warnings etc. Definitely room for improvement along these lines. Adding a remove button wouldn't be too tricky but I wonder if we wait for a proper refactor to add all of this stuff together.. 🤔

ewels:
Ok @senthil10, I fixed the first-file-upload thing 👍🏻 

> As soon as new project is created it is listed in the project list, so I click the project id to go to the project page, then when I go back (via browser back), the list doesn't contain the newly added project, we have to reload it again then its fine. Guess have to do with cache

This is not due to code in this PR. And to be honest, it's not trivial to solve. As it's a bit of an edge case, I'm tempted to ignore for now.. That ok? Can make a low-priority issue for it if you want.

dds_web/templates/project/project.html
```diff
@@ -111,5 +109,209 @@ <h2 class="mb-4">Project Files</h2>
...
+            // Human readable file sizes
+            function humanFileSize(size) {
+                if(size == 0){ return '0 B'; }
+                var i = Math.floor( Math.log(size) / Math.log(1000) );
+                return ( size / Math.pow(1000, i) ).toFixed(2) * 1 + ' ' + ['B', 'KB', 'MB', 'GB', 'TB'][i];
+            };
+            // This is a hack, and it pains me to write it. Needs refactoring.
+            for(path in upload_files){
+                let filesize = upload_files[path].size;
+                console.log(upload_files[path]);
```
alneberg, line 217: Do you want to keep this one?

ewels: Already gone :)

ewels:
Ok, added another fairly major improvement - interface now shows you how large files are before uploading and skips them when the upload runs. Previously you had to wait for the huge file to upload before you got an error that it was too big, so this is kind of needed 😄

ewels:
Updated to also display + check the _total_ combined filesize. Just throws an alert if you try to upload with too much and doesn't let you. Not the nicest interface as the only way around for now is to reload the page and try again. But at least it now tells you before you spend hours waiting for the upload to complete.

I think that this is as far as I want to take this page before a rewrite in Vue etc. I've said that before but I am starting to mean it now.. 😆

senthil10:
Great, Tack Phil :)

PR #456, opened 2021-05-31
Clean up some files in the root

ewels:
Noticed that `.gitmodules` was now empty and can be deleted. I also moved the `s3_config_example.json` file into the `run_dir/sensitive` folder, as this is where the proper `s3_config.json` file should go anyway. Moving it here also means that the `.empty` file there is no longer needed.

Not sure if `.pylincrc` is used for anything / needed? Could maybe delete that too if not.


senthil10:
Yes, I too couldn't find where the `.pylincrc`is used, @inaod568 can we remove it ?

ewels:
Ok I zapped it 👀 💥

PR #459, opened 2021-06-03
Web upload - refactor staging area display.

ewels:
Avoid async problem with displaying file list. Fixes bug where very large numbers of files would not show up.

Adding the `node_modules` directory now shows the correct number of files:


ewels:
(ok to merge this PR if we want, as the fix is complete - just needs a second fix in addition to this one).

senthil10:
I will leave the merging to @alneberg, since he knows the issue more

alneberg (approved):
Looks good! I created https://github.com/ScilifelabDataCentre/dds_web/issues/460 to track the problem that not all files from subdirectories are added. We'll deal with that later.

PR #468, opened 2021-06-08
Use Node 14 for the ccs builder

talavis:
Node 15 reached end of life in May, and is thus no longer available on Dockerhub. Use Node 14 (LTS release) instead.

If you already have `node_modules` from Node 15, the folder must be deleted before you can use this version.


inaod568:
> Node 15 reached end of life in May, and is thus no longer available on Dockerhub. Use Node 14 (LTS release) instead.
> 
> If you already have `node_modules` from Node 15, the folder must be deleted before you can use this version.

Where would that folder be in that case? The repo root?

PR #471, opened 2021-06-09
Homepage redesign

ewels:
First stab at making the homepage look a little nicer...

* Split the log-in form into a component
    * Means it can be imported into both the homepage and the Log in page.
* Restyled the homepage to have a 2-column layout
    * Login form on the left under a lead sentence (if not logged in).
* Added a new "hero" splash image
* Did some fancy CSS animations of certain SVG elements

The image / animation idea comes courtesy of Johan at SciLifeLab, originally done for the DDLS SciLifeLab podcast: https://youtu.be/8pL3_ZGJUzE?t=175 - he sent me the original vector file. I re-coloured it as he had done previously (he didn't have a version saved with the nice colours) and then exported as a SVG. Once exported, I messed around with CSS transformation animations to re-create the animations.

I'd still like to do the blobs, but after playing around with it for a while it's getting too late. So putting in the PR with the current state.

More to come later.

https://user-images.githubusercontent.com/465550/121439800-f188b100-c986-11eb-8fbc-0ca072c4d098.mov


ewels:
Note to self - hoping to get the blobs animated using this tutorial: https://css-irl.info/fun-with-css-motion-path/

inaod568 (approved):
This looks really nice!

inaod568:
Btw, the animation doesn't work in Firefox. Still looks nice though.

ewels:
I wrote this last night, but apparently didn't click "Comment":

---

> we can’t scale our SVG and have the path match. If we try and change the SVG width then the path remains at its original size.

So close to getting the blobs to work, but alas 😭 The animation path has to be able to resize with the image, without that it can't happen.

There's a proof of concept here, but it's pretty hacky: https://css-tricks.com/create-a-responsive-css-motion-path-sure-we-can/

I think that I will have to abandon the blobs for now 😢

ewels:
But... I just managed to get it to work! PR incoming..

PR #473, opened 2021-06-10
OIDC support for logins

talavis:
Add support for login via OpenID Connect.

Adds a new table `Identifiers` with a mapping of `user_id` to `identifier`, intended to replace `User.username`.

On login, the system will compare the email from oidc to all identifiers in the table and log in as the corresponding user.

Adds the following configuration options:
```
OIDC_CLIENT_ID = ""
OIDC_CLIENT_SECRET = ""
OIDC_ACCESS_TOKEN_URL = ""
```

### DIFF
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -62,11 +62,13 @@ def create_app():
     # ma.init_app(app)
 
     # initialize OIDC
-    oauth.register("default_login",
-                   client_secret=app.config.get("OIDC_CLIENT_SECRET"),
-                   client_id=app.config.get("OIDC_CLIENT_ID"),
-                   server_metadata_url=app.config.get("OIDC_ACCESS_TOKEN_URL"),
-                   client_kwargs={"scope": "openid profile email"})
+    oauth.register(
+        "default_login",
+        client_secret=app.config.get("OIDC_CLIENT_SECRET"),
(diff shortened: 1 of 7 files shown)


inaod568 (approved):

dds_web/__init__.py
```diff
@@ -59,6 +61,13 @@ def create_app():
     db.init_app(app)  # Initialize database
     # ma.init_app(app)
 
+    # initialize OIDC
+    oauth.register("default_login",
```
ewels, line 65: Can you run `black -l 100` quickly please? CI isn't added to this repo yet, but will be soon.. (same setup as `dds_cli`)

dds_web/user.py
```diff
@@ -59,6 +61,68 @@ def login():
         return redirect(to_go_url)
 
 
+def do_login(session, identifier:str, password:str = "") -> bool:
+    """
+    Check if a user with matching identifier exists. If so, log in as that user.
+
+    TODO:
+      * Add support for passwords
```
ewels, line 69: How do you mean?

ewels (approved):
LGTM! I haven't spent much time working with the DB back-end and stuff though, so my review isn't worth very much.

dds_web/user.py
talavis: I intend that function to be the way to log in, but I didn't spend the time to make it compatible with the existing password login

PR #475, opened 2021-06-10
Animated data blobs :)

ewels:
Finally managed to get the little animated data blobs going for maximum coolness...

https://user-images.githubusercontent.com/465550/121584222-8942d980-ca31-11eb-8026-cdce71d042e6.mov

Blobs aren't as nice as the originals, but good enough I think. Pretty happy with this now!

@inaod568 - yes, only tested in Chrome so far. Does it degrade nicely in Firefox? As in, just no animation? I can try to look into improving compatibility..


ewels:
Continuation of https://github.com/ScilifelabDataCentre/dds_web/pull/471

ewels:
Just checked on MacOS and the animations work fine for me in Chrome `90.0.4430.212` and Firefox `89.0` (also an older version before I hit the update button). In Safari `14.1.1` the cogs spin and cloud pulses, but the blobs do not move. They do fade in and out. It looks fine though.

We're already using an auto-prefixer for maximum compatibility, so I'm not sure that there's much more that we can do (easily).

inaod568:
I like it! I will try again with Firefox. It looked fine though, nothing moving but otherwise it looked the same.

inaod568:
Yes, animations shown in Firefox as well 👏🏻

inaod568 (approved):

PR #498, opened 2021-06-14
Automated linting of Python

ErikDanielsson:
Added github action for automatic linting of Python code with black.


ewels (approved):
This CI test will immediately fail when merged, but we can take that in a follow-up PR I guess - nice to check that it works 😅

Issue #506, opened 2021-06-15
`size` and `size_encrypted` to `size_original` and `size_in_system`


inaod568:
Clearer syntax. Will both make it easier to add a non-encryption option, and simplify the invoicing calls.

inaod568:
Fixed in PR #507

PR #507, opened 2021-06-15
size columns in db changed

inaod568:
`File.size` --> `File.size_original`
`File.size_encrypted` --> `File.size_stored`

Issue #506 

Have tried to find all references to the file size, if I missed one it will probably be in the web interface.


ewels (approved):
LGTM 👍🏻

PR #508, opened 2021-06-15
Dark tweaks

ewels:
Dark mode fixes:

* Transparent shadows on the hero image, so no white circles in dark mode
* Automatically uses dark or light mode, depending on OS / browser settings (until the switcher toggle is used)

Don't totally love how this is set up, code could perhaps do with a little tidying (eg. moving some logic out of the template and into the Python code, then just using variable names in the template). But I think it's good enough and can always come back and polish it later if we feel the need.

Other tweaks:

* Removed 'Register' button from the nav bar
    * Didn't do anything, and I suspect that we will not have a registration feature ever (other than logging in via OIDC). Can always add back later if we need it.
* Fix the text colour in dark mode for the "Log In" button in the navbar
* Added label to dark mode toggle button
* Layout tweak for login form


ewels:
If you've tested already, you'll need to clear cookies to try this new automatic selection.

inaod568 (approved):

PR #511, opened 2021-06-16
Use offical docker actions instead (also fix build)

talavis:
- Use the official Github Actions to build the images
- (Hopefully) fix the Docker builds


ewels (approved):
Looks good! Are we going to start doing versioned releases as well? In which case it might be nice to have a tagged build for each release too..

inaod568:
We should start doing versioned releases, yes. Should that be fixed in this PR or shall I merge it?

talavis:
I added a workflow for building images when releases are `created`. Not sure if I'm using the variable for the tag correctly, would be great if e.g. @ewels could check.

inaod568 (approved):

PR #513, opened 2021-06-17
Invoicing setup

inaod568:
- Added `Version` table which is updated each time a file is added to tile `File` table
- Removed the timestamps for when the File rows are added and deleted, instead this is saved in the `Version` table.
- If a file is overwritten, the corresponding row in `Version` is set as deleted (with a timestamp `time_deleted`) and a new row is added.
- If a file is deleted, the corresponding row in `Version` is set as deleted as when overwritten, and the `active_file` column in `Version` is automatically set to NULL when the file row no longer exists.

This breaks some db calls, but I think I've found the affected parts and fixed them. 

_Update June 24th:_  Refactoring of the project listing endpoint, and also added a Backgroundscheduler which will calculate the facility usage and cost every day. This is a WIP though, not done yet.


dds_web/utils.py
```diff
@@ -58,3 +70,184 @@ def format_byte_size(size):
...
+    old_file = parent_dir / pathlib.Path("development/invoicing/safespring_invoicespec.csv")
+
+    current_time = timestamp(ts_format="%Y-%m-%d_%H-%M-%S")
+    to_file = old_file
+    # to_file = parent_dir / pathlib.Path(f"development/invoicing/{current_time}.csv") # TODO (ina): uncomment later
+
+    # shutil.copy(old_file, to_file)    # TODO (ina): uncomment later
+
+    # Get data
+    csv_contents = pandas.read_csv(to_file, sep=";", header=1)
```
senthil10, line 95: If `pandas` is only used to read the csv file, its better to look into [csv](https://docs.python.org/3/library/csv.html) package. It light weight compared to pandas

senthil10 (approved):
Great work! Looks good to me, had only very minor comments. Could merge it if you wanna fix it later :)

dds_web/utils.py
inaod568: It's also to get specific rows, easier with `pandas`. But I will look into changing to `csv` instead.

PR #523, opened 2021-07-05
Profile page

pericsson:
This PR includes the addition of a profile page that lists user permissions and first name and last name, located at `<domain>/user/account`.

Updates:
* Added a field to the User table in the db for first name and last name where nullable is set to `True`.
* Added a table for the email addresses associated with a user that also has a column for whether it is primary or not.
* Added a function for getting all email addresses to a user.
* Added a tab in the navbar named as the username that routes to the profile page. The previous route with this name was renamed to `Dashboard` if you are an admin and otherwise to `projects`.

This page was built using Vue during development of this page but the vue part was removed from this PR. Also placeholders for buttons to edit name, add email, change password and deleting an email address were removed from this PR. The code before removing these parts from the PR can be found on the git branch [profile_account_vue](https://github.com/pericsson/dds_web/tree/profile_account_vue). The strikethrough text below is from work on the `profile_account_vue` branch.

~~All buttons are placeholders and not functioning at the moment. The buttons `Change password` and `Add New Email Address` should maybe not be included in this PR since I think it might need some extra layer of verification for example with an email being sent to the user to confirm an email address before it is added.~~

~~## TO DO~~
~~Modifications to Vue:~~
~~- [ ] move the large code chunk in `account.js` into smaller parts so that the `account.html` template is accessing mostly `data()` elements from vue instead of having the whole page in the vue html tag.~~
~~- [ ] Tooltips broke when moving the code to vue, try to fix this (maybe it is easier to fix when the large code chunk is mostly moved to the html).~~
~~- [ ]  Change vue from dev to prod.~~

~~Still have to do functions for:~~
~~- [ ] changing first name and last name~~
~~- Remove console log response used during testing~~
~~- [ ] removing email addresses~~
~~- [ ] select primary email address~~
~~- Add confirm for change or primary email~~
~~- [ ] Fix response container (currently commented out from `account.js`).~~
~~- [ ] change password and add email~~
~~- As mentioned above I suppose it would need another layer of verification though an email and verification code or so~~


inaod568 (approved):
LGTM! 👍🏻

PR #540, opened 2021-08-09
Db idchanges andinit

inaod568:
The `User.username` column is now the primary key and the `id` is removed. Have changed the project listing and  project page accordingly but keep this in mind if something does not work -- very possible that something needs to change from id to username etc.

Have also removed the hard coded argon2id passwords and set them as called functions in the db initiation.

### DIFF
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -73,8 +73,10 @@ class AuthenticateUser(flask_restful.Resource):
         try:
             user_ok = dds_auth.verify_user_pass(username=auth.username, password=auth.password)
         except sqlalchemy.exc.SQLAlchemyError as sqlerr:
+            app.logger.exception(sqlerr)
             return flask.make_response(str(sqlerr), 500)
         except exceptions.AuthenticationError as autherr:
+            app.logger.exception(autherr)
             return flask.make_response(str(autherr), 401)
 
         # User credentials incorrect
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -62,14 +62,11 @@ class AuthenticateUser(flask_restful.Resource):
         # Get username and password from CLI request
         auth = flask.request.authorization
         if not auth or not auth.username or not auth.password:
-            return flask.make_response("Could not verify", 401)
+            return flask.make_response("User credentials missing.", 401)
 
         # Project not required, will be checked for future operations
         args = flask.request.args
-        if "project" not in args:
-            project = None
-        else:
-            project = args["project"]
+        project = args.get("project")
 
         # Check if user in db
         try:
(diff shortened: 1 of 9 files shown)


dds_web/crypt/auth.py
```diff
@@ -2,10 +2,39 @@
...
+    except sqlalchemy.exc.SQLAlchemyError:
+        raise
+
+    # User does not exist
+    if not user:
+        raise exceptions.AuthenticationError("Incorrect username and/or password!")
+
+    # Verify password and generate token
+    if verify_password_argon2id(user.password, password):
+        if "l" not in list(user.permissions):
```
worukan, line 29: Does this (if block for permissions) belong here? Permission is related to authorization. If this code piece is related to verifying the authenticity of a user, then by single responsibility principle it would be a better to move the authorization check to elsewhere.

inaod568: No, it doesn't. I had actually written a comment about that but I seem to have removed it again... will comment it out for now and move it somewhere else later.

inaod568: Fixed.

dds_web/api/user.py
```diff
@@ -68,35 +69,27 @@ def get(self):
...
+            return flask.make_response(str(autherr), 401)
 
-        # Verify user password and generate token
-        if verify_password_argon2id(user.password, auth.password):
-            if "l" not in list(user.permissions):
-                return flask.make_response(
-                    f"The user '{auth.username}' does not have any permissions", 401
-                )
+        # User credentials incorrect
+        if not user_ok:
```
worukan, line 83: I think this part is already covered in auth.py on lines 23-24 so expect that it will be caught on line 77.

inaod568: At the moment `verify_user_pass` raises the exception if the user doesn't exist but returns `False` if the password is incorrect. So at the moment the `if not user_ok` is needed, but I was thinking about changing

```
# Verify password and generate token
    if verify_password_argon2id(user.password, password):
        # Password correct
        return True

    # Password incorrect
    return False
```
to
```
if not verify_passwork_argon2id(user.password, password): 
    raise AuthenticationError("Incorrect username and/or password!")

return True
```
And in that case, I could remove the `if not user_ok` part -- should I do that?

worukan: perhaps we could write it as follows

if user and verify_password_argon2id(user.password, password):
    # password correct
    return True

# if not user will come here and incorrect password as well, so we either return true or throw an exception
raise AuthenticationError("Incorrect username and/or password!")

inaod568: Sure! I actually did 
```
    # User does not exist or password does not match
    if not user or not verify_password_argon2id(user.password, password):
        raise exceptions.AuthenticationError("Incorrect username and/or password!")

    # Password correct
    return True
```

But what you wrote is better. Will switch to that.

inaod568: Will merge this though and the change will be in the next PR together with logging.

PR #543, opened 2021-08-10
Revert "Revert "Db idchanges andinit""

inaod568:
Reverts ScilifelabDataCentre/dds_web#542

Shouldn't have merged the first PR so had to revert and then redo the PR.

### DIFF
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -23,6 +23,8 @@ from dds_web import app, timestamp
 from dds_web.database import models
 from dds_web.crypt.auth import gen_argon2hash, verify_password_argon2id
 from dds_web.api.dds_decorators import token_required
+from dds_web import exceptions
+from dds_web.crypt import auth as dds_auth
 
 ###############################################################################
 # FUNCTIONS ####################################################### FUNCTIONS #
@@ -32,7 +34,6 @@ from dds_web.api.dds_decorators import token_required
 def jwt_token(username, project_id, project_access=False, permission="ls"):
     """Generates and encodes a JWT token."""
 
-    token, error = (None, "")
     try:
         token = jwt.encode(
             {
@@ -42,10 +43,10 @@ def jwt_token(username, project_id, project_access=False, permission="ls"):
             },
             app.config["SECRET_KEY"],
         )
-    except Exception as err:
-        token, error = (None, str(err))
-
-    return token, error
+    except Exception:
+        raise
+    else:
+        return token
 
 
 ###############################################################################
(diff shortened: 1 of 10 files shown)


dds_web/api/user.py
```diff
@@ -62,46 +63,35 @@ def get(self):
...
 
-        # Verify user password and generate token
-        if verify_password_argon2id(user.password, auth.password):
-            if "l" not in list(user.permissions):
-                return flask.make_response(
-                    f"The user '{auth.username}' does not have any permissions", 401
-                )
+        # User credentials incorrect
+        if not user_ok:
+            return flask.make_response("Incorrect username and/or password!", 401)
```
inaod568, line 84: Changed to 
```
if user and verify_password_argon2id(user.password, password):
    # password correct
    return True
```
but not in this PR/branch.

dds_web/crypt/auth.py
```diff
@@ -2,10 +2,86 @@
...
+        #     )
+
+        # Password correct
+        return True
+
+    # Password incorrect
+    return False
+
+
+def user_session_info(username):
```
worukan, line 41: Why do we need to load the following information in the session? I think the username should already be kept in the authenticated session. The other information such as is_admin and is_facility can be checked on a need to check basis. The way we keep it in session causes extra work and can cause inconsistency issues as the data needs to be updated in several places such as both db and session. I recommend keeping the session variables to a minimum and only when really needed.

dds_web/crypt/auth.py
```diff
@@ -61,6 +137,7 @@ def validate_user_credentials(username, password):
```
worukan, line 137: As a reminder, this function is no longer in use and will be removed.

dds_web/database/db_utils.py
```diff
@@ -16,10 +18,19 @@ def get_facility_column_by_username(fname, column) -> (str):
...
+    # Get all projects connected to facility
+    try:
+        project_list = models.Project.query.filter_by(facility_id=facility_id).all()
+    except sqlalchemy.exc.SQLAlchemyError:
+        raise
+
+    if not project_list:
+        project_list = []
+
+    return project_list
```
worukan, line 33: Not a big deal but you could write something like
`return project_list if project_list else []`
instead of writing a conditional first and then a return

dds_web/user.py
```diff
@@ -22,29 +23,48 @@ def login():
...
-        )
-        if not credentials_validated:
+
+        # Authenticate user
+        try:
+            user_ok = dds_auth.verify_user_pass(username=username, password=password)
+        except sqlalchemy.exc.SQLAlchemyError as sqlerr:
+            # TODO (ina): Create custom error page
+            app.logger.exception(sqlerr)
+            return str(sqlerr), 500
```
worukan, line 43: Could it be meaningful to return in a similar way as in the next except-block?

dds_web/user.py
```diff
@@ -22,29 +23,48 @@ def login():
...
-        session["current_user"] = user_info["username"]
-        session["current_user_id"] = user_info["id"]
-        session["is_admin"] = user_info.get("admin", False)
-        session["is_facility"] = is_facility
-        session["facility_name"] = user_info.get("facility_name")
-        session["facility_id"] = user_info.get("facility_id")
 
-        # temp admin fix
+        # User authentication not passed
+        if not user_ok:
```
worukan, line 51: I think this is the same thing as in the other user.py... What is the difference between  **dds_web/api/user.py** and **dds_web/user.py** ?

dds_web/api/user.py
inaod568: Correction: Fixing in this PR.

dds_web/crypt/auth.py
inaod568: Fair enough. Basically just created this function from another old function. But this is only used in the web interface - should I leave it for now since we're focusing on the REST API and CLI ?

dds_web/user.py
inaod568: I'm not sure what you mean here, which next except block? Since it's an endpoint we should probably create custom error pages and return those with the exception message, right?

dds_web/user.py
inaod568: They are very similar. The difference is that `dds_web/user.py` uses pure flask, while `dds_web/api/user.py` uses Flask Restful (-> Resources), so `def login()` accepts both `GET` and `POST` while `AuthenticateUser(flask_restful.Resource)` atm just accepts get.

Also that they return different things - `dds_web/user.py` functions return templates and redirects to different pages in the web interface, and `dds_web/api/user.py` returns json to the CLI.

dds_web/user.py
```diff
@@ -22,29 +23,48 @@ def login():
...
+        except sqlalchemy.exc.SQLAlchemyError as sqlerr:
+            # TODO (ina): Create custom error page
+            app.logger.exception(sqlerr)
+            return str(sqlerr), 500
+        except exceptions.AuthenticationError as autherr:
+            app.logger.exception(autherr)
             return render_template(
-                "user/login.html", next=request.form.get("next"), login_error_message=message
+                "user/login.html", next=request.form.get("next"), login_error_message=str(autherr)
             )
```
worukan, line 48: From line 44 to 48 is the next except block from the previous one in the previous comment. For login related errors, one usually would like to show a highlighted or popped-up error text/bubble instead of returning a custom error page.

inaod568: Ah, right, yes. Will fix!

PR #545, opened 2021-08-13
func.binary removed from some db queries due to errors and safespring…

inaod568:
… project added to docker-compose >>for now<<

Fixed the exceptions caused by `func.binary` in the db queries, but then the setup couldn't find the safespring project and therefore could not enter the rows in `db_init.py` -- Added `DDS_SAFE_SPRING_PROJECT` to the `docker-compose.yml` and set it as a local env variable. This will be changed since each facility will have their own, but just do that for now so that you can actually test stuff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ my_*
 dds_conf.yml
 safespring_invoicespec.csv
 dds_web/development/invoicing/*
+dds_variables.env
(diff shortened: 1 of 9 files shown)


ErikDanielsson (approved):
LGTM :+1: --  `dds put` works on my machine now. I'm not able to create projects now when logged in as `facility` though -- is that by design?

inaod568:
> LGTM 👍 -- `dds put` works on my machine now. I'm not able to create projects now when logged in as `facility` though -- is that by design?

Ah ok. No, but there have been changes in the db and the backend so it's probably calling a non-existent column somewhere. Will have a look at some point but there'll be changes to the project creation stuff so not prioritizing it a lot at the moment. Let me know if it creates a huge problem for you!

PR #547, opened 2021-08-13
Replace deprecated function call with new one

ErikDanielsson:
The API was using a deprecated function -- it should be up to date now.


inaod568 (approved):
Thanks for finding that.

PR #565, opened 2021-09-09
Minimal setup for the new authentication which supports both basic an…

worukan:
…d token authentication. Changed auth endpoint to token as this name reflects its function better than the previous one.

### DIFF
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -38,7 +38,7 @@ def jwt_token(username):
                 "exp": datetime.utcnow() + timedelta(hours=48),
             },
             app.config["SECRET_KEY"],
-            algorithm="HS256"
+            algorithm="HS256",
         )
         app.logger.debug(f"token: {token}")
     except (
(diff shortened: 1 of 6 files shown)


dds_web/api/user.py
```diff
@@ -5,44 +5,40 @@
 ###############################################################################
 
 # Standard library
-import datetime
+from datetime import datetime, timedelta
```
inaod568, line 8: We (dev group) decided on only importing modules, not functions and classes, if not absolutely necessary

dds_web/api/user.py
```diff
@@ -5,44 +5,40 @@
...
 
     try:
         token = jwt.encode(
             {
                 "user": username,
-                "project": {"id": project_id, "verified": project_access, "permission": permission},
-                "exp": datetime.datetime.utcnow() + datetime.timedelta(hours=48),
+                "exp": datetime.utcnow() + timedelta(hours=48),
             },
             app.config["SECRET_KEY"],
```
inaod568, line 40: Can we change this to `app.config.get("SECRET_KEY")` instead? Been thinking about this for a while.

dds_web/api/user.py
worukan: Done!

dds_web/api/user.py
worukan: Done!

inaod568 (approved):

PR #566, opened 2021-09-09
Cleanout web WIP/Suggestion

alneberg:
I probably went overboard with deleting the web related code. A lot of the python code will probably be reused in the api quite soon, but it's easy to fetch from the git history.

I'm mostly making this PR to see what you think, should we be more careful with deleting or should we bring things back one by one when we need them?


MatthiasZepper (approved):
Funny. I was suggesting exactly the same to Volkan this morning during the meeting...

inaod568:
Do we have a dedicated branch or tag for remembering the version before the deletion?

alneberg:
Yes I created the https://github.com/ScilifelabDataCentre/dds_web/tree/web_EOL branch. But maybe a tag is easier to find, maybe both?

inaod568:
Ah okey great. Yes I think both can be good 👍🏻

inaod568 (approved):

inaod568:
Will merge this now, @alneberg can you fix the tag?

PR #567, opened 2021-09-13
Cleanup unused

inaod568:
Have just cleaned up a bit and removed things that were not used.


alneberg (approved):
Haven't tested it, but everything looks fine what I can see. Will be much easier (hopefully) to understand the code when all the things not used is gone.

PR #571, opened 2021-09-14
Admin add user

inaod568:
These are the endpoints and the template for when inviting new users.


dds_web/__init__.py
```diff
@@ -117,33 +126,42 @@ def create_app():
...
     # Initialize database
     db.init_app(app)
+    mail.init_app(app)
+    # Setup admin
+    # import dds_web.database.models as models
 
+    # admin.init_app(app)
+    # admin.add_view(sqla.ModelView(models.User, db.session))
+
+    # FIXME
```
alneberg, line 136: This section was removed in a different PR right?

dds_web/__init__.py
```diff
@@ -117,33 +126,42 @@ def create_app():
...
-        server_metadata_url=app.config.get("OIDC_ACCESS_TOKEN_URL"),
-        client_kwargs={"scope": "openid profile email"},
-    )
+    # oauth.register(
+    #     "default_login",
+    #     client_secret=app.config.get("OIDC_CLIENT_SECRET"),
+    #     client_id=app.config.get("OIDC_CLIENT_ID"),
+    #     server_metadata_url=app.config.get("OIDC_ACCESS_TOKEN_URL"),
+    #     client_kwargs={"scope": "openid profile email"},
+    # )
```
alneberg, line 144: You know I prefer to delete code and not comment it out. What do you say?

dds_web/__init__.py
```diff
@@ -6,8 +6,9 @@
 
 # Standard library
 from datetime import datetime, timedelta
-import pytz
 import logging
+import pytz
+import time
```
alneberg, line 11: Are these imports here to be accessible for a separate file? It seems `time` is not used in this `__init__.py` file for example.

dds_web/forms.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # At least: (one lower case letter)(one upper case letter)(one digit)(special character)
+    password = wtforms.PasswordField(
+        "password",
+        validators=[
+            wtforms.validators.InputRequired(),
+            wtforms.validators.Length(min=10, max=64),
+            wtforms.validators.EqualTo("confirm", message="Passwords must match!"),
+            wtforms.validators.Regexp(
+                regex="(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[^A-Za-z0-9])",
```
MatthiasZepper, line 38:
```suggestion
                regex="(?=[^a-z\n]*[a-z])(?=[^A-Z\n]*[A-Z])(?=[^\d\n]*\d)^[\S]{10,64}$",
```
1.) Here the minimum length is already checked by the regex, so the length validator could be removed (unless we want to specifically return a separate error).
2.) The regex as it currently is does not capture the string, but matches only at one position. This means the regex might return several matches if the passwords starts with multiple characters of the same type, e.g. several lower-case letters. One can nicely check this by pasting it at [Regex101](https://regex101.com). I have no idea if this matters to the `Regexp()` function. 

_Additionally, one might want to consider [to just increase password length over complexity anyway.](https://xkcd.com/936/)_

dds_web/__init__.py
inaod568: `pytz` is used in `__init__.py` but `time` was probable used in a function I deleted or moved to another module, so I'll remove it 👍🏻

dds_web/api/user.py
```diff
@@ -60,6 +66,88 @@ def jwt_token(username):
...
 
+class ConfirmInvite(flask_restful.Resource):
+    def get(self, token):
+        """ """
+
+        s = itsdangerous.URLSafeTimedSerializer(app.config.get("SECRET_KEY"))
+
+        try:
+            # Get email from token
+            email = s.loads(token, salt="email-confirm", max_age=10000)
```
alneberg, line 77: 10000 seconds is a little short maybe? Maybe 24 hours, aka 86400 seconds?

alneberg: Assuming it's not the same person creating the invite that will confirm the invite, that person might not chech their email that quick...

dds_web/api/user.py
```diff
@@ -60,6 +66,88 @@ def jwt_token(username):
...
+        facility_name = None
+        if invite_row.is_facility:
+            try:
+                facility_info = (
+                    models.Facility.query.filter(models.Facility.id == invite_row.facility_id)
+                    .with_entities(models.Facility.name)
+                    .first()
+                )
+            except sqlalchemy.exc.SQLAlchemyError as sqlerr:
+                raise DatabaseError(str(sqlerr))
```
alneberg, line 101: Is this sqlerr only sent to the log or is it shown in the response as well? I guess it should not be shown to the end user?

dds_web/database/models.py
```diff
@@ -179,6 +180,26 @@ def __repr__(self):
...
+class Invite(db.Model):
+    """Invites for users not yet confirmed in DDS"""
+
+    # Table setup
+    __tablename__ = "invites"
+    __table_args__ = {"extend_existing": True}
+
+    # Columns
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    email = db.Column(db.String(80), unique=True, nullable=False)
```
alneberg, line 192: Apparently 255 characters is the maximum for an email address: https://stackoverflow.com/a/1199238/1472158, might not be worth the trouble of having the string field smaller than that?

dds_web/api/errors.py
```diff
@@ -236,7 +236,20 @@ def __init__(self, project, message="No key found for current project"):
...
+        general_logger.warning(self.message)
+
+
+# ----------------------------------------------------------------------------------- #
+
+
+# ----------------------------------------------------------------------------------- #
 
 
 errors = {
```
worukan, line 255: Do you consider including NoSuchInviteError in the set of errors and assign a response status?

inaod568: Ah, yes. Will add that. `FORBIDDEN` or `INTERNAL_SERVER_ERROR`? Since it should exist but it doesn't so something has gone wrong.

dds_web/forms.py
MatthiasZepper: I just had a look at the `wtforms.validators.Regexp()` and you are completely right: Both regex will do!

But just to elaborate on 2. to clarify why I suggested the change, have a look how both are matched on "passWord1!":

```
import re

regex = [
    "(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[^A-Za-z0-9])",
    "(?=[^a-z\n]*[a-z])(?=[^A-Z\n]*[A-Z])(?=[^\d\n]*\d)^[\S]{10,64}$",
]

for element in regex:
    m = re.findall(re.compile(element), "passWord1!")
    if m:
        print(m)

```

This returns:
```
['', '', '', '', '']
['passWord1!']
```
Because the password starts with "pass" (several characters of the same character class e.g. lowercase letters) the regex is also matched multiple times. `len(m)` is therefore 5 for the first and 1 for the second regex. 

In R, which I know much better than Python, it often causes an error if the length of an object changes unexpectedly because this typically also requires changing the class (type in Python) dynamically. For example, a function would usually return only one numerical value that is stored in an integer variable, but under certain conditions returns several numerical values at once, so they are coerced into a list.

@inaod568

dds_web/api/errors.py
inaod568: @worukan Which code is best here? Forbidden is for when the clients identity is known. `NOT_ACCEPTABLE`? Or something else?

worukan: Under which circumstance(s) will this be triggered?

inaod568: When a invited user clicks on the emailed link, and for some reason there is now row in the `Invite` table that matches the link.

worukan: We could use bad request for this. It should not happen but, in case it happens, the request refers to a non-existing entry which is a bad request.

inaod568:
The tests are failing, unsure why, would like some help figuring it out and implementing more tests.

inaod568:
> Added the tests I could think of regarding the `AddUser` endpoint, and then started looking into how to test the `ConfirmInvite` and `NewUser`. Started trying out Selenium which is really nice if we can get it to work. We need to get the links that are sent to the email, OR get the token in the `AddUser` endpoint when it's created, and somehow save that and then use it for the testing of the other two endpoints.

**Update:** I have added a few tests, including some for `ConfirmInvite` and one for `NewUser`. Additional tests can be added for `NewUser` in the `test_user_new.py` file by using `selenium`. Should be possible to do something like this:
1. Get the token in the same way as in `test_user_confirminvite.py:test_valid_token`
2. driver.get(tests.DDSEndpoint.USER_CONFIRM + token) --> Should open a browser and redirect to the form
3. Fill in form with selenium and submit
4. Assert that the user has been added to the database (User and Invite) and that the row no longer exists in the `Invite` table
_To do this you need to install selenium and do some config, there were a bunch of useful YouTube videos that helped._

However, I don't know how to do this with GitHub Actions. I could have a look and get into it myself but I don't have the time now, so I would like to merge this PR without adding all tests so that we can move on and this is no longer a blocker, and create a card for the additional required tests. 

@alneberg @MatthiasZepper @worukan @aanil

tests/test_user_add.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def test_add_user_with_researcher(client):
+    response = client.post(
+        tests.DDSEndpoint.USER_ADD,
+        headers=tests.UserAuth(tests.USER_CREDENTIALS["researchuser"]).post_headers(),
+        data=json.dumps(first_new_user),
+        content_type="application/json",
+    )
+    assert response.status == "403 FORBIDDEN"
```
alneberg, line 22: I thought we said researchers would be able to add users?

dds_web/api/marshmallows.py
```diff
@@ -80,3 +94,194 @@ def return_items(self, data, **kwargs):
...
+        invite = models.Invite.query.filter(
+            models.Invite.email == sqlalchemy.func.binary(data.get("email"))
+        ).one_or_none()
+        if not invite:
+            raise ddserr.InviteError
+
+        data["invite"] = invite
+
+    @marshmallow.post_load
+    def make_user(self, data, **kwargs):
```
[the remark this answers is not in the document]
inaod568: I kind of agree, but I will leave this for now. Since I'm in the process of adding marshmallows to other endpoints too I can have a look then what makes the most sense.

dds_web/api/user.py
```diff
@@ -47,6 +56,118 @@ def jwt_token(username):
...
+        s = itsdangerous.URLSafeTimedSerializer(flask.current_app.config.get("SECRET_KEY"))
+
+        try:
+            # Get email from token
+            email = s.loads(token, salt="email-confirm", max_age=10000)
+
+            # Get row from invite table
+            invite_row = models.Invite.query.filter(models.Invite.email == email).first()
+
+        except (itsdangerous.exc.SignatureExpired, itsdangerous.exc.BadSignature) as signerr:
```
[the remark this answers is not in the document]
inaod568: Good point!

PR #572, opened 2021-09-14
Flask command Init-dev-db

alneberg:
An additional flask command `init-dev-db` to decide when to create the tables and put the development data in the database. In the end this turned out to be more of an exercise to understand the flask internals a tiny bit better. It's actually not needed anymore since it's controlled by the `USE_LOCAL_DB` and the integrity error is already fixed with the `check_if_fill` method in `init_db`. 

I guess one can argue that this PR represents a slightly cleaner approach since the data is there is only one attempt to add the development data in the initial start of the container, while the solution in place right now will be triggered (but not followed through) for every restart of flask (read every code change).


inaod568:
I see that the `black` check is failing - do you have black installed locally?

inaod568 (approved):
Tried it out and it works 👍🏻 A bit concerned that it doesn't seem to notice if something is obviously wrong in the code though but I don't think that's an issue created in this PR

inaod568:
Just make sure to merge dev first as well

alneberg:
Great!

> A bit concerned that it doesn't seem to notice if something is obviously wrong in the code though but I don't think that's an issue created in this PR

Not sure what you mean with this

inaod568:
> Great!
> 
> > A bit concerned that it doesn't seem to notice if something is obviously wrong in the code though but I don't think that's an issue created in this PR
> 
> Not sure what you mean with this

If I write something in the code e.g. just some jibberish then it should give errors/exceptions etc but it doesn't.

PR #573, opened 2021-09-14
Increase field lengths for project title, file name and file subpath in database

aanil:
Updated fields from String to Text.


inaod568:
Closes #573

inaod568 (approved):

PR #574, opened 2021-09-15
hopefully all app changed to app_obj

inaod568:
It was a bit confusing with `app.py` and the `app` object, so the object was changed to `app_obj`, but not all imports in the rest of the DDS. Hopefully all are fixed in this PR.


alneberg:
I think this one can be closed in favour of #575.

PR #575, opened 2021-09-15
Factory all the way

alneberg:
Untested really since I can't get the only existing get URL to work... Please have a look and try it out. I think it's quite a lot cleaner to get rid of the `app` object.


inaod568:
The logging seems to be initiated multiple times: 
```
dds_backend | [2021-09-18 10:46:44,693] __init__ [INFO] Logging initiated.
dds_backend | [2021-09-18 10:46:45,376] __init__ [INFO] Initializing development db
dds_backend | [2021-09-18 10:46:45,808] __init__ [INFO] DB filled
dds_backend | Pending jobs:
dds_backend |     No pending jobs
dds_database | 2021-09-18 10:46:45 5 [Warning] Aborted connection 5 to db: 'DeliverySystem' user: 'apiuser' host: '172.20.0.3' (Got an error reading communication packets)
dds_backend |  * Serving Flask app "dds_web" (lazy loading)
dds_backend |  * Environment: development
dds_backend |  * Debug mode: on
dds_backend |  * Running on all addresses.
dds_backend |    WARNING: This is a development server. Do not use it in a production deployment.
dds_backend |  * Running on http://172.20.0.3:5000/ (Press CTRL+C to quit)
dds_backend |  * Restarting with stat
dds_backend |  * Debugger is active!
dds_backend |  * Debugger PIN: 326-456-840
dds_backend | [2021-09-18 10:46:46,603] __init__ [INFO] Logging initiated.
dds_backend | [2021-09-18 10:46:46,894] __init__ [INFO] Logging initiated.
```

dds_web/api/files.py
```diff
@@ -553,10 +554,10 @@ def put(self, _, project):
             file.time_latest_download = dds_web.utils.timestamp()
         except sqlalchemy.exc.SQLAlchemyError as err:
             db.session.rollback()
-            app.logger.exception(str(err))
+            flask.current_app.logger.exception(str(err))
             return flask.make_response(str(err), 500)
```
inaod568, line 558: This is something we need to look into too. What info to display and not display to the end user.

dds_web/utils.py
```diff
@@ -92,7 +92,7 @@ def invoice_units():
     """Get invoicing specification from Safespring, calculate and save GBHours and cost for each
     facility and project."""
 
-    app.logger.debug("Calculating invoicing info...")
+    current_app.logger.debug("Calculating invoicing info...")
```
inaod568, line 95: Change the import to `flask` and use `flask.current_app` like in the other files.

inaod568 (changes requested):

inaod568:
I like it 👏🏻 Just change the one file I requested a change for

alneberg:
> The logging seems to be initiated multiple times:
> 

Yes this is a bit of mystery to me as well. I will look into it a bit.

dds_web/utils.py
alneberg: Done! It was the only file where current_app was already imported, so I just used that one. Changed to `flask.current_app` in all the places now.

dds_web/api/files.py
alneberg: Yes, I don't think the error has to be very informative unless it can be predicted, basically giving a standard "500 Internal Server Error" in other cases.

alneberg:
> > The logging seems to be initiated multiple times:
> 
> Yes this is a bit of mystery to me as well. I will look into it a bit.

@inaod568, I managed to get rid of this when disabling the reloader  `flask run -h 0.0.0.0 -p 5000 --no-reload` but I think we want to keep that. Don't know why we didn't see this before though? Or maybe we did?

inaod568:
> > > The logging seems to be initiated multiple times:
> > 
> > 
> > Yes this is a bit of mystery to me as well. I will look into it a bit.
> 
> @inaod568, I managed to get rid of this when disabling the reloader `flask run -h 0.0.0.0 -p 5000 --no-reload` but I think we want to keep that. Don't know why we didn't see this before though? Or maybe we did?

@alneberg Ok! I believe that there has been double logging in the beginning before too, but now it's even more, at least 3 times I think.

alneberg:
> > > > The logging seems to be initiated multiple times:
> > > 
> > > 
> > > Yes this is a bit of mystery to me as well. I will look into it a bit.
> > 
> > 
> > @inaod568, I managed to get rid of this when disabling the reloader `flask run -h 0.0.0.0 -p 5000 --no-reload` but I think we want to keep that. Don't know why we didn't see this before though? Or maybe we did?
> 
> @alneberg Ok! I believe that there has been double logging in the beginning before too, but now it's even more, at least 3 times I think.

I guess the first one of the three is from the `init-dev-db` command? Potentially we could move that logging statement outside of the factory...

inaod568:
> > > > > The logging seems to be initiated multiple times:
> > > > 
> > > > 
> > > > Yes this is a bit of mystery to me as well. I will look into it a bit.
> > > 
> > > 
> > > @inaod568, I managed to get rid of this when disabling the reloader `flask run -h 0.0.0.0 -p 5000 --no-reload` but I think we want to keep that. Don't know why we didn't see this before though? Or maybe we did?
> > 
> > 
> > @alneberg Ok! I believe that there has been double logging in the beginning before too, but now it's even more, at least 3 times I think.
> 
> I guess the first one of the three is from the `init-dev-db` command? Potentially we could move that logging statement outside of the factory...

Yes that too, but I'm mostly thinking about the `dds_backend | [2021-09-18 10:46:44,693] __init__ [INFO] Logging initiated.` line.

inaod568 (approved):
The fixes look good and tested locally and at least the token endpoint is working great so 👏🏻

PR #577, opened 2021-09-15
Flask command Init-dev-db

inaod568:
Aand this is the original PR. Reverted the Revert.


alneberg:
Closing this one since it's now included in the #575 instead.

PR #579, opened 2021-09-16
Remove project auth endpoint, Hand the project to apis3connector and …

worukan:
…dbconnector after verifying it, Provide initial auth error handling, Clean up unused imports and obsolete code, Tidy up error classes, Apply new authentication to all endpoints allowing them to be used with either basic or token authentication, Verify project request parameter for the corresponding endpoints


dds_web/api/errors.py
```diff
@@ -35,55 +36,16 @@ class ItemDeletionError(exceptions.HTTPException):
...
-                **extra_info,
-                "current_user": username,
-                "action": actions.get(flask.request.endpoint),
-                "project": None,
-            },
-        )
 
-
-class ProjectPermissionsError(exceptions.HTTPException):
+class AccessDeniedError(exceptions.HTTPException):
```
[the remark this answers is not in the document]
inaod568: @worukan I think we should add the ID to the user table again. We need to be able to log which username attempts a specific action and since the username is the primary key at the moment this generates security warnings.

dds_web/api/project.py
```diff
@@ -44,161 +36,125 @@
...
-            proj_priv = (
-                models.Project.query.filter_by(public_id=project["id"])
-                .with_entities(
-                    models.Project.private_key,
-                    models.Project.privkey_nonce,
-                    models.Project.privkey_salt,
-                )
-                .first()
-            )
-        except sqlalchemy.exc.SQLAlchemyError as err:
```
inaod568, line 162: I think this is fine for now but we need to look into how to show what information to the end user.

inaod568 (commented):
Aside from my review comments: The goal of each PR should be to have something that is working, and I don't think the DDS_CLI works with this version atm right? I know that's not the case with `dds_web/dev` and `dds_cli/dev` atm either but all the more reason that we become more strict with this now before the next PR in my opinion. So I'm thinking we should wait with this PR and merge it once you have altered the CLI so that it matches the changes in the API.

dds_web/api/errors.py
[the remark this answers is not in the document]
worukan: Fixed! I think there is no need to bring back the previously removed ID.

dds_web/api/project.py
worukan: I will work on the keys and related parts in another branch later. It doesn't belong in the current work.

inaod568: I meant the part with `flask.make_response(str(err), 500)` below, that's what produces the codeql warning. But this is fine for now 👍🏻

inaod568 (approved):
Looks good! 👏🏻

inaod568 (approved):

PR #589, opened 2021-09-24
Moving the schedulers to init

MatthiasZepper:
Addresses #522

Suggested changes:

  - Moved the scheduling from utils to init.
  - Spread the jobs more out over the night.

Testing:

 - Verified that all functions and required packaged are imported.
 - Further testing currently not possible because both, the upstream dev and my feature branch, crash the container with `dds_backend   | Error: No such command 'init-dev-db'.` in the logs.

### DIFF
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -143,9 +143,7 @@ def create_app():
         app.register_blueprint(api_blueprint, url_prefix="/api/v1")
 
         # Set-up the schedulers
-        from dds_web.utils import scheduler_wrapper
-
-        scheduler_wrapper()
+        dds_web.utils.scheduler_wrapper()
 
         return app
 
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -134,7 +134,6 @@ def create_app():
     )
 
     app.cli.add_command(fill_db_wrapper)
-    app.cli.add_command(scheduler_wrapper)
 
     with app.app_context():  # Everything in here has access to sessions
         from dds_web.database import models
@@ -146,6 +145,8 @@ def create_app():
         from dds_web.api import api_blueprint
 
         app.register_blueprint(api_blueprint, url_prefix="/api/v1")
+
+        scheduler_wrapper()
 
         return app
 
@@ -167,8 +168,6 @@ def fill_db_wrapper():
 ####################################################################################################
 
 
-@click.command("start-schedulers")
-@flask.cli.with_appcontext
 def scheduler_wrapper():
     scheduler = BackgroundScheduler(
         {
@@ -182,7 +181,7 @@ def scheduler_wrapper():
     )
     flask.current_app.logger.info("Initiated main scheduler")
 
-    from dds_web.utils import invoice_units, remove_invoiced, remove_expired, permanent_delete
+    # from dds_web.utils import invoice_units, remove_invoiced, remove_expired, permanent_delete
 
     # Schedule invoicing calculations every 30 days
     # TODO (ina): Change to correct interval - 30 days
@@ -241,4 +240,4 @@ def scheduler_wrapper():
 
     # Print the currently scheduled jobs as verification:
     flask.current_app.logger.info("Currently scheduled jobs:")
-    scheduler.print_jobs()
+    flask.current_app.logger.info(scheduler.print_jobs())
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -18,6 +18,11 @@ from logging.config import dictConfig
 from authlib.integrations import flask_client as auth_flask_client
 from flask_httpauth import HTTPBasicAuth, HTTPTokenAuth, MultiAuth
 
+# imports related to scheduling
+import atexit
+from apscheduler.schedulers.background import BackgroundScheduler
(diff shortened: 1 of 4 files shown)


inaod568:
There seems to be a problem with the imports added to the `__init__.py` file. When I commented out the new imports and the backgroundscheduler stuff, it doesn't complain about the `init-dev-db`.

inaod568:
@MatthiasZepper I will convert this to a draft PR and you can just mark it as ready for review when you're ready again 👍🏻

MatthiasZepper:
I think, the last commit should close the issue  #589 now.

Unfortunately, it was a bit complicated to get APScheduler and Flask to work nicely together. There is a existing package [flask-apscheduler](https://github.com/viniciuschiele/flask-apscheduler), but this introduces several new dependencies and possibly also a conflicting REST-API. 

[Root problem is, that Flask in debug mode spawns a child process so that it can restart the process each time the code changes](https://github.com/agronholm/apscheduler/issues/250). The new child process initializes and starts a new APScheduler on its own, causing e.g. the jobs to be created twice within in the same database table: `pymysql.err.IntegrityError: (1062, "Duplicate entry 'calc_costs' for key 'PRIMARY'") error`

Apparently, the reload is done with a subprocess.call, so there are two different, isolated Python interpreters running at the same time. If there are multiple instances running in parallel can only be detected at the level of the webserver. Therefore, I had to introduce [_Werkzeug_ ](https://github.com/pallets/werkzeug), a utility library for WSGI, as additional dependency. WSGI itself is a protocol that ensures that web applications work nicely together on a webserver. This package contains a convenient function that will return TRUE if the application is running from within the Werkzeug reloader subprocess. Fortunately, this also allows to detect if an application is executed within a Flask subprocess.

In that case the execution of the _scheduler_wrapper()_ function is stopped with a return and no conflicts with the scheduler in the main process arise. 

**When Flask is run in production, this shouldn't matter anyway, because it is just Flask's debug mode that causes this issue in the first place.**

dds_web/__init__.py
```diff
@@ -144,6 +142,11 @@ def create_app():
 
         app.register_blueprint(api_blueprint, url_prefix="/api/v1")
 
+        # Set-up the schedulers
+        from dds_web.utils import scheduler_wrapper
```
inaod568, line 146: Could you change to this instead (just import changes)
```
        # Register blueprints
        import dds_web.api

        app.register_blueprint(dds_web.api.api_blueprint, url_prefix="/api/v1")

        # Set-up the schedulers
        import dds_web.utils
        dds_web.utils.scheduler_wrapper()
```

dds_web/utils.py
```diff
@@ -10,13 +10,17 @@
...
 import pandas
 from contextlib import contextmanager
 import flask
 import sqlalchemy
 
+# # imports related to scheduling
+import atexit
+
+import werkzeug
+from apscheduler.schedulers.background import BackgroundScheduler
```
inaod568, line 22: `from apscheduler.schedulers import background`
And 
`background.BackgroundScheduler` where it's used

inaod568 (changes requested):

inaod568:
Tested and it works great! So just change the import stuff I commented on and I'll merge this 👏🏻

inaod568 (approved):

dds_web/utils.py
```diff
@@ -268,62 +271,109 @@ def permanent_delete():
...
-scheduler.start()
-
-# Shut down the scheduler when exiting the app
-atexit.register(lambda: scheduler.shutdown())
+####################################################################################################
+# BACKGROUND SCHEDULER ###################################################### BACKGROUND SCHEDULER #
+####################################################################################################
+
+# Mind this discussion on the intricate peculiarities of running APScheduler in combination with Flask
+# https://github.com/agronholm/apscheduler/issues/250
```
alneberg, line 279: From the latest responses on this issue it seems to only affect setups without the factory pattern that we are using. So I think you can remove this comment from the code.

requirements.txt
```diff
@@ -13,6 +13,7 @@ pytz==2020.1
 zstandard
 authlib
 apscheduler
+werkzeug
```
alneberg, line 16: werkzeug is a dependency of Flask already: https://github.com/pallets/flask/blob/aa1d34dc51fc63dabc46f388822ba73b15d3ef97/setup.py#L7

So you can remove this line again.

alneberg: Or maybe it's good to make it explicit? I don't know

alneberg (approved):
Great! I think the solution checking for flask debug mode and reloader looks beautiful! Two minor comments about things that can be removed, but I would be happy to merge with them in as well actually.

requirements.txt
inaod568: I think we can keep it explicit, just cause we're actually importing it.

dds_web/utils.py
MatthiasZepper: True. Updated.

requirements.txt
MatthiasZepper: After personal discussion, we decided to keep for now, but are not strongly opinionated on the issue.

Issue #587, opened 2021-09-22
Add function to create the db tables but not fill them 


inaod568:
Before the next dev deployment we need to have another function which just calls the `db.create_all()`, without actually filling the database tables. Currently the db is filled with info publicly available on GitHub, including the passwords. Emptying the database in the dev deployment for now.

alneberg:
Apparently the `db.create_all()` method does nothing to tables that already exists so it seems fairly safe to run it by default:

https://github.com/pallets/flask-sqlalchemy/blob/25edfae1835f0ac70483cb38e0d83d3f00aeb208/src/flask_sqlalchemy/__init__.py#L1044-L1048
```
def create_all(self, bind="__all__", app=None):
    """Create all tables that do not already exist in the database.
        This does not update existing tables, use a migration library
        for that.
```

But as it states, if the tables exists but are different than what the model states, this will still not do anything.

inaod568:
Yes 👍🏻 Great

PR #590, opened 2021-09-27
Automatic testing with docker-compose and github actions

alneberg:
- A framework to run tests with docker-compose
- Two simple tests for the token endpoint.
- A github workflow to run the tests
- Creating db and tables on start (issue #587) 
- database_uri as parameter to the create_app method
- Documenting in the README how to run the tests locally
- Default value for safe spring development variable


inaod568:
Setting this as a draft for now, just mark it as ready for review when done 👍🏻

inaod568 (approved):
LGTM! Tested locally, it complains about stuff, so it works

PR #594, opened 2021-09-28
Proof of concept for User model inheritance

alneberg:
Just showing the idea in python.

1. The authentication is unaffected from what I can tell, the username and password is accessed just as before.
2. The role property works as in `current_user.role`
3. Queries can be based on either User, ResearchUser or UnitUser.

The role implementation below is an even more draft than the rest, so please ask if anything looks weird. I made a test using the `@auth.login_required(role=[])` method, but it needs more granularity as well as in the `verify` method already in place.

### DIFF
--- a/dds_web/database/models.py
+++ b/dds_web/database/models.py
@@ -122,7 +122,7 @@ class User(db.Model):
     def __repr__(self):
         """Called by print, creates representation of object"""
 
-        return f"<User {self.username}>"
+        return f"<User {self.username}:{self.role}>"
 
 
 class ResearchUser(User):
@@ -136,7 +136,8 @@ class ResearchUser(User):
         "polymorphic_identity": "research_user",
     }
 
-    def role():
+    @property
+    def role(self):
         return "researcher"
 
 
@@ -150,6 +151,13 @@ class UnitUser(User):
     __mapper_args__ = {
         "polymorphic_identity": "unit_user",
     }
+
+    @property
+    def role(self):
+        if self.admin:
+            return "unit_admin"
+        else:
+            return "unit_user"
 
 
 class Identifier(db.Model):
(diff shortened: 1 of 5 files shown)


inaod568:
Yes, this is more or less what I have too (plus the association table etc with projects). I'm on board so if @worukan is too then we can work on this together @alneberg 👍🏻 I have these changes plus more where I am altering according to the ER diagram so I want to work from that branch in that case, just so that we don't do double work

dds_web/database/models.py
```diff
@@ -130,7 +122,42 @@ class User(db.Model):
...
+    __mapper_args__ = {
+        "polymorphic_identity": "unit_user",
+    }
+
+    @property
+    def role(self):
+        if self.admin:
+            return "unit_admin"
+        else:
+            return "unit_user"
```
worukan, line 161: Would there be a SuperAdminUser class inheriting from User? If not how do you think of it?

dds_web/database/models.py
```diff
@@ -130,7 +122,42 @@ class User(db.Model):
     def __repr__(self):
         """Called by print, creates representation of object"""
 
-        return f"<User {self.username}>"
+        return f"<User {self.username}:{self.role}>"
```
worukan, line 125: The base User class does not have a role column. How does this work?

alneberg: It would indeed crash for a object which is of the class User but not of any of the subclasses UnitUser or ResearchUser. We should probably define that property in the base class as well and set it to None or something.

dds_web/database/models.py
alneberg: Yeah I think I would do it that way.

inaod568: I agree.

worukan: +1

dds_web/database/models.py
worukan: +1

PR #598, opened 2021-09-29
Er diagram alterations

inaod568:
Have changed the database to match the ER diagram in [lucidchart](https://lucid.app/lucidchart/5ff0f0bb-7683-4807-9c1d-6a53b6fff562/edit?viewport_loc=-2052%2C-1249%2C4779%2C2512%2C0_0&invitationId=inv_00e6985e-992a-4513-a36d-749ac91dbda5)
(With exception to the `owner` column in `project_users` - under discussion and WIP)

This includes changing some columns to `DateTime` instead of `String` and changing the places where the `dds_web.utils.timestamp()` function was used to `dds_web.utils.current_time()`


dds_web/database/models.py
```diff
@@ -103,17 +107,15 @@ class User(db.Model):
     __table_args__ = {"extend_existing": True}
 
     # Columns
-    # id = db.Column(db.Integer, primary_key=True, autoincrement=True)
-    username = db.Column(db.String(20), primary_key=True, autoincrement=False)
-    password = db.Column(db.String(120), unique=False, nullable=False)
-    role = db.Column(db.String(50), unique=False, nullable=False)
-    first_name = db.Column(db.String(50), unique=False, nullable=True)
-    last_name = db.Column(db.String(50), unique=False, nullable=True)
+    username = db.Column(db.String(50), primary_key=True, autoincrement=False)
```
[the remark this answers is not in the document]
inaod568: Yes, the users themselves set the username and password when they get an invite to get a dds account.

dds_web/development/db_init.py
```diff
@@ -23,16 +23,16 @@
 # Create Units
 units = [
     Unit(
-        public_id="public_unit_id",
+        public_id="unit1",
         name="Unit 1",
-        internal_ref="fac",
-        safespring=current_app.config.get("DDS_SAFE_SPRING_PROJECT"),
+        internal_ref="someunit",
+        safespring=current_app.config.get("DDS_SAFE_SPRING_PROJECT", "dds.example.com"),
```
alneberg, line 29: Did this fix actually make a difference for you? For me, the `DDS_SAFE_SPRING_PROJECT` was set to `None` so this get call would still return `None`...

alneberg: Insted, adding a similar get call in the `config.py` file would fix the issue.

dds_web/utils.py
```diff
@@ -16,17 +16,25 @@
 from contextlib import contextmanager
 import flask
 import sqlalchemy
+import pytz
 
 # Own modules
 from dds_web.database import models
 from dds_web import db, C_TZ
+import dds_web.utils
```
alneberg, line 24: This would import this file inside this file?

dds_web/utils.py
```diff
@@ -95,7 +104,7 @@ def invoice_units():
     # From safespring
     old_file = parent_dir / pathlib.Path("development/invoicing/safespring_invoicespec.csv")
 
-    current_time = timestamp(ts_format="%Y-%m-%d_%H-%M-%S")
+    current_time = dds_web.utils.current_time()
```
alneberg, line 107: This should just be `current_time`? or is it standard praxis to refer to things inside the same file with the full path? I just haven't seen it before.

dds_web/development/db_init.py
inaod568: Did some more testing and you're right, it doesn't. I have the `dds_app.cfg` file too so that's why it was filling it in.

dds_web/utils.py
inaod568: Removing.

dds_web/development/db_init.py
inaod568: We already have that in the config.py. But what if the config.py file isn't used and the app config is actually done by the dds_app.cfg and DDS_SAFE_SPRING_PROJECT isn't set?

alneberg: I think `dds_app.cfg` only overwrites whatever is actually in that file. Anything else is still from the config.py, so it shouldn't be possible to have it unset then?

dds_web/database/models.py
```diff
@@ -8,6 +8,7 @@
 import datetime
 
 # Installed
+import pytz
```
aanil, line 11: Maybe I missed something, but is this module used in this file?

inaod568: No, it's not, will remove!

dds_web/development/db_init.py
inaod568: Ah okey, I thought it just overwrites everything, and those that were set before are removed
But in that case, yes, I can remove this

inaod568:
All checks have passed and I have fixed the parts you've commented on. Ok for me to merge? 
@worukan @alneberg @aanil

dds_web/api/project.py
```diff
@@ -4,6 +4,9 @@
 # IMPORTS ################################################################################ IMPORTS #
 ####################################################################################################
 
+# Standard Library
+import datetime
```
aanil, line 8: This module seems to not be used?

alneberg:
👍 here as well, sorry for the delay!

PR #608, opened 2021-10-04
Fix scheduler issue with Johannes

MatthiasZepper:
I think the scheduler issue is now sorted out thanks to Johannes pointing out that in order to `get_jobs()` the scheduler must not only be configured but also be started. After moving the `start_scheduler()` command before adding the jobs it seems that there a no problems anymore.


inaod568 (approved):
Looks good and works for me so if @alneberg and @aanil is not experiencing issues either I say we can merge. 👍🏻

alneberg:
Looks great! 👍

PR #620, opened 2021-10-06
Add tests for CreateProject

aanil:


tests/test_project.py
```diff
@@ -0,0 +1,116 @@
...
+        db.session.query(models.Project)
+        .filter_by(
+            created_by="username",
+            title=proj_data["title"],
+            pi=proj_data["pi"],
+            description=proj_data["description"],
+        )
+        .one_or_none()
+    )
+    assert created_proj == None
```
inaod568, line 29: Small thing but a general rule to use `is None` instead of `== None`

inaod568: @aanil

inaod568 (approved):

PR #627, opened 2021-10-12
User inheritance and roles

inaod568:
Joined table inheritance for users:
* `User` : Base user table, including all rows which all user types should have
    * `ResearchUser`: Connected to projects via the `ProjectUsers` association table.
        - The association table also has an extra column called `owner` which allows the user to be a projectowner for some projects (with some admin permissions) and only a team member in some (only list and download permissions)
    * `UnitUser`: Does not have a connection to the projects table directly, only through the connected unit. 
         - A UnitUser has two different possible roles depending on the `is_admin` column: `Unit Personnel` / `Unit Admin`
    * `SuperAdmin`: Us at DC -- we can do anything. Not a lot implemented here, mostly focusing on the others. 

The endpoints should also be fixed so that they have `@auth.login_required(roles=[".."])` where it's applicable. 

**NOTE! The code is being refactored, this is an initial PR for the roles and inheritance to work. This is due to that the PR will become huge if I do everything at once.**

### DIFF
--- a/dds_web/api/errors.py
+++ b/dds_web/api/errors.py
@@ -14,6 +14,7 @@ import http
 
 # Own modules
 from dds_web import actions
+from dds_web import auth
 
 ####################################################################################################
 # LOGGING ################################################################################ LOGGING #
@@ -50,7 +51,6 @@ class AccessDeniedError(exceptions.HTTPException):
 
     def __init__(
         self,
-        username=None,
         project=None,
         message="The user does not have the necessary permissions.",
     ):
@@ -60,7 +60,7 @@ class AccessDeniedError(exceptions.HTTPException):
             message,
             extra={
                 **extra_info,
-                "current_user": username,
+                "current_user": auth.current_user(),
                 "action": actions.get(flask.request.endpoint),
                 "project": project,
             },
@@ -70,9 +70,7 @@ class AccessDeniedError(exceptions.HTTPException):
 class DatabaseError(exceptions.HTTPException):
     """Baseclass for database related issues."""
 
-    def __init__(
-        self, message="The DDS encountered an Flask-SQLAlchemy issue.", username=None, project=None
-    ):
+    def __init__(self, message="The DDS encountered an Flask-SQLAlchemy issue.", project=None):
         super().__init__(message)
 
         general_logger.warning(message)
@@ -81,7 +79,7 @@ class DatabaseError(exceptions.HTTPException):
             message,
             extra={
                 **extra_info,
-                "current_user": username,
+                "current_user": auth.current_user(),
                 "action": actions.get(flask.request.endpoint),
                 "project": project,
             },
@@ -100,7 +98,7 @@ class EmptyProjectException(exceptions.HTTPException):
             message,
             extra={
                 **extra_info,
-                "current_user": username,
+                "current_user": auth.current_user(),
                 "action": actions.get(flask.request.endpoint),
                 "project": project,
             },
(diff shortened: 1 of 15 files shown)


dds_web/api/marshmallows.py
```diff
@@ -0,0 +1,87 @@
...
+import flask
+import marshmallow
+import sqlalchemy
+
+# Own modules
+from dds_web import utils
+from dds_web.api import errors as ddserr
+from dds_web import auth
+from dds_web.database import models
+from dds_web.api import db_connector
```
alneberg, line 20: Some unused here as well I think.

dds_web/security/auth.py
```diff
@@ -49,7 +49,8 @@ def get_user_roles(user):
 
 
 def get_user_roles_common(user):
-    return user.role
+    flask.current_app.logger.debug(f"ROLE: {getattr(user, 'role')}")
+    return getattr(user, "role")
```
alneberg, line 53: Should user.role ever be missing? I guess that's why you have the `getattr` instead of what was before...?

dds_web/api/errors.py
```diff
@@ -226,14 +215,24 @@ def __init__(self, message="No method found in request."):
...
+class KeyNotFoundError(exceptions.HTTPException):
+    """Key not found in database."""
 
     def __init__(self, project, message="No key found for current project"):
         self.message = f"{message}: {project}"
         super().__init__(self.message)
 
-        general_logger.warning(self.message)
+
+class PublicKeyNotFoundError(KeyNotFoundError):
```
alneberg, line 226: Do we need more types of exceptions?

dds_web/security/auth.py
inaod568: I had a bug, that's why this was changed. Will change back. `role` should always be set.

dds_web/database/models.py
```diff
@@ -36,38 +60,33 @@ class Unit(db.Model):
...
 
-    # Foreign key -- One unit can have many projects
-    unit_id = db.Column(db.Integer, db.ForeignKey("units.id"))
-    # Foreign key -- A user can have many projects
-    created_by = db.Column(db.String(50), db.ForeignKey("users.username"))
+    # Foreign keys
+    # One unit can have many projects
+    unit_id = db.Column(db.Integer, db.ForeignKey("units.id"), nullable=False)
+
+    # One unit can be created by one user
```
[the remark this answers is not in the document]
alneberg: But it's the `created_by` attribute on Project that the comment is for right?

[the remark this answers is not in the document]
alneberg: So instead of `# one unit can be created by one user` it should be `# One project can be created by one user`?

dds_web/database/models.py
```diff
@@ -91,54 +110,133 @@ class Project(db.Model):
...
 
+class ResearchUser(User):
+    """Data model for research user accounts."""
+
+    __tablename__ = "researchusers"
+    __mapper_args__ = {"polymorphic_identity": "researchuser"}
+
+    # primary key and foreign key pointing to users
+    username = db.Column(db.String(50), db.ForeignKey("users.username"), primary_key=True)
+    project_associations = db.relationship("ProjectUsers", back_populates="researchuser")
```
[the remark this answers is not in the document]
alneberg: Ok, no dealbreaker now. Can revisit if it turns out to be an issue, which it might not be.

dds_web/api/errors.py
inaod568: No I understood what you meant :) It's changed to `KeyNotFoundError` instead of `PublicKey...`, `PrivateKey...` and `S3Keys...`

inaod568: They were used after the refactoring in the other branches (the next PRs) but I agree that it may be a bit too picky to have so many different custom ones

PR #631, opened 2021-10-13
Encrypt the existing signed token

worukan:
This can be considered as an initial step for the encrypted token. We can change the encryption to elliptic curve but it requires more configuration on the server and either a public key rest endpoint or registering the public key where it can be found easily. The current selection of signature and encryption schemes seem to be adequate according to [https://datatracker.ietf.org/doc/html/rfc7518]. 

Testing is done with token and project listing endpoints via both basic and token authentication. Inspected the encrypted token contents.

### DIFF
--- a/dds_web/api/errors.py
+++ b/dds_web/api/errors.py
@@ -42,7 +42,7 @@ class KeyLengthError(SystemExit):
 
     def __init__(self, encryption_key_char_length):
         message = (
-            f"SECRET KEY MUST BE AT LEAST {encryption_key_char_length} "
+            f"SECRET KEY MUST BE {encryption_key_char_length} "
             f"CHARACTERS LONG IN ORDER TO SATISFY THE CURRENT TOKEN ENCRYPTION!"
         )
         super().__init__(message)


worukan:
- It is now possible to authenticate with both a signed-only and a signed-encrypted token while the code is currently using a signed-only token for now. Testing is done manually by changing the token construction.
- datetime.utcnow is replaced with datetime.now as it caused a time validation issue and the python docs states that now is preferred over utcnow.
- KeyLengthError http status value is removed as it is not applicable
- Nonce is added to the token to prevent replay attacks and tested manually by constructing tokens with the same sub and exp claims which constitute the minimal token for us.
- Check if token is expired and if yes, return a token expired response.
- Token parameters are implemented for flexible expiry time and additional claims.
- Expiration time has been tested manually for -2, -1 and +1 minutes in timestamp. Exactly on time and 2 days ahead expiry times are tested as well. Tampering with the expiration time or any other part of a signed or an encrypted token are tested as well. The manual tests consist of collecting different tokens at runtime, decoding/encoding them in base64, changing a specific part of a token and feed it back to the server as part of a request.
- There are three test cases in test_token. This file is meant for testing any token related functionality whether a signed-only or a signed-encrypted token.

dds_web/__init__.py
```diff
@@ -164,6 +163,13 @@ def create_app(testing=False, database_uri=None):
         # Set-up the schedulers
         dds_web.utils.scheduler_wrapper()
 
+        from dds_web.api.user import ENCRYPTION_KEY_CHAR_LENGTH
+
+        if len(app.config.get("SECRET_KEY")) != ENCRYPTION_KEY_CHAR_LENGTH:
+            from dds_web.api.errors import KeyLengthError
+
+            raise KeyLengthError(ENCRYPTION_KEY_CHAR_LENGTH)
+
```
inaod568, line 172: Should it be equal or at least the number of characters? This checks equal, but the `KeyLengthError` says "at least". Change the KeyLengthError message?

worukan: This is intentionally written as it is. We don't accept it if it is any shorter than the required length, but we also don't accept it if it is longer than the required length, because a longer key is not fully used by the selected algorithm. It could give a false sense of security with a longer key. Perhaps we can change the error message? For example, remove "at least" from the error message?

PR #642, opened 2021-10-20
Fix foreign keys and relationships

inaod568:
Some of the relationships in the database were specified as `back_populates` and some as `backref`. Have changed to `backref` since it automatically updates the foreign key in the other table, and reduces the amount of code needed.

Example: 
```
class Unit(db.Model):
    """Data model for unit accounts."""

    # Relationships
    # One unit can have many users
    users = db.relationship("UnitUser", backref="unit")


class UnitUser(User):
    """Data model for unit user accounts"""

    # Foreign key and backref with infrastructure
    unit_id = db.Column(db.Integer, db.ForeignKey("units.id"), nullable=False)
```
This "creates a column" called `unit` in `UnitUser` which can be used to access the responsible unit, and you can create entries in the database in the following way:
```
new_unit = Unit(<all the columns>)
new_user = UnitUser(<all the columns except unit_id>)

new_user.unit = new_unit 

db.session.add(new_user)
db.session.commit()
```

This automatically adds `new_unit` to the database as well, and sets the `new_user.unit_id` automatically to `new_unit.id`.
For `back_populates` you also need to explicitly create the relationship in the `UnitUser` class: 
```
class Unit(db.Model):
    """Data model for unit accounts."""

    # Relationships
    # One unit can have many users
    users = db.relationship("UnitUser", back_populates="unit")
    

class UnitUser(User):
    """Data model for unit user accounts"""

    # Foreign key and backref with infrastructure
    unit_id = db.Column(db.Integer, db.ForeignKey("units.id"), nullable=False)
    unit = db.relationship("Unit", back_populates="users")
```

Backref simplifies.

### DIFF
--- a/dds_web/database/models.py
+++ b/dds_web/database/models.py
@@ -55,7 +55,7 @@ class Unit(db.Model):
     name = db.Column(db.String(255), unique=True, nullable=False)
     internal_ref = db.Column(db.String(50), unique=True, nullable=False)
     safespring = db.Column(db.String(255), unique=False, nullable=False)  # unique=True later
-    days_to_expire = db.Column(db.Integer, unique=False, nullable=False, default=30)
+    days_to_expire = db.Column(db.Integer, unique=False, nullable=False, default=90)
     counter = db.Column(db.Integer, unique=False, nullable=True)
 
     # Relationships
--- a/dds_web/database/models.py
+++ b/dds_web/database/models.py
@@ -60,7 +60,7 @@ class Unit(db.Model):
 
     # Relationships
     # One unit can have many users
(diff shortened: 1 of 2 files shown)


MatthiasZepper (approved):
Except for the (irrelevant) typo _dye_ instead of _due_  in the comment, it looks good to me. I agree that we should be consistent in terms of using either `backref` or `back_populates` and not mix the two. 

In general, some people seem to prefer `back_populates` for making the model code more explicit without incurring hidden/implicit properties, but since you are maintaining the nice database schema which we can put into the documentation, I think we will be fine.

PR #643, opened 2021-10-21
Remove `public_id` column from `File` and `ExpiredFile`

i-oden:
Barely used, basically only in print and actually creating the Files, which just complicates since we need to make sure that the ids are in fact unique for the tables. 

If we find we need it later then we can add it again but atm I don't see the point.


alneberg (approved):
Looks good!

Issue #662, opened 2021-10-26
Remove timezones from database config and inserts


inaod568:
Remove any time zone changes in the database / queries.
Translate to local time zone if/when the time stamps should be displayed to the user.

alneberg:
We probably want to have this one done before the database is taken into production at least. Maybe not strictly needed before testing period though.

inaod568:
Agreed!

PR #669, opened 2021-10-28
Move marshmallows to schema folder

inaod568:
First attempted to put the Schemas in the corresponding modules, but that resulted in circular imports. This way all schemas are not in the same file and it's more structured.


dds_web/api/schemas/file_schemas.py
```diff
@@ -0,0 +1,93 @@
...
+# SCHEMAS ################################################################################ SCHEMAS #
+####################################################################################################
+
+
+class NewFileSchema(project_schemas.ProjectRequiredSchema):
+    """Validates and creates a new file object."""
+
+    # Length minimum 1 required, required=True accepts empty string
+    name = marshmallow.fields.String(required=True, validate=marshmallow.validate.Length(min=1))
+    name_in_bucket = marshmallow.fields.String(
```
MatthiasZepper, line 29: The length validators alone might not be sufficient? Aren't there additional limitations on the filenames allowed in buckets?

dds_web/api/schemas/file_schemas.py
```diff
@@ -0,0 +1,93 @@
...
+                    )
+                )
+                .with_entities(models.File.id)
+                .one_or_none()
+            )
+        except sqlalchemy.exc.SQLAlchemyError:
+            raise
+
+        if file:
+            raise FileExistsError
```
MatthiasZepper, line 65: Just for me to understand: The built-in OSerror `FileExistsError` is raised here, because raising a custom error with a respective API response would leak information?

dds_web/api/schemas/file_schemas.py
```diff
@@ -0,0 +1,93 @@
...
+            size_original=data.get("size"),
+            size_stored=data.get("size_processed"),
+            compressed=data.get("compressed"),
+            salt=data.get("salt"),
+            public_key=data.get("public_key"),
+            checksum=data.get("checksum"),
+        )
+
+        new_version = models.Version(
+            size_stored=new_file.size_stored, time_uploaded=dds_web.utils.current_time()
```
MatthiasZepper, line 84: I didn't follow the Slack discussion entirely, but I think we agreed on using UTC times on all timestamps, but `dds_web.utils.current_time()` seems to return the time in the current time zone?

MatthiasZepper (approved):
Minor modifications may or may not be advisable, but in general I didn't find any bugs or issues.

dds_web/api/schemas/project_schemas.py
```diff
@@ -0,0 +1,86 @@
...
+        raise
+
+    if not project:
+        flask.current_app.logger.warning("No such project!!")
+        raise ddserr.NoSuchProjectError(project=spec_proj)
+
+    return project
+
+
+def verify_project_access(project):
```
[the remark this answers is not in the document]
inaod568: `project` is an actual project - the row in the `Project` table. 
`spec_proj` is the project id specified by the user.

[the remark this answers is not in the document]
inaod568: `project` is an actual project - the row in the `Project` table. 
`spec_proj` is the project id specified by the user.

dds_web/api/schemas/project_schemas.py
```diff
@@ -0,0 +1,86 @@
...
+####################################################################################################
+
+
+class ProjectRequiredSchema(marshmallow.Schema):
+    """Schema for verifying an existing project in args and database."""
+
+    project = marshmallow.fields.String(required=True)
+
+    class Meta:
+        unknown = marshmallow.EXCLUDE  # TODO: Change to RAISE
```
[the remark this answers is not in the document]
inaod568: In some of the endpoints the `ProjectRequiredSchema` is called with `ProjectRequiredSchema().load(flask.request.args)` --> if there are other valid, but not used for this specific schema, args passed in it will raise a validation error which it shouldn't. The comment about changing back is for after the refactoring, then this shouldn't still happen.

dds_web/api/schemas/project_schemas.py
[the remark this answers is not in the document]
MatthiasZepper: I see. Thanks for clarification!

dds_web/api/schemas/file_schemas.py
inaod568: I haven't been able to find any limitations regarding that. Only the buckets. But it's possible. We haven't run into that issue yet though.

dds_web/api/schemas/file_schemas.py
inaod568: No just trying to use as many non-custom exceptions as possible. In this case this really shouldn't happen so it should completely fail. Could also use ValidationError which will fail as well, I just added FileExistsError because it seemed appropriate at the time. 👍🏻

dds_web/api/schemas/file_schemas.py
inaod568: Yes, it does. This should be changed but that's another card/PR.

aanil (approved):
Looks ok to me! I'm assuming [this](https://github.com/ScilifelabDataCentre/dds_web/issues/662) is the card related to changing all the calls to `dds_web.utils.current_time()`?

inaod568:
> Looks ok to me! I'm assuming [this](https://github.com/ScilifelabDataCentre/dds_web/issues/662) is the card related to changing all the calls to `dds_web.utils.current_time()`?

Yes, exactly!

PR #670, opened 2021-10-29
Update code to reflect previous refactoring

aanil:


dds_web/api/project.py
```diff
@@ -392,7 +390,9 @@ def post(self):
                 else:
                     # If it is an existing user, add them to project.
                     add_user_result = AddUser.add_user_to_project(
-                        existing_user=existing_user, project=new_project.public_id, owner=owner
+                        existing_user=existing_user,
+                        project=new_project.public_id,
+                        owner=user.get("role") == "Project Owner",
```
inaod568, line 395: Is the `add_user_to_project` code adapted to if the new added user has role Unit Admin or Personnel? I think there needs to be a few checks here before the user is added, and there needs to be a way of just adding the user to the unit if it's not a Researcher or Personnel.

aanil: I've added a check for 'Project Owner' and 'Researcher' when adding users to a project. I think that adding users to a Unit needs to be a different function and not mixed in with the project addition functionality.

inaod568: Great! Would be great if we could add some tests for when attempting to add Unit Personnel / Admins, to check that it fails.

aanil: Added a few tests!

inaod568: Nice! Thanks

tests/test_project_creation.py
```diff
@@ -399,3 +411,17 @@ def test_create_project_with_invited_users(client):
     assert response.json and response.json.get("user_addition_statuses")
     for x in response.json.get("user_addition_statuses"):
         assert "Invitation sent" in x
+
+
+def test_create_project_with_unsuitabl_(client):
```
inaod568, line 416: A little typo

tests/test_user_add.py
```diff
@@ -268,3 +268,18 @@ def test_existing_user_change_ownership_same_permissions(client):
...
+    response = client.post(
+        tests.DDSEndpoint.USER_ADD,
+        headers=tests.UserAuth(tests.USER_CREDENTIALS["unituser"]).post_headers(),
+        data=json.dumps(user_with_unsuitable_role),
+        content_type="application/json",
+    )
+    assert response.status == "403 FORBIDDEN"
+
+
+existing_research_user
```
inaod568, line 285: Remove the last line?

tests/test_project_creation.py
aanil: Oops!

tests/test_user_add.py
aanil: Probably a bit hasty in pushing the last commit 😅

aanil: Changed it now!

inaod568 (approved):

inaod568:
@aanil Looks great! Just waiting for the CodeQL to pass

PR #672, opened 2021-11-01
Remove `Project.size`

inaod568:
Instead of having the `size` column in the `Project` table which also complicates the code a bit since we have to update it when we upload and delete etc (also vulnerable to errors if some error occurs), I added `size` as a `@property` which when called computes the size from the files connected to the project. I also added the `num_files` property.


dds_web/database/models.py
```diff
@@ -124,6 +123,18 @@ def safespring_project(self):
 
         return self.responsible_unit.safespring
 
+    @property
+    def size(self):
+        """Calculate size of project."""
+
+        return sum([f.size_stored for f in self.files])
```
alneberg, line 130: I recall there might be a way to do the sum in SQL. There might be some magic in this line, but to me it looks like the sum is done in pure python. I think it's possible to use `func.sum`. This is an optimization though so it might not be needed.

alneberg (approved):
Looks good, there are some alternatives if we find this to not perform optimally, but let's wait with that optimisation.

dds_web/database/models.py
inaod568: Yeah, there is, but that means you have to perform a query and then anyway get the list of file ids in `self.files` I think. But we can definitely look at it later on.

Issue #671, opened 2021-11-01
Usage calculation returns inconsistent values


MatthiasZepper:
While working on #654 and experimenting with uploaded files, I noticed that the displayed values in the Usage column of `dds ls -u "unituser_1" --usage` are inconsistent and implausible. A project with 2.7 MB size had 100.00 MBh usage after a few minutes and the total usage returned was only 1MBh, a 1% fraction of these. 

So the total column was not the total of all projects and the usage was apparently not in agreement with the actual size of the files uploaded. This should be double-checked.

inaod568:
Yes, this is very strange... Looking into it now.

inaod568:
![Screenshot 2021-11-04 at 13 08 49](https://user-images.githubusercontent.com/35953392/140315455-0ebf7639-c418-4566-9a52-63b582c15445.png)

There seems to be a bug in both the CLI and the API.

MatthiasZepper:
Don't close this yet. While going through the code I realized that we have some duplicate code for this in the `dds_web/api/user.py` endpoint code and that the corresponding CLI code is also not yet up to date. 

In this duplicated code the same bug (division instead of multiplication) is present. We can either fix in place or try to write a generic function that works for both endpoints.

inaod568:
Status? Was this fixed somewhere or should it still be open? @MatthiasZepper

MatthiasZepper:
@inaod568: No, not fixed yet. I was planning to address this soon together with the name_in_bucket issue, since I am anyway having a look at the files endpoint. 

But need to learn the ropes first, because I am lacking familiarity with this part of the code.

MatthiasZepper:
If the `api.add_resource(user.ShowUsage, "/usage", endpoint="usage")` endpoint is anyway not needed at the moment, I guess there is no point in fixing it then there.

PR #687, opened 2021-11-04
Fixed the faulty GBhours calculation

MatthiasZepper:
Adresses #671:

- This PR fixes the faulty GBhours calculation. The _version size_ need to be multiplied with the _file hours_ and not divided by them. 
- Also changes to using Bytehours throughout the endpoint code and skips rounding early to avoid floating point errors.


inaod568 (approved):
Awesome, thanks for fixing this! Now it shows Kbhours etc 👏🏻

Issue #658, opened 2021-10-26
Secure site


inaod568:
I started looking at the Code-scanning alerts in the Security tab and it complained about (among other things) "Reflected server-side cross-site scripting". So I looked into it and got to the following page: 
https://flask.palletsprojects.com/en/2.0.x/security/
This, among other things, mentions the Flask-Talisman extension. 

At the very least we need to look into this and make sure we have covered the major vulnerabilities.

inaod568:
Very big. Closing and if needed opening smaller tasks.

PR #691, opened 2021-11-08
Incorporating minor code improvements suggested by CodeQL

MatthiasZepper:
This PR fixes the low-hanging fruits from the CodeQL bug report that shouldn't have any large-scale implications. All suggestions that require larger code changes are not yet addressed, so issue #658 is not yet completely resolved. 

Also modified some error classes to prevent returning the full error stack trace to the user. Instead, a generic message is returned by default unless the new option `pass_message=TRUE` is explicitly set.

### DIFF
--- a/dds_web/api/errors.py
+++ b/dds_web/api/errors.py
@@ -87,8 +87,17 @@ class AccessDeniedError(exceptions.HTTPException):
 class DatabaseError(exceptions.HTTPException):
     """Baseclass for database related issues."""
 
-    def __init__(self, message="The DDS encountered an Flask-SQLAlchemy issue.", project=None):
-        super().__init__(message)
+    def __init__(
+        self,
+        message="The DDS encountered an Flask-SQLAlchemy issue.",
+        pass_message=False,
+        project=None,
+    ):
+
+        if pass_message:
+            super().__init__(message)
+        else:
+            super().__init__("The system encountered an error in the database.")
 
         general_logger.warning(message)
 
@@ -128,8 +137,11 @@ class EmptyProjectException(exceptions.HTTPException):
 class DeletionError(exceptions.HTTPException):
     """Deletion of item failed."""
 
-    def __init__(self, project, message="Deletion failed."):
-        super().__init__(message)
+    def __init__(self, project, message="Deletion failed.", pass_message=False):
+        if pass_message:
+            super().__init__(message)
+        else:
+            super().__init__("Deletion of the file failed.")
 
         general_logger.warning(message)
 
@@ -202,8 +214,12 @@ class S3InfoNotFoundError(exceptions.HTTPException):
 class JwtTokenGenerationError(exceptions.HTTPException):
     """Errors when generating the JWT token during authentication."""
 
-    def __init__(self, message):
-        super().__init__(message)
+    def __init__(self, message, pass_message=False):
+
+        if pass_message:
+            super().__init__(message)
+        else:
+            super().__init__("Unrecoverable error during the authentication process. Aborting.")
 
         general_logger.warning(message)
 
@@ -238,9 +254,15 @@ class MissingMethodError(exceptions.HTTPException):
 class KeyNotFoundError(exceptions.HTTPException):
     """Key not found in database."""
 
-    def __init__(self, project, message="No key found for current project"):
+    def __init__(self, project, message="No key found for current project", pass_message=False):
         self.message = f"{message}: {project}"
-        super().__init__(self.message)
+
+        if pass_message:
+            super().__init__(self.message)
+        else:
+            super().__init__("Unrecoverable encryption error. Aborting.")
+
+        general_logger.warning(self.message)
 
 
 class InviteError(exceptions.HTTPException):
--- a/dds_web/api/errors.py
+++ b/dds_web/api/errors.py
@@ -217,7 +217,8 @@ class JwtTokenGenerationError(exceptions.HTTPException):
 
         super().__init__(
             "Unrecoverable error during the authentication process. Aborting." 
-            if not pass_message else message
+            if not pass_message 
+            else message
         )
 
 
--- a/dds_web/api/errors.py
+++ b/dds_web/api/errors.py
@@ -251,10 +251,7 @@ class KeyNotFoundError(exceptions.HTTPException):
 
         general_logger.warning(self.message)
 
-        if pass_message:
-            super().__init__(self.message)
-        else:
-            super().__init__("Unrecoverable encryption error. Aborting.")
+        super().__init__("Unrecoverable key error. Aborting." if not pass_message else message)
 
 
 class InviteError(exceptions.HTTPException):
--- a/dds_web/api/errors.py
+++ b/dds_web/api/errors.py
@@ -89,7 +89,7 @@ class DatabaseError(exceptions.HTTPException):
 
     def __init__(
         self,
-        message="",
+        message,
         pass_message=False,
         project=None,
     ):
@@ -106,10 +106,7 @@ class DatabaseError(exceptions.HTTPException):
             },
         )
 
-        if pass_message and message:
-            super().__init__(message)
-        else:
-            super().__init__("The system encountered an error in the database.")
+        super().__init__("The system encountered an error in the database." if not pass_message else message)
 
 
 class EmptyProjectException(exceptions.HTTPException):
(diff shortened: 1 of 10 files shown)


dds_web/api/errors.py
```diff
@@ -238,9 +254,15 @@ def __init__(self, message="No method found in request."):
...
+    def __init__(self, project, message="No key found for current project", pass_message=False):
         self.message = f"{message}: {project}"
-        super().__init__(self.message)
+
+        if pass_message:
+            super().__init__(self.message)
+        else:
+            super().__init__("Unrecoverable encryption error. Aborting.")
+
+        general_logger.warning(self.message)
```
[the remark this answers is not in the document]
inaod568: Also, the different exceptions are basically using the same code for checking the pass_message so I'm wondering if we should have a base custom exception with that code in there, and then the rest inheriting from that class?

dds_web/api/errors.py
```diff
@@ -87,8 +87,17 @@ def __init__(
...
+        self,
+        message="The DDS encountered an Flask-SQLAlchemy issue.",
+        pass_message=False,
+        project=None,
+    ):
+
+        if pass_message:
+            super().__init__(message)
+        else:
+            super().__init__("The system encountered an error in the database.")
```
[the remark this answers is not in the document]
MatthiasZepper: The `pass_message` parameter was newly introduced by me and I have not used it at any error that is being raised somewhere in the code so far, so your question can be answered with _"No"_. 

The rationale was that it is by default not desirable to expose tracebacks of database errors to the users because this could expose sensitive data to the users and also be quite confusing. To set `pass_message=False` was an easy way of achieving that without changing the invocation of the single errors. However, by specifying this parameter as `True`, one can deliberately allow an error message to be seen by the user in the future.

The default message "_The DDS encountered an Flask-SQLAlchemy issue._" is indeed somewhat dispensable by now.  I maybe was a bit too conservative here...should I make this the default message to return? I somehow felt it would be too technical for the average user to understand...

dds_web/api/errors.py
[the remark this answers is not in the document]
MatthiasZepper: I only introduced this parameter for selected errors, where I felt it would be specifically inappropriate to expose the details of the error to the users. So the majority of the errors doesn't need that code. 

I am not strictly opposed to introducing a second superclass to handle that part, but in relation to writing this simple if-statement out for a few times, it didn't really seem worth the effort to me. I guess there is little gain in terms of preventable code duplication compared to the complexity of juggling between two superclasses the errors will then inherit from. Unless we decide that all errors should have that parameter, I would say it is unnecessary perfectionism at this point?

dds_web/api/errors.py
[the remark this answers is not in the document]
inaod568: Yes, I like the `pass_message` solution. 

No I actually meant to perhaps replace the default message and set it (in this case) to "The system encountered an error in the database"

dds_web/api/errors.py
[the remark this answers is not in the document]
inaod568: Agreed. Skip this for now 👍🏻

dds_web/api/errors.py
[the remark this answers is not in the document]
MatthiasZepper: However, the predominant invocation of the DatabaseError looks like this:

```
        except sqlalchemy.exc.SQLAlchemyError as err:
            raise DatabaseError(str(err))
```

This currently logs `err` and returns "_The system encountered an error in the database_" to the user.  

The contents of the default message would only matter if the error was invoked like so:

```
        except sqlalchemy.exc.SQLAlchemyError as err:
            raise DatabaseError(pass_message=True)
```
Passing on a non-existing message however seems counter-intuitive to me, but it could the used to show the user the default message instead of the one specified in `super().__init__()`. I however believe that there is no real use case for this kind of behavior. 

A more likely thing might be this case:

```
        except sqlalchemy.exc.SQLAlchemyError as err:
            raise DatabaseError
```
The user sees "_The system encountered an error in the database_" and the systems logs the default message "_The DDS encountered an Flask-SQLAlchemy issue._". 


But if you like, I can of course get rid of the _The DDS encountered an Flask-SQLAlchemy issue_ statement in the code.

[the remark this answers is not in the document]
inaod568: Yes, in my opinion it doesn't make sense to show and log two different vague messages. If we were to just have `super().__init__(message)` we could simply have `message="The system encountered an error in the database."` where it says `message="The DDS encountered an Flask-SQLAlchemy issue."`, right? At least to me this makes sense regarding this class.

dds_web/api/errors.py
```diff
@@ -87,8 +87,19 @@ def __init__(
...
+        pass_message=False,
+        project=None,
+    ):
+
+        if pass_message and message != "":
+            super().__init__(message)
+        else:
+            super().__init__("The system encountered an error in the database.")
+            if message != "":
+                message = super.message
```
inaod568, line 102: Actually what I meant was this:
```
 def __init__(
        self,
        message="The system encountered an error in the database.",
        project=None,
    ):

        super().__init__(message)
```

Would that not be ok?

dds_web/api/errors.py
```diff
@@ -128,8 +139,11 @@ def __init__(self, project, username=None, message="The project is empty."):
 class DeletionError(exceptions.HTTPException):
     """Deletion of item failed."""
 
-    def __init__(self, project, message="Deletion failed."):
-        super().__init__(message)
+    def __init__(self, project, message="Deletion failed.", pass_message=False):
+        if pass_message:
+            super().__init__(message)
+        else:
+            super().__init__("Deletion of the file failed.")
```
inaod568, line 146: Same here:
```
    def __init__(self, project, message="Deletion failed."):
        super().__init__(message)
```

dds_web/api/errors.py
```diff
@@ -238,9 +256,15 @@ def __init__(self, message="No method found in request."):
 class KeyNotFoundError(exceptions.HTTPException):
     """Key not found in database."""
 
-    def __init__(self, project, message="No key found for current project"):
+    def __init__(self, project, message="No key found for current project", pass_message=False):
```
inaod568, line 259: And same here

dds_web/api/errors.py
MatthiasZepper: This is essentially how it looked like before I started to change it. 

But that would result in the error message to be shown to the user or am I missing something? I assumed that, because the superclass is the HTTP.exception, whatever is passed as message to  `super().__init__(message) ` will be returned to the user. 

Wouldn't we need to take care of logging the database error within the `except` statements that raise the error in that case and not pass anything to the error itself?  

```
       except sqlalchemy.exc.SQLAlchemyError as err:
            raise DatabaseError(str(err))
```

inaod568: Yes in that case we would have not pass the info in when we raise. But ok, then how about:
```
    def __init__(
        self,
        message,
        pass_message=False,
        project=None,
    ):

        if pass_message and message:
            super().__init__(message)
        else:
            super().__init__("The system encountered an error in the database.")
```
I don't understand the 
```
            if message != "":
                message = super.message
```
part though.

MatthiasZepper: Yes, that was a mistake. I wanted to make sure that there is a non-empty message to log without typing out the default message twice. 

So I get it back from `super()` after setting it there. I also wasn't aware that in Python an empty string and not only _NULL_ evaluates to False - this somewhat simplifies the `if` statement.

inaod568: Ah okey. Then 
```
    def __init__(
        self,
        message,
        pass_message=False,
        project=None,
    ):

        if not (pass_message and message):
            message = "The system encountered an error in the database."
        
        super().__init__(message)
        general_logger.warning(message)
```

alneberg: Sorry for jumping in here but I think the point was to also be able to log the real error even though we send the vague message to the user?

inaod568: Oh ok! I didn't understand that that was what we were going for here. In that case I would probably log it first and then check if pass_message and then raise it with super().__init

inaod568: @MatthiasZepper

inaod568:
@MatthiasZepper I'm waiting to approve this until the conflict is and comments have been resolved

inaod568 (approved):

PR #692, opened 2021-11-08
project key generation updated

inaod568:


alneberg (approved):
Works and looks much better!

PR #696, opened 2021-11-11
Encrypted token and authentication tests

### LINKED ISSUE #695: Enable the encrypted token endpoint

alneberg:
- Enabled the encrypted token endpoint (all code needed was already in place)
- Added tests for token authentication:
  -  malformed tokens
  - expired tokens
  - tokens using the wrong secret key
- The same for encrypted tokens
- Added error handling for the verify_token method to catch errors raised by tests above
- As a bonus, added possibility to supply arguments to pytest to enable running only a few tests which is really useful during development.

This PR should fix #695.


worukan (approved):
Looks good.

inaod568 (approved):
Looks good 👏🏻

PR #708, opened 2021-11-17
Make db queries consistent

inaod568:
I have changed all `db.session.query` to `model.<table name>.query`. Flask-SQLAlchemy and SQLAlchemy in general handles the sessions. They also should not be vulnerable to SQL injections etc since they sanitize the inputs before executing. However, in the case of `f`-strings in queries, we need to have a look at if this actually introduces some vulnerabilities. I haven't been able to find a good solution so far, any help appreciated.


alneberg (approved):
All good from what I can see.

PR #710, opened 2021-11-18
Implement 2fa

inaod568:
Have implemented 2FA. It should later be possible to request a renewal of the `otp_secret`, but currently it should just show the qrcode and secret once and then never again --> can only be setup once for a specific secret. 
Steps to test: 
1. `dds add-user`
2. Go to email and click on invite link
3. Create account
4. Setup 2FA

_**Note: This setup will change once we introduce the mail confirmation steps as well.**_


dds_web/web/user.py
```diff
@@ -0,0 +1,223 @@
...
+
+        # User has now setup 2FA
+        flask_login.current_user.has_2fa = True
+        try:
+            db.session.commit()
+        except sqlalchemy.exc.SQLAlchemyError as sqlerr:
+            raise ddserr.DatabaseError from sqlerr
+        return flask.redirect(flask.url_for("auth_blueprint.index"))
+    else:
+        flask.flash("You have supplied an invalid 2FA token!", "danger")
```
alneberg, line 222: I didn't see this flash when submitting the wrong code...

dds_web/web/user.py
```diff
@@ -0,0 +1,223 @@
...
+
+    return flask.redirect(flask.url_for("auth_blueprint.index"))
+
+
+@auth_blueprint.route("/twofactor", methods=["GET"])
+def two_factor_setup():
+    """Setup two factor authentication."""
+    # since this page contains the sensitive qrcode, make sure the browser
+    # does not cache it
+    flask_login.current_user.set_2fa_seen()
```
alneberg, line 174: This is never checked at the moment right? And also, I'm wondering how strict we should be. This metod is named 'set_2fa_seen'  but it sets 'has_2fa'. I'm thinking there might be a difference between seeing the token and actually having it set up. We can perhaps discuss this more in a meeting.

inaod568: So the idea is that the secret should only be seen once, because otherwise we can't guarantee that it's the correct user that's actually using the 2fa authentication later on. So it should be seen once, and then never again. If they would need to redo this for some reason there should be another new secret instead, but how this would work and if we need some checks or verification etc, I'm not sure. 

Regarding the `set_2fa_seen` --> not entirely sure why I decided to it like that, probably thought it looked better than `flask_login.current_user.has_2fa = True` and it gives more power to the db to handle it.

inaod568: What do you mean with checked?

inaod568: Ok maybe I'll remove this for now. I anyway have it in two places ish so I think I just missed something somewhere. Will remove this part and it will be (for now) that they can only set it up once.

alneberg: with 'checked' I meant that you set it True but we should perhaps check when the request is made whether it's already set or not?

dds_web/templates/user/two-factor-setup.html
```diff
@@ -0,0 +1,84 @@
...
+            </div>
+            {% if error %}
+            <p class=error><strong>Error:</strong> {{ error }}
+                {% endif %}
+            </form>
+        </div>
+
+    </div>
+    {% else %}
+    Hellu
```
alneberg, line 64: 😂 Que?

inaod568: 🤣

inaod568: Jaa den skulle ju inte vara där, skulle bara testa en sak 😂

dds_web/__init__.py
```diff
@@ -17,6 +17,12 @@
 from authlib.integrations import flask_client as auth_flask_client
 from flask_httpauth import HTTPBasicAuth, HTTPTokenAuth
 import flask_mail
+import flask_security
+import passlib
```
alneberg, line 21: I believe flask_security and passlib is not used?

dds_web/__init__.py
```diff
@@ -146,6 +150,14 @@ def create_app(testing=False, database_uri=None):
     # Initialize marshmallows
     ma.init_app(app)
 
+    # Initialize login manager
+    login_manager.init_app(app)
+
+    @login_manager.user_loader
+    def load_user(user_id):
+        # since the user_id is just the primary key of our user table, use it in the query for the user
+        return models.User.query.get(user_id)
```
alneberg, line 159: Do you think we need some SQLAlchemy Error handling here? Feel free to create a card about though.

inaod568: True, I can add that now.

inaod568: @alneberg Have added an extremely simple error handler above this with a comment on what more is needed there.

dds_web/database/models.py
```diff
@@ -174,45 +178,100 @@ class User(db.Model):
...
+        """Get uri for user otp_secret if the 2fa has not already been setup."""
+        if self.has_2fa:
+            return None
+        return pyotp.totp.TOTP(self.otp_secret).provisioning_uri(
+            name=self.username, issuer_name="Data Delivery System"
+        )
+
+    def verify_totp(self, token):
+        """Verify the otp token."""
+        totp = pyotp.TOTP(self.otp_secret)
```
alneberg, line 265: Should this be `self._otp_secret`?

inaod568: No, `otp_secret` is the property which also checks if they have setup the 2fa and in that case does not display the secret again

dds_web/database/models.py
```diff
@@ -174,45 +178,100 @@ class User(db.Model):
...
+        return totp.verify(token)
```
alneberg, line 266: I think there is an option to make this slightly more allowing (basically help for slow people like myself): `topt.verify(token, valid_window=1)`. This way it will be verified to True even though the timer has just flipped to the next interval.

dds_web/run_app.py
```diff
@@ -10,6 +10,8 @@
 
 # Own modules
 from dds_web import create_app
+from dds_web import auth
```
alneberg, line 13: Don't think you need to add this here?

dds_web/database/models.py
alneberg: yes, but don't you want to use the verify_totp method every time you log in using 2fa?

inaod568: Yes, but this is another card. This PR just adds the setup of the 2FA, not the login with it.

inaod568: We probably need to change some things but then other changes are also needed so this is just a first, the actual setup of 2FA. Logging in or getting access to projects with 2FA requires us to have added the checks for sensitive projects and probably the statuses too.

dds_web/database/models.py
inaod568: Oh, I didn't even think of that. Sure, sounds good!

dds_web/database/models.py
alneberg: Fair enough we can fix it later.

inaod568:
Closing this in favor of a new PR

Issue #638, opened 2021-10-19
Add better message to `ConfirmInvite` endpoint if invite already approved


inaod568:
If the user has already approved the invite the endpoint should say that, not only that there is no invite for the user -- they already have an account and the link is used.

https://github.com/ScilifelabDataCentre/dds_web/blob/d6de61c94ce01e257dc3a9ea9177f5134bf5a7dd/dds_web/api/user.py#L127-L129

PR #714, opened 2021-11-19
Added a custom HTML page that will be shown for already registered users

MatthiasZepper:
This PR addresses #638. 

A new check is introduced in the `dds_web.api.user.ConfirmInvite.get()`. In case the invite link has not yet expired, but the system nonetheless can't find an invitation, it will check if the user has already been signed up. 

In that case the user is redirected to a HTML page that will point to the documentation in the future or even summarize the most important steps directly.


dds_web/api/user.py
```diff
@@ -310,9 +310,20 @@ def get(self, token):
 
         # Check the invite exists
         if not invite_row:
-            raise ddserr.InviteError(
-                message=f"There is no invitation for the found email adress: {email}"
-            )
+
+            try:
+                # check if the user has already registered at the system
+                already_registered = models.Email.query.filter(models.Email.email == email).first()
```
inaod568, line 316: There's a function called `email_in_db` in `user_schemas.py` - use that here instead to keep the consistency between the different endpoints and schemas.

inaod568 (changes requested):

inaod568 (approved):
Nice! I like it!

PR #716, opened 2021-11-22
Change in s3 info retrieval

inaod568:
Moved the info from `s3_config.json` into the `Unit` table instead. This also entailed rewriting (and simplifying) the retrieval of s3 information, such as the safespring access keys. How these are handled is not a part of this PR, this is just removing the use of the `s3_config.json` file.


worukan (approved):
Nice rewrite and clean up! I see that there are a few comments, e.g. unique=true in models and your comment (in pr) hinting that there will be follow-up work in the future. The current changes looks good to me.

ewels (approved):
Nice!  I'm all for having fewer places to set configuration options 👍🏻

PR #717, opened 2021-11-22
First step in creating a project status framework

aanil:
* Added table `projectstatuses` with columns `project_id`, `status` and `date_created` to the db models.
* Updated the Project model with the properties `current_status` and `has_been_available`.
* Added new endpoint `/proj/status` to display and update project Statuses.
* Added checks in file upload, download and deletion. 
* Added some tests.

### DIFF
--- a/dds_web/api/files.py
+++ b/dds_web/api/files.py
@@ -34,7 +34,7 @@ from dds_web.api.schemas import project_schemas
 def check_eligibility_for_upload(status):
     """Check if a project status is eligible for upload/modification"""
     if status != "In Progress":
-        raise DDSArgumentError("Project not in right status to upload/modify files")
+        raise DDSArgumentError("Project not in right status to upload/modify files.")
     return True
 
 
@@ -43,16 +43,21 @@ def check_eligibility_for_download(status, user_role):
     if status != "Available":
         if status == "In Progress":
             if user_role not in ["Unit Admin", "Unit Personnel"]:
-                raise DDSArgumentError("Current Project status limits file download")
+                raise DDSArgumentError("Current Project status limits file download.")
         else:
-            raise DDSArgumentError("Current Project status limits file download")
+            raise DDSArgumentError("Current Project status limits file download.")
     return True
 
 
 def check_eligibility_for_deletion(status, has_been_available):
     """Check if a project status is eligible for download"""
-    if status != "In Progress" or has_been_available:
-        raise DDSArgumentError("Project Status/Status history prevents files from being deleted")
+    if status != "In Progress":
+        raise DDSArgumentError("Project Status prevents files from being deleted.")
+    else:
+        if has_been_available:
+            raise DDSArgumentError(
+                "Existing project contents cannot be deleted since the project has been previously made available to recipients."
+            )
     return True
 
 
@@ -169,6 +174,8 @@ class MatchFiles(flask_restful.Resource):
         """Matches specified files to files in db."""
 
         project = project_schemas.ProjectRequiredSchema().load(flask.request.args)
+
+        check_eligibility_for_upload(project.current_status)
 
         try:
             matching_files = (
--- a/dds_web/api/files.py
+++ b/dds_web/api/files.py
@@ -49,7 +49,7 @@ def check_eligibility_for_download(status, user_role):
 
 
 def check_eligibility_for_deletion(status, has_been_available):
-    """Check if a project status is eligible for download"""
+    """Check if a project status is eligible for deletion"""
     if status != "In Progress":
         raise DDSArgumentError("Project Status prevents files from being deleted.")
     else:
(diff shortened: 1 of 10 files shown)


dds_web/api/files.py
```diff
@@ -30,6 +30,32 @@
...
+                raise DDSArgumentError("Current Project status limits file download")
+        else:
+            raise DDSArgumentError("Current Project status limits file download")
+    return True
+
+
+def check_eligibility_for_deletion(status, has_been_available):
+    """Check if a project status is eligible for download"""
+    if status != "In Progress" or has_been_available:
+        raise DDSArgumentError("Project Status/Status history prevents files from being deleted")
```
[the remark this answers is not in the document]
inaod568: Just to give the user more information

dds_web/api/project.py
```diff
@@ -149,6 +149,75 @@ def create_project(self, data, **kwargs):
...
+
+    def is_transition_possible(self, current_status, new_status):
+        """Check if the transition is valid"""
+        possible_transitions = [
+            ("In Progress", ["Available", "Deleted"]),
+            ("Available", ["In Progress", "Expired", "Archived"]),
+            ("Expired", ["Available", "Archived"]),
+        ]
+        result = False
+        if current_status != new_status:
```
inaod568, line 215: Maybe we could add something like this
```
if current_status == new_status: 
    raise DDSArgumentError(f"The project status is already {new_status}")
```
and then underneath
```
for transition in possible_transitions:
    .... etc
```
Thoughts?

dds_web/database/models.py
```diff
@@ -10,6 +10,7 @@
 # Installed
 from sqlalchemy.ext import hybrid
 import sqlalchemy
+from sqlalchemy.orm import column_property
```
inaod568, line 13: Unused I think

dds_web/api/project.py
aanil: Maybe we don't need the check at all as it is not defined in `possible_transitions`? It will then then flag it as invalid and raise `DDSArgumentError("Invalid status transition")` in the calling function.

dds_web/api/files.py
```diff
@@ -30,6 +30,32 @@
...
+
+def check_eligibility_for_download(status, user_role):
+    """Check if a project status makes it eligible to download"""
+    if status != "Available":
+        if status == "In Progress":
+            if user_role not in ["Unit Admin", "Unit Personnel"]:
+                raise DDSArgumentError("Current Project status limits file download")
+        else:
+            raise DDSArgumentError("Current Project status limits file download")
+    return True
```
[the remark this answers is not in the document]
aanil: Should the Unit personnel be able to download if the project is 'Available'?

dds_web/api/files.py
```diff
@@ -30,6 +30,32 @@
 from dds_web.api.schemas import file_schemas
 from dds_web.api.schemas import project_schemas
 
+
+def check_eligibility_for_upload(status):
```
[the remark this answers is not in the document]
aanil: Done!

dds_web/api/files.py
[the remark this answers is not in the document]
aanil: Done!

dds_web/database/models.py
aanil: Removed!

dds_web/api/files.py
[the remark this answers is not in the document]
inaod568: Ah ok. Right.. It just feels like it should be possible to simplify this a bit, since both are raising the same exception.

dds_web/api/project.py
inaod568: True, do that instead!

dds_web/api/files.py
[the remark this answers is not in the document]
aanil: Changed it a bit now.

dds_web/api/files.py
```diff
@@ -30,6 +30,36 @@
...
+
+
+def check_eligibility_for_download(status, user_role):
+    """Check if a project status makes it eligible to download"""
+    if status == "Available":
+        return True
+    if (status == "In Progress") and (user_role in ["Unit Admin", "Unit Personnel"]):
+        return True
+
+    raise DDSArgumentError("Current Project status limits file download.")
```
inaod568, line 48: Better! Can we do 
```
if status == "Available" or (status == "In Progress" and user_role in ["Unit Admin", "Unit Personnel"]): 
    return True
```

inaod568: It's not a deal breaker though, just a personal preference!

aanil: Done.

inaod568: Great, thanks!

dds_web/api/files.py
```diff
@@ -30,6 +30,36 @@
...
+
+
+def check_eligibility_for_deletion(status, has_been_available):
+    """Check if a project status is eligible for deletion"""
+    if status != "In Progress":
+        raise DDSArgumentError("Project Status prevents files from being deleted.")
+    else:
```
inaod568, line 55: Tiny note: `else` is not needed here

inaod568 (approved):
Tested and seems to work! Looks good! Had a tiny comment but nothing you need to change.

PR #718, opened 2021-11-22
Fix implement 2fa

inaod568:
Same as https://github.com/ScilifelabDataCentre/dds_web/pull/710 before something messed it up.


alneberg (approved):
Works great! I suspect a lot of it will be rewritten at some point. I'll try to create a few cards for the things I have in mind.

inaod568:
> Works great! I suspect a lot of it will be rewritten at some point. I'll try to create a few cards for the things I have in mind.

Great! Yep, a lot will most likely be rewritten but I like the idea of adding these things step by step and see what we need as we go. 
Yes, would be great if you could create cards, thanks.

PR #722, opened 2021-11-29
Password reset

inaod568:
Added the functionality of resetting passwords.

### DIFF
--- a/dds_web/forms.py
+++ b/dds_web/forms.py
@@ -12,6 +12,8 @@ import marshmallow
 
 # Own modules
 import dds_web.utils
+from dds_web.database import models
+
 
 # FORMS #################################################################################### FORMS #
 
@@ -42,7 +44,9 @@ class RegistrationForm(flask_wtf.FlaskForm):
 
     name = wtforms.StringField("name", validators=[wtforms.validators.InputRequired()])
     email = wtforms.StringField(
-        "email", validators=[wtforms.validators.Email()], render_kw={"readonly": True}
+        "email",
+        validators=[wtforms.validators.DataRequired(), wtforms.validators.Email()],
+        render_kw={"readonly": True},
     )
     username = wtforms.StringField(
         "username",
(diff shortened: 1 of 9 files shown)


dds_web/forms.py
```diff
@@ -62,6 +66,20 @@ class RegistrationForm(flask_wtf.FlaskForm):
     confirm = wtforms.PasswordField("Repeat password")
     submit = wtforms.SubmitField("submit")
 
+    def validate_username(self, username):
+        user = models.User.query.filter_by(username=username.data).first()
+        if user:
+            raise wtforms.validators.ValidationError(
+                "That username is taken. Please choose a different one."
+            )
```
alneberg, line 74: There's some characters not allowed in usernames, for example `.`, should we check that already in the WTForms? I believe that would give a better user experience since otherwise it would fail on user creation?

dds_web/forms.py
```diff
@@ -62,6 +66,20 @@ class RegistrationForm(flask_wtf.FlaskForm):
...
+
+    def validate_email(self, email):
+        email = models.Email.query.filter_by(email=email.data).first()
+        if email:
+            raise wtforms.validators.ValidationError(
+                "That email is taken. Please choose a different one."
```
alneberg, line 80: Can you even choose the email at signup?

dds_web/forms.py
inaod568: Yes, I will add that check, thanks.

dds_web/forms.py
inaod568: No, it's not suppose to, but it's good to have this check here anyway.

dds_web/utils.py
```diff
@@ -22,17 +22,36 @@
...
 
 
+def send_reset_email(email_row):
+    """Generate password reset email."""
+    # Generate token
+    token = email_row.user.get_reset_token()
+
+    # Create and send email
+    message = flask_mail.Message(
+        "Password Reset Request", sender="noreply@mailtrap.io", recipients=[email_row.email]
```
alneberg, line 45: We probably shouldn't have mailtrap hardcoded here?

alneberg (approved):
Looks good! Only a few minor comments

dds_web/forms.py
```diff
@@ -76,10 +73,47 @@ class LogoutForm(flask_wtf.FlaskForm):
...
+
+
+class ResetPasswordForm(flask_wtf.FlaskForm):
+    """Form for setting a new password."""
+
+    password = wtforms.PasswordField(
+        "Password",
+        validators=[
+            wtforms.validators.DataRequired(),
+            wtforms.validators.EqualTo("confirm", message="Passwords must match!"),
```
alneberg, line 107: `"confirm"` here leads to an error when attempting to reset password since the field is still called `confirm_password`. Change either here or the field and I think it should work

alneberg:
Just a tiny bug that needs fixing but besides that I'm happy to merge.

PR #723, opened 2021-12-01
Add more statuses

aanil:
* Added statuses Deleted, Expired and Archived
* Added `is_aborted` and `deadline` to ProjectStatuses table. `deadline` will be used for both Available and Expired statuses. We need to decide if we need to set a Unit wide `deadline` default for Expired. The deadline is set to a second before midnight of the day it is set to.


inaod568:
@aanil Unit wide `deadline` default for `Expired` - yes

dds_web/api/project.py
```diff
@@ -183,20 +183,69 @@ def post(self):
...
+                ) + datetime.timedelta(days=deadline)
+
+            if not project.has_been_available:
+                project.released = curr_date
+                add_deadline["deadline"] = dds_web.utils.current_time(
+                    to_midnight=True
+                ) + datetime.timedelta(days=deadline)
+
+        # Moving to Expired
+        if new_status == "Expired":
```
inaod568, line 215: This should be an automatic thing, probably not changable by the actual users. But we can have this here for now anyway, until we create the scheduler.

dds_web/api/project.py
```diff
@@ -183,20 +183,69 @@ def post(self):
...
+            **{"project_id": project.id, "status": new_status, "date_created": curr_date},
+            **add_deadline,
+            **add_aborted,
+        )
+        try:
             project.project_statuses.append(add_status)
             db.session.commit()
+            if new_status in ["Deleted", "Archived"]:
+                # TODO Call function to delete files
+                print("TODO")
```
inaod568, line 245: Prints don't show up for me but maybe you can change this to `print("TODO", flush=True)` or/and logging it?

aanil: They will be replaced with function calls in the future, but I have added `flush=True` for now.

aanil:
> @aanil Unit wide `deadline` default for `Expired` - yes

Changed `days_to_expire` to `days_in_available` and added `days_in_expired` in `Unit`

dds_web/api/project.py
aanil: I agree that it should be an automatic thing in the usual course but it'll be nice to have a manual option just in case.

inaod568: Ok! I'm not sure I see why unit admins / personnel should be able to do this though, so in that case I'd limit it to `Super Admins`

dds_web/api/project.py
inaod568: Yeah, I understood that 👍🏻 Just want it to be actually printed so that it's easier to notice that something else needs to be added there

inaod568 (approved):
Looks good! 👏🏻

Issue #703, opened 2021-11-12
Delete user / user info


inaod568:
![image.png](https://images.zenhubusercontent.com/60190598ba808fccea954870/787fac29-33ed-4be5-825c-8422e6882595)

MatthiasZepper:
In the user roles we currently list _Remove themselves_ permission for every role. But I guess this refers to the project associations?

Should any user be able to delete their own DDS user account? Or should only unit users be able to delete accounts? Do we want to have some kind of approval system, e.g. an e-mail which contains a link that needs to be clicked to confirm the deletion or a 2FA token?

PR #725, opened 2021-12-01
Issue703: Delete user endpoint

MatthiasZepper:
This PR addresses #703:

- Adds two new API endpoints `DeleteUser` and `DeleteUserSelf` as restful resources. Also a new route to allow for the deletion of user accounts via confirmation links is added.
- This deletion request is then processed by the downstream function in `db-connector` . 
-  In case of a self deletion request, which can be issued by any user, an e-mail containing a confirmation link is send to the primary email address.
-  In case of a foreign deletion, permissions are checked. A Super Admin can delete everyone, a Unit Admin only Unitusers of the own unit. Self-deletion via this endpoint is not possible.  

Unfortunately commit history couldn't be reasonably retained during rebasing due to comprehensive conflicts, but major parts of this have been written by @alneberg!


dds_web/api/user.py
```diff
@@ -279,6 +280,146 @@ def add_user_to_project(existing_user, project, role):
...
+                message=f"Creation of self-deletion request failed due to database error: {sqlerr}",
+                pass_message=False,
+            )
+
+        # Create link for deletion request email
+        link = flask.url_for("auth_blueprint.confirm_self_deletion", token=token, _external=True)
+        subject = f"Confirm deletion of your user account {username} in the SciLifeLab Data Delivery System"
+        projectnames = "; ".join(proj_ids)
+
+        flask.current_app.logger.debug(f"Self-Deletion link for {username}: " + link)
```
alneberg, line 332: This should probably be removed

dds_web/database/models.py
```diff
@@ -274,8 +274,11 @@ class User(flask_login.UserMixin, db.Model):
 
     # Relationships
     identifiers = db.relationship("Identifier", back_populates="user", passive_deletes=True)
-    emails = db.relationship("Email", back_populates="user", passive_deletes=True)
+    emails = db.relationship("Email", back_populates="user", passive_deletes=True, cascade="all")
```
alneberg, line 277: I think this was not needed after all, or am I confusing it with a different relationship?

dds_web/utils.py
```diff
@@ -193,6 +193,19 @@ def username_in_db(username):
     return False
 
 
+def email_return_user(email):
```
alneberg, line 196: I would probably rename this function to `user_from_email` or similar. But that's probably personal preference.

dds_web/web/user.py
```diff
@@ -6,15 +6,18 @@
...
 # Installed
 import flask
+from dds_web.api.db_connector import DBConnector
 import flask_login
 import pyqrcode
 import pyotp
 import itsdangerous
 import sqlalchemy
 import marshmallow
+import logging
```
alneberg, line 20: This import is not needed I think?

tests/test_user_delete.py
```diff
@@ -0,0 +1,206 @@
...
+        tests.DDSEndpoint.USER_DELETE,
+        headers=tests.UserAuth(tests.USER_CREDENTIALS["delete_me_unitadmin"]).token(client),
+        data=json.dumps({"email": email_to_delete}),
+        content_type="application/json",
+    )
+    assert response.status_code == http.HTTPStatus.OK
+
+    # Make sure that user was deleted
+    exists = utils.email_return_user(email_to_delete)
+    assert exists is None
```
alneberg, line 188: Would be nice to add the case we struggled with here, namely that the user is deleted but not the email related to that user.

dds_web/database/models.py
inaod568: No you're remembering correctly. This is not needed.

dds_web/utils.py
inaod568: This function is not needed at all actually, this schema already does this.
https://github.com/ScilifelabDataCentre/dds_web/blob/bba1ebf390439ce6bd1d0e39e5eabc1887fb7323/dds_web/api/schemas/user_schemas.py#L28-L44 

Just call `user_schemas.UserSchema().load({"email": email})` where ever you need to do this.

alneberg: Maybe, it's used in tests though. I would feel awkward calling the schemas in the tests just to do this query...

inaod568: But why add a new function in `dds_web/utils.py` just for tests? Personally I'd rather use something that was already in the web code.

inaod568: If we want to use it in the tests and don't want to use the UserSchema then we should move this function to a `utils.py` module in the tests directory

MatthiasZepper: Yes, I created that function yesterday with regard to the tests and also pondered where I should put it. Indeed, it is currently not used in the main code, but while writing the initial code for the endpoints, I repeatedly wanted to have such a function badly. 

Arguably, this was due to bad design practices, but is there really no possible use (outside of tests) for a function that returns the user for a given email-address without performing de-serialization at the same time like the schema does? 

So as an alternative, I would propose to add it as static method to the Email db.model class then, such that one can not only return the `User.primary_email` but also the `Email.user` ?

dds_web/web/user.py
MatthiasZepper: According to VS Code, also `flask_mail` and `pyotp` are not used. Should I delete them as well?

dds_web/utils.py
inaod568: I'm not sure I follow. 

The UserSchema returns a User object by first getting the row in the Email table which matches the email address. The `email_return_user` also returns a User object but with a join. You don't need this though, instead you could just do
```
def email_return_user(email):
    """Helper function to return the User for a given email"""
    email = models.Email.query.filter_by(email=email).one_or_none()
    return email.user
```
The `Email.user` as you suggest is already available: https://github.com/ScilifelabDataCentre/dds_web/blob/bba1ebf390439ce6bd1d0e39e5eabc1887fb7323/dds_web/database/models.py#L566
(This is what the schema uses.)

inaod568: So I would rather just use the schema than introduce a new function, but if no one agrees, then I'm fine with having the function (it's not a deal breaker) but in that case I want the function to be moved to `dds_web/tests/utils.py` 👍🏻

MatthiasZepper: > The `UserSchema` returns a `User` object by first getting the row in the `Email` table which matches the email address. The `email_return_user` also returns a `User` object but with a join. You don't need this though, instead you could just do
> 
> ```
> def email_return_user(email):
>     """Helper function to return the User for a given email"""
>     email = models.Email.query.filter_by(email=email).one_or_none()
>     return email.user
> ```

I am pretty sure I tried exactly this with no avail yesterday in the tests, because after trying out this and about 5 different versions derived from StackOverflow, I finally found one that worked (the one using the join). 

**But anyway: I moved it now to the tests!**

inaod568 (approved):
Looks good! Approved.

PR #726, opened 2021-12-02
Fix foreign key constraints

inaod568:
- I changed my mind regarding the implicit/explicit relationships: Now we have them explicitly defined and placed in connection to the corresponding foreign key for clarity.
- I also went through the different foreign key contraints and added `ondelete=CASCADE` if the child row should be deleted on parent deletion, and `ondelete=SET NULL` if the column should just be filled with NULL on deleton of the parent, e.g. for `created_by`. I had an idea of setting a value instead but I didn't find a solution for that and this works well now.

### DIFF
--- a/dds_web/database/models.py
+++ b/dds_web/database/models.py
@@ -35,7 +35,9 @@ class ProjectUsers(db.Model):
     __tablename__ = "projectusers"
 
     # Primary keys / Foreign keys
-    project_id = db.Column(db.Integer, db.ForeignKey("projects.id"), primary_key=True)
+    project_id = db.Column(
+        db.Integer, db.ForeignKey("projects.id", ondelete="CASCADE"), primary_key=True
+    )
     user_id = db.Column(
         db.String(50), db.ForeignKey("researchusers.username", ondelete="CASCADE"), primary_key=True
     )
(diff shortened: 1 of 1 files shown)


alneberg:
Would be quite nice with tests for this? It's very critical parts and I would hesitate to delete things without knowing the actual behaviour? If you want I can have a go at writing the tests?

inaod568:
> Would be quite nice with tests for this? It's very critical parts and I would hesitate to delete things without knowing the actual behaviour? If you want I can have a go at writing the tests?

What do you mean with deleting things? 
Sure, it would be great if you could add some tests 👍🏻

alneberg:
No, the comment was not about things you delete in this PR. It was about the future case where we actually need to use this code when for example deleting a unit. It would be quite scary to do it, any tiny mistake in these cascades can have quite bad consequences. I would be especially worried about things we perform very rarely as they might be neglected in testing and not used for a few years.

inaod568:
> No, the comment was not about things you delete in this PR. It was about the future case where we actually need to use this code when for example deleting a unit. It would be quite scary to do it, any tiny mistake in these cascades can have quite bad consequences. I would be especially worried about things we perform very rarely as they might be neglected in testing and not used for a few years.

@alneberg Agreed.

inaod568:
@alneberg An alternative is to not be able to delete units or projects at all (unless all the foreign keys are deleted first), just to set stuff to null. Do we want that instead?

alneberg:
No, It's a good thing to make it possible. Need to think if worth it to create the tests now or if I should just test this manually first for a PR review and create a card about tests.

alneberg:
I haven't managed to write any automatic tests, but some manual testing showed that the corresponding User row is not deleted when for example deleting a UnitUser. Not sure how to solve it though.

alneberg (approved):
Works great now! Sorry for the delay, the factories are almost done. Will be useful for stress testing at a later stage at least.

inaod568:
> Works great now! Sorry for the delay, the factories are almost done. Will be useful for stress testing at a later stage at least.

Great! No problem! Yes it will 👍🏻

PR #727, opened 2021-12-07
Replace boto with signedurl

inaod568:
Applies to `dds get`: Signed URLS are now generated by the DDS server side and returned, instead of sharing the Safespring keys with the client side. Also, the server checks that the items are found both in the database and in the bucket before returning the info.


dds_web/api/api_s3_connector.py
```diff
@@ -126,3 +126,25 @@ def remove_one(self, file, *args, **kwargs):
             removed = True
 
         return removed, error
+
+    def generate_get_url(self, key):
+        """Generate presigned urls for get requests."""
+
+        url = self.resource.meta.client.generate_presigned_url(
```
alneberg, line 133: Would be nice to add a comment here about the question we had regarding http requests made or not. The conclusion was that it probably doesn't make requests I think.

inaod568: Yeah, it doesn't. And sure, I can add a comment about it.

dds_web/api/schemas/project_schemas.py
```diff
@@ -84,3 +86,120 @@ def return_items(self, data, **kwargs):
...
+
+        # Not found
+        not_found = {x: folder_contents.pop(x) for x, y in list(folder_contents.items()) if not y}
+
+        return files, folder_contents, not_found
+
+    @marshmallow.post_dump
+    def return_items(self, data, **kwargs):
+        """Return project contents as serialized."""
+
```
alneberg, line 125: Personally I would move almost everything in this Schema back into the get method of the files endpoint. I think of the schemas as a place to validate parameters and possibly check user permissions. How do you think of them? The decision where to put the logic in the end is of course yours though.

inaod568: That would mean that we would have very similar/identical code for `FileInfo` and `FileInfoAll`, and I was thinking that this schema could also be used in `MatchFiles` later on since we can specify which columns should be returned etc.

inaod568: But Schemas are also good for serializing and deserializing objects

alneberg: Code reusability is a good point. I guess I haven't gotten used to the schemas yet then.

dds_web/api/schemas/project_schemas.py
```diff
@@ -84,3 +86,120 @@ def return_items(self, data, **kwargs):
...
+            for page in pages:
+                found_files.update(
+                    {
+                        x.name: {
+                            **fileschema.dump(x),
+                            "url": s3.generate_get_url(key=x.name_in_bucket) if url else None,
+                        }
+                        for x in files
+                        if x.name_in_bucket in page
+                    }
```
alneberg, line 183: If I understand this correctly, I think we can remove the check if a file is in the bucket or not. If a file that is in the database is not found in the bucket, that's a fairly serious error and we should raise an exception instead of silently filtering them out of the download list. I think you said the http request in the presigned URL would return 403 for non-existing files so I guess it would be easiest to raise that exception in the CLI for the user.

inaod568: Ok! I was thinking of doing something like that originally, I have no objections, will change.

inaod568: Done

dds_web/config.py
```diff
@@ -35,7 +35,6 @@ class Config(object):
     LOCAL_TEMP_CACHE = "/dds_web/local_temp_cache"
     DDS_S3_CONFIG = "/code/dds_web/sensitive/s3_config.json"
     SAFESPRING_URL = os.environ.get("DDS_SAFESPRING_URL", "https://example.endpoint.net")
-    DDS_SAFESPRING_PROJECT = os.environ.get("DDS_SAFESPRING_PROJECT", "dds.example.com")
```
alneberg, line 38: I think there might be a few other places that needs updating for this as well? The README and the db_init perhaps? Should we even remove `safespring_name` column for Units?

alneberg (approved):
Very nice! Works well as well. Might need some cleanup to get rid of the safespring_project variable completely but it works as it is as well.

dds_web/config.py
inaod568: No that should be kept for invoicing purposes 💰

inaod568: But other places aside from the column, yes

PR #728, opened 2021-12-08
Minor updates and fixes to statuses

aanil:
* Sort statuses by descending in history
* Fix deadline bug
* Clarify error message.

### DIFF
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -163,7 +163,7 @@ class ProjectStatus(flask_restful.Resource):
             history = []
             for pstatus in project.project_statuses:
                 history.append(tuple((pstatus.status, pstatus.date_created)))
-            history.sort(key=lambda x: x[1])
+            history.sort(key=lambda x: x[1], reverse=True)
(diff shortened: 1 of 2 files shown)


inaod568 (changes requested):
I'm thinking that we should change `add_deadline` from a dict. And have like `new_deadline = None` and then in the `ProjectStatuses` we do 
```
add_status = models.ProjectStatuses(
    **{"project_id": project.id, "status": new_status, "date_created": curr_date},
    deadline=new_deadline, 
    ...
)
```

instead?

aanil:
> I'm thinking that we should change `add_deadline` from a dict. And have like `new_deadline = None` and then in the `ProjectStatuses` we do
> 
> ```
> add_status = models.ProjectStatuses(
>     **{"project_id": project.id, "status": new_status, "date_created": curr_date},
>     deadline=new_deadline, 
>     ...
> )
> ```
> 
> instead?

Done!

inaod568 (approved):
Looks great and works with the CLI PR!

PR #732, opened 2021-12-09
Remove expiredfiles

inaod568:
Removed the ExpiredFiles table and the scheduled cronjobs since they need to be rewritten a lot.


alneberg (approved):
Looks good!

PR #733, opened 2021-12-09
Factories to create larger test datasets

alneberg:
Initiate the bigger database by running `export DDS_INIT_DEV_DB=big` before starting with `docker-compose up`.  To change back again use `export DDS_INIT_DEV_DB=small`. The size of the database is currently hardcoded in the `factories.py` file which is of course not ideal, but we will only need it for stress testing on dev anyway. 

With the values set now it took about 10 minutes to populate the database. There is a risk that higher values might start consuming a lot of memory. We might need to add intermediate commit statements in that case.


inaod568 (approved):
Nice! Tried it out and it works really well.

PR #741, opened 2021-12-14
Deleting data and files while archiving and deleting projects

aanil:
* Does not touch the ProjectStatuses table - Should we keep the status history or only keep the latest status(which would be deleted or archived).
* Tests updated.

### DIFF
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -139,7 +139,7 @@ class ProjectStatus(flask_restful.Resource):
         try:
             project.project_statuses.append(add_status)
             if not project.is_active:
-                # Deletes files (also commits session in the function - should it be decoupled?)
+                # Deletes files (also commits session in the function - possibly refactor later)
                 removed = RemoveContents().delete_project_contents(project)
                 delete_message = f"\nAll files in {public_id} deleted"
                 if new_status == "Deleted" or is_aborted:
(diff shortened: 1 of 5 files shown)


aanil:
Added the special case to be able to abort a project if it has been made available.

dds_web/api/project.py
```diff
@@ -115,38 +116,56 @@ def post(self):
...
         )
+        delete_message = ""
         try:
             project.project_statuses.append(add_status)
-            db.session.commit()
-            if new_status in ["Deleted", "Archived"]:
-                # TODO Call function to delete files
-                print("TODO", flush=True)
+            if not project.is_active:
+                # Deletes files (also commits session in the function - should it be decoupled?)
```
inaod568, line 142: We can keep it like it is for now and possibly refactor later 👍🏻

dds_web/api/project.py
```diff
@@ -284,43 +323,35 @@ def delete(self):
...
-            if not removed:
-                raise DeletionError(
-                    message="No project contents deleted.",
-                    username=current_user.username,
-                    project=project.public_id,
-                )
+        if not project.files:
+            raise EmptyProjectException("The are no project contents to delete.")
+        try:
+            delete_project_contents(project)
```
inaod568, line 329: `self.delete_project_contents(project)`

inaod568 (changes requested):
Looks good! Just a couple of small changes

dds_web/api/project.py
aanil: 👍

dds_web/api/project.py
aanil: Fixed!

inaod568 (approved):

inaod568:
> * Does not touch the ProjectStatuses table - Should we keep the status history or only keep the latest status(which would be deleted or archived).
> 
>     * Tests updated.

Can you update this description a little bit? Thanks!

PR #744, opened 2021-12-15
Possible to change password using the current password

alneberg:
I implemented a way to change password in the web, but I then realised we maybe should have it in the api instead?

### DIFF
--- a/dds_web/forms.py
+++ b/dds_web/forms.py
@@ -7,6 +7,7 @@ import re
 
 # Installed
 import flask_wtf
+import flask_login
 import wtforms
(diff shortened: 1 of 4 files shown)


inaod568:
> I implemented a way to change password in the web, but I then realised we maybe should have it in the api instead?

Repetition of what was said in Slack channel: I think it's fine to have this in the web since we have the "Forgot Password" part there too. But yes, aside from this lets avoid adding stuff to the web.

inaod568 (approved):
Nice! Will merge as soon as the tests are done

Issue #712, opened 2021-11-19
Username validation in form


alneberg:
Include username format in the form validation instead of failing in the NewUserSchema as it's much harder for the user to adjust it then. 

Also do not prefill the username field on the form with something which is not valid. Currently a '.' in the email is invalid in the suggested username.

PR #747, opened 2021-12-16
Invite tweaks

alneberg:
Typofix and avoid prefilling the sign up form with an invalid username (leave blank instead).


alneberg:
Closes #712

inaod568:
It's like you read my mind (was going to suggest to fix this next).

inaod568 (approved):

alneberg:
> It's like you read my mind (was going to suggest to fix this next).

No, but I read the issues you submit. 👍

PR #748, opened 2021-12-16
Add new endpoint to remove an associated user from a project

aanil:
* Add endpoint `/user/rm_from_project` to remove a user from a project with arguments email and project.
* Add some tests.


tests/__init__.py
```diff
@@ -113,6 +113,9 @@ class DDSEndpoint:
     TOKEN = BASE_ENDPOINT + "/user/token"
     ENCRYPTED_TOKEN = BASE_ENDPOINT + "/user/encrypted_token"
 
+    # Remove user from project
+    REMOVE_USER_FROM_PROJ = BASE_ENDPOINT + "/user/rm_from_project"
```
inaod568, line 117: https://github.com/ScilifelabDataCentre/dds_cli/pull/210#discussion_r773077434

dds_web/api/user.py
```diff
@@ -418,6 +418,55 @@ def post(self):
...
+            )
+        else:
+            message = f"Error: {user_email} does not exist in the db."
+            status = ddserr.error_codes["NoSuchUserError"]["status"].value
+
+        return flask.make_response(
+            flask.jsonify({"message": message}),
+            status,
+        )
+
```
inaod568, line 468: Use this instead: `return {"message": message}, status`
You don't need flask make response for `flask_restful` apparently as of Flask 1.1 flask automatically jsonifies any dictionary that's passed as the first return value. And the status can just be placed after. But 200 is the default so you don't have to specify that in the different elses.

dds_web/api/user.py
```diff
@@ -418,6 +418,55 @@ def post(self):
...
+            else:
+                message = "User already not associated with this project"
+                status = 200
+
+            flask.current_app.logger.debug(
+                f"User {existing_user.username} no longer associated with project {project.public_id}."
+            )
+        else:
+            message = f"Error: {user_email} does not exist in the db."
+            status = ddserr.error_codes["NoSuchUserError"]["status"].value
```
inaod568, line 462: We shouldnt return that a user doesn't exist, instead return that "the user is not associated to this project now" (not exactly that text but just saying "sure, the user doesn't have access" )

tests/test_user_remove_association.py
```diff
@@ -0,0 +1,87 @@
+# Installed
+import json
+import http
+import copy
+
+# Own
+import tests
+from tests.test_project_creation import proj_data_with_existing_users
+
+
```
inaod568, line 10: Add test for if the project doesn't exist maybe?

inaod568:
Maybe you've already seen this but you probably need to change the endpoint in the test directory @aanil

dds_web/api/user.py
aanil: 👍

dds_web/api/user.py
aanil: Changed it now.

tests/test_user_remove_association.py
aanil: Done.

inaod568 (approved):

PR #751, opened 2021-12-20
Delete user tests with web login

alneberg:
- A way to fake a web login in unit tests
 - Basic error handling for the web part
 - Tests for delete user where web login is needed.

### DIFF
--- a/dds_web/web/user.py
+++ b/dds_web/web/user.py
@@ -38,12 +38,6 @@ auth_blueprint = flask.Blueprint("auth_blueprint", __name__)
 ####################################################################################################
 
 
-class UserDeletionErrorWeb(werkzeug.exceptions.BadRequest):
-    def __init__(self, message):
-        super().__init__()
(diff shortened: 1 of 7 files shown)


dds_web/web/user.py
```diff
@@ -28,13 +29,35 @@
...
+####################################################################################################
+# ERROR HANDLING ################################################################## ERROR HANDLING #
+####################################################################################################
+
+
+class UserDeletionErrorWeb(werkzeug.exceptions.BadRequest):
+    def __init__(self, message):
+        super().__init__()
+        self.message = message
+
```
inaod568, line 45: Bugs me a bit that we can't use the same exceptions for the api and web. Feels like there should be a way of doing this instead of having duplicates?

alneberg: Hopefully fixed now!

inaod568 (approved):
Nice, looks good!

PR #752, opened 2021-12-20
Buggy constraints?

alneberg:
The tests `test_delete_unit_row__with_users` and `test_delete_unit_row__with_users2` are identical except for the `print(unit.users)` statement. Locally it seems that the `test_delete_unit_row__with_users` test always fails while `test_delete_unit_row__with_users2` always passes...

Let's see what github actions find.


alneberg:
It worked as it should when the tests ran on github actions so false alarm. Seems to be some issue with my local containers.

PR #753, opened 2021-12-20
Issue703: Change from post to delete requests in the User deletion endpoints.

MatthiasZepper:
Nanoscale PR to change the _post_ requests to _delete_ requests for the user deletion endpoints.

### DIFF
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -286,7 +286,7 @@ class DeleteUserSelf(flask_restful.Resource):
     """
 
     @auth.login_required
-    def post(self):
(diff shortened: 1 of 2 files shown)


inaod568 (changes requested):
Yeah just change to delete in the tests as well and it'll all be fixed

MatthiasZepper:
> Yeah just change to delete in the tests as well and it'll all be fixed

Changing the tests? That will quadruple the amount of work I have to do for this PR! 🤣😝

inaod568 (approved):
Approved! Thanks! That was fast!

PR #758, opened 2021-12-22
Preliminary look at automatic capture of update time and updater on Project

aanil:
* Add a listener to `Project` that populates the fields `Project.date_updated` and `Project.last_updated_by` right before an update.
* Works when changing statuses, uploading/deleting files, adding users.
* Removing users from a project does not update the Project row with this addition(that functionality is not yet merged yet either).


inaod568 (approved):
Works and looks nice!

PR #759, opened 2021-12-22
Speedup tests

alneberg:
Somewhat messy PR that probably should be kept until after Xmas. Opening it now so I won't forget about it.


alneberg:
Closing this as I'm not sure why some tests fail and don't think it's important enough to spend time on.

PR #761, opened 2022-01-03
Add public REST Endpoint for listing roles and permissions

aanil:
List name, role, username and emails for a logged-in user.


inaod568 (approved):
Looks good!

PR #763, opened 2022-01-10
Validation for username in schema same as in form

alneberg:
Mostly to keep it consistent. It's quite unlikely that one registers via a POST request without the form, but still.


inaod568 (approved):
Looks good!

PR #768, opened 2022-01-14
Retry if sending invite fails

aanil:
No easy way to find out if sending succeeded, so made a function to retry sending twice if an exception occurs during sending an invite.

Can close this if we feel this is overkill.

### DIFF
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -238,7 +238,6 @@ class AddUser(flask_restful.Resource):
         try:
             mail.send(msg)
         except Exception as err:
-            print("here")
             # Wait a little bit
             time.sleep(10)
             # Retry twice
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -218,10 +218,33 @@ class AddUser(flask_restful.Resource):
             unit_email=unit_email,
         )
 
-        mail.send(msg)
+        AddUser.send_email_with_retry(msg)
+
+        # Append invite to unit if applicable
+        if new_invite.role in ["Unit Admin", "Unit Personnel"]:
+            auth.current_user().unit.invites.append(new_invite)
+        else:
+            db.session.add(new_invite)
+
+        db.session.commit()
 
         # TODO: Format response with marshal with?
         return {"email": new_invite.email, "message": "Invite successful!", "status": 200}
+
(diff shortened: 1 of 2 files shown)


dds_web/api/user.py
```diff
@@ -218,11 +218,34 @@ def invite_user(args):
             unit_email=unit_email,
         )
 
-        mail.send(msg)
+        AddUser.send_email_with_retry(msg)
```
inaod568, line 221: Not `self`?

dds_web/api/user.py
```diff
@@ -218,11 +218,34 @@ def invite_user(args):
...
         return {"email": new_invite.email, "message": "Invite successful!", "status": 200}
 
+    @staticmethod
+    def send_email_with_retry(msg, times_retried=0):
+        """Send email with retry on exception"""
+
+        try:
+            mail.send(msg)
+        except Exception as err:
+            print("here")
```
inaod568, line 241: Remove

inaod568 (changes requested):
I like it! Don't see any reason not to keep this. 

Just remove the `print` and I'll approve. The other comment is not a dealbreaker but (unless I'm missing something and it doesn't work with self for some reason) I think `self` is better.

dds_web/api/user.py
aanil: Oops! Removed.

dds_web/api/user.py
aanil: Its a static function, so can't be called with `self`.

inaod568: Riiight nvm

dds_web/api/user.py
```diff
@@ -218,11 +218,33 @@ def invite_user(args):
...
         # TODO: Format response with marshal with?
         return {"email": new_invite.email, "message": "Invite successful!", "status": 200}
 
+    @staticmethod
+    def send_email_with_retry(msg, times_retried=0):
+        """Send email with retry on exception"""
+
+        try:
+            mail.send(msg)
+        except Exception as err:
```
inaod568, line 240: Change this to the `SMTPException` and we're all good 👍🏻

inaod568 (approved):
Looks good!

Issue #686, opened 2021-11-03
Go through and improve action and access logging 


inaod568:

aanil:
The issue https://github.com/ScilifelabDataCentre/dds_web/issues/629 seems to be really similar to this one from the description.

inaod568:
> The issue #629 seems to be really similar to this one from the description.

@aanil Yes! Have closed the other one and referred to this one.

PR #771, opened 2022-01-17
Issue 686 - Introduce Structured Logging

MatthiasZepper:
This PR introduces the `structlog` package and aims to extent and improve action logging (#686), while the general logger is untouched. To achieve this a new decorator `logging_bind_request` was written that can be attached to the API endpoints. 

Additionally, a new intermediate superclass for the `HTTP.exceptions` was written that is suposed to standardize action logging for exceptions. Instead of inheriting directly from `HTTP.exceptions`, now most exceptions inherit from this superclass which itself in a child of `HTTP.exceptions`.

### DIFF
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -6,11 +6,10 @@
 
 # Standard library
 import logging
-import structlog
 
 # Installed
-import flask
 import click
+import flask
 from flask_sqlalchemy import SQLAlchemy
 from flask_marshmallow import Marshmallow
 from logging.config import dictConfig
@@ -23,8 +22,10 @@ import flask_login
 # import flask_qrcode
 from flask_limiter import Limiter
 from flask_limiter.util import get_remote_address
-import werkzeug
 import sqlalchemy
+import structlog
+import werkzeug
+
 
 ####################################################################################################
 # GLOBAL VARIABLES ############################################################## GLOBAL VARIABLES #
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -68,13 +68,8 @@ def setup_logging(app):
             "version": 1,
             "disable_existing_loggers": False,
             "formatters": {
+                "default": {"format": "%(message)s"},
                 "general": {"format": "[%(asctime)s] %(module)s [%(levelname)s] %(message)s"},
-                "actions": {
-                    "format": (
-                        "[%(asctime)s] [%(levelname)s] <%(module)s> :: [%(result)s | "
-                        "Attempted : %(action)s | Project : %(project)s | User : %(current_user)s]"
-                    )
-                },
             },
             "handlers": {
                 "general": {
@@ -89,7 +84,7 @@ def setup_logging(app):
                     "class": "dds_web.dds_rotating_file_handler.DDSRotatingFileHandler",
                     "filename": "actions",
                     "basedir": app.config.get("LOGS_DIR"),
-                    "formatter": "actions",
+                    "formatter": "default",
                 },
                 "console": {
                     "level": logging.DEBUG,
@@ -103,14 +98,21 @@ def setup_logging(app):
                     "level": logging.DEBUG,
                     "propagate": False,
                 },
-                "actions": {"handlers": ["actions"], "level": logging.INFO, "propagate": False},
+                "actions": {
+                    "handlers": ["actions", "console"],
+                    "level": logging.INFO,
+                    "propagate": False,
+                },
             },
         }
     )
 
     structlog.configure(
         processors=[
-            # If log level is too low, abort pipeline and throw away log entry.
+            # Merge the bindings from Thread-Local Context, a sort of global context storage.
+            structlog.threadlocal.merge_threadlocal,
+            # Mimics the level configuration of the standard logging lib.
+            # e.g. logger.debug() event will be dropped if logging level is set to INFO or higher.
             structlog.stdlib.filter_by_level,
             # Add the name of the logger to event dict.
             structlog.stdlib.add_logger_name,
@@ -118,7 +120,7 @@ def setup_logging(app):
             structlog.stdlib.add_log_level,
             # Perform %-style formatting.
             structlog.stdlib.PositionalArgumentsFormatter(),
-            # Add a timestamp in ISO 8601 format.
+            # Add a timestamp to the event dict in ISO 8601 format.
             structlog.processors.TimeStamper(fmt="iso"),
             # If the "stack_info" key in the event dict is true, remove it and
             # render the current stack trace in the "stack" key.
@@ -133,7 +135,7 @@ def setup_logging(app):
             structlog.processors.JSONRenderer(),
         ],
         # `wrapper_class` is the bound logger that you get back from
-        # get_logger(). This one imitates the API of `logging.Logger`.
+        # get_logger(). This one imitates `logging.Logger`.
         wrapper_class=structlog.stdlib.BoundLogger,
         # `logger_factory` is used to create wrapped loggers that are used for
         # OUTPUT. This one returns a `logging.Logger`. The final value (a JSON
@@ -171,7 +173,7 @@ def create_app(testing=False, database_uri=None):
         limiter.logger.addHandler(handler)
 
     # Set app.logger as the general logger
-    app.logger = structlog.getLogger("general")
+    app.logger = logging.getLogger("general")
     app.logger.info("Logging initiated.")
 
     # Initialize database
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -6,6 +6,7 @@
 
 # Standard library
 import logging
+import structlog
 
 # Installed
 import flask
@@ -107,6 +108,43 @@ def setup_logging(app):
         }
     )
 
+    structlog.configure(
+        processors=[
+            # If log level is too low, abort pipeline and throw away log entry.
+            structlog.stdlib.filter_by_level,
+            # Add the name of the logger to event dict.
+            structlog.stdlib.add_logger_name,
+            # Add log level to event dict.
+            structlog.stdlib.add_log_level,
+            # Perform %-style formatting.
+            structlog.stdlib.PositionalArgumentsFormatter(),
+            # Add a timestamp in ISO 8601 format.
+            structlog.processors.TimeStamper(fmt="iso"),
+            # If the "stack_info" key in the event dict is true, remove it and
+            # render the current stack trace in the "stack" key.
+            structlog.processors.StackInfoRenderer(),
+            # If the "exc_info" key in the event dict is either true or a
+            # sys.exc_info() tuple, remove "exc_info" and render the exception
+            # with traceback into the "exception" key.
+            structlog.processors.format_exc_info,
+            # If some value is in bytes, decode it to a unicode str.
+            structlog.processors.UnicodeDecoder(),
+            # Render the final event dict as JSON.
+            structlog.processors.JSONRenderer(),
+        ],
+        # `wrapper_class` is the bound logger that you get back from
+        # get_logger(). This one imitates the API of `logging.Logger`.
+        wrapper_class=structlog.stdlib.BoundLogger,
+        # `logger_factory` is used to create wrapped loggers that are used for
+        # OUTPUT. This one returns a `logging.Logger`. The final value (a JSON
+        # string) from the final processor (`JSONRenderer`) will be passed to
+        # the method of the same name as that you've called on the bound logger.
+        logger_factory=structlog.stdlib.LoggerFactory(),
+        # Effectively freeze configuration after creating the first bound
+        # logger.
+        cache_logger_on_first_use=True,
+    )
+
 
 def create_app(testing=False, database_uri=None):
     """Construct the core application."""
@@ -133,7 +171,7 @@ def create_app(testing=False, database_uri=None):
         limiter.logger.addHandler(handler)
 
     # Set app.logger as the general logger
-    app.logger = logging.getLogger("general")
+    app.logger = structlog.getLogger("general")
     app.logger.info("Logging initiated.")
 
     # Initialize database
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -100,7 +100,7 @@ def setup_logging(app):
                     "propagate": False,
                 },
                 "actions": {
-                    "handlers": ["actions", "console"],
+                    "handlers": ["actions"],
                     "level": logging.INFO,
                     "propagate": False,
                 },
--- a/dds_web/api/errors.py
+++ b/dds_web/api/errors.py
@@ -10,14 +10,15 @@ import logging
 # Installed
 from werkzeug import exceptions
 import flask
+import dds_web
 import flask_login
 import http
 import json
 import structlog
 
 # Own modules
-from dds_web import actions
-from dds_web import auth
+from dds_web import actions, auth
+from dds_web.utils import get_username_or_request_ip, remove_sensitive_args
 
 ####################################################################################################
 # LOGGING ################################################################################ LOGGING #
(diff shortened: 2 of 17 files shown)


dds_web/__init__.py
```diff
@@ -102,11 +98,55 @@ def setup_logging(app):
                     "level": logging.DEBUG,
                     "propagate": False,
                 },
-                "actions": {"handlers": ["actions"], "level": logging.INFO, "propagate": False},
+                "actions": {
+                    "handlers": ["actions", "console"],
```
MatthiasZepper, line 102: Currently, action logging is also output to console for debugging purposes. Will remove that before merging!

dds_web/api/errors.py
```diff
@@ -21,10 +24,44 @@
...
+        if auth.current_user():
+            current_user = auth.current_user().username
+        elif flask_login.current_user.is_authenticated:
+            current_user = flask_login.current_user.username
+        elif flask.request.remote_addr:
+            current_user = flask.request.remote_addr  # log IP instead of username
+        elif flask.request.access_route:
+            current_user = flask.request.access_route[0]  # log IP instead of username
+        else:
+            current_user = "anonymous"
```
inaod568, line 46: Since this is the same as in [here](https://github.com/ScilifelabDataCentre/dds_web/pull/771/files#diff-bacb9c85e5e141775ed3b7a626bf14730d2ec362398022c166bead23297eecc8R92-R101) maybe we should have a utils function for this?

dds_web/api/errors.py
```diff
@@ -68,23 +105,16 @@ def __init__(
         username=None,
         message="The user does not have the necessary permissions.",
     ):
-        super().__init__(message)
-
-        if not username:
-            username = auth.current_user()
+        if username:
+            structlog.threadlocal.bind_threadlocal(user=username)
```
inaod568, line 109: Do we need to release after binding too or is this fine? Or unbind or what it's called

dds_web/api/files.py
```diff
@@ -69,6 +70,7 @@ class NewFile(flask_restful.Resource):
     """Inserts a file into the database"""
 
     @auth.login_required(role=["Super Admin", "Unit Admin", "Unit Personnel"])
+    @logging_bind_request
```
inaod568, line 73: We definitely need logging here too but I'm wondering about the `flask.request.json` -- in this (and a few more) case the json info will include all info that will be put into the database. So we should maybe have a check and for some endpoints not log that.

dds_web/api/user.py
```diff
@@ -449,6 +461,12 @@ def delete(self):
         msg = f"The user account {user.username} ({user_email_str}, {user.role})  has been terminated successfully been by {current_user.name} ({current_user.role})."
         flask.current_app.logger.info(msg)
 
+        with structlog.threadlocal.bound_threadlocal(
+            who={"user": user.username, "role": user.role},
+            by_whom={"user": current_user.username, "role": current_user.role},
+        ):
+            action_logger.info(self.__class__)
+
```
inaod568, line 469: Do we need this now that we have the decorator?

inaod568 (commented):
I like it! 
I have a few comments/questions to clear out, some of them because I'm not entirely sure how it works. But in general looks really good 👍🏻

dds_web/api/errors.py
MatthiasZepper: Sure. It also grew over time...

dds_web/api/errors.py
MatthiasZepper: In theory yes and there is also the  s`tructlog.threadlocal.clear_threadlocal()` command to clear the threadlocal cache. However, I have mostly struggled with a lack of persistence rather than the other way round during development. I will do a bit more testing.

dds_web/api/user.py
MatthiasZepper: Not necessarily. It depends on how verbose you as Super Admin would like the logs to be.

The basic idea was that the decorator would only cover the bare minimum of information that is required for each log entry and that additional logs can be written for important actions inside the respective functions.

In this case, I felt that user deletion is such a relevant action, that one might want to access additional information quickly without studding it from the request and username fields. Having the separate, function-specific JSON fields in the logs would simplify aggregating and accessing this information later with `jq`.

inaod568: Okey! Good point, lets keep this the way it is.

dds_web/api/files.py
inaod568: I will have a look!

inaod568: `dds_cli/dev` works with this branch, but you're probably getting the bucket error because you're using the development entries - the projects that are automatically created when starting up the service. So do (in dds_cli):
```
git checkout dev
git pull
dds project create -t test -d test -pi test
dds data put -p <new-project-id-displayed-from-create> -s <file-you-want-to-upload>
```

When I do this the action logger outputs the following
```
dds_backend | [2022-01-19 07:40:06,260] _base [INFO] {"resource": "/api/v1/file/new", "request_args": {"name": "requirements.txt", "name_in_bucket": "0fe9fe9c-1e57-5095-8f94-9b7102c41ae2", "subpath": ".", "size": 267, "size_processed": 250, "compressed": true, "salt": "B639892A751B4ABC94E33FCAAB6D9972", "public_key": "E091222CBE0E063E00F811C015ABE9F383C61A578DE12A5DD5351B596CEF4B55", "checksum": "e449dba65a98df64fc7eedb4ed1d49dc2cc58d89f6098ae8783cb284d0d16362"}, "user": "unituser_1", "response": "200 OK", "event": "api_blueprint.new_file.post", "logger": "actions", "level": "info", "timestamp": "2022-01-19T07:40:06.260174Z"}
```

inaod568: @MatthiasZepper 
We could have something like this:
```python
# dds_web/api/dds_decorators.py

def get_user_to_log():
    if auth.current_user():
        current_user = auth.current_user().username
    elif flask_login.current_user.is_authenticated:
        current_user = flask_login.current_user.username
    elif flask.request.remote_addr:
        current_user = flask.request.remote_addr  # log IP instead of username
    elif flask.request.access_route:
        current_user = flask.request.access_route[0]  # log IP instead of username
    else:
        current_user = "anonymous"
    return current_user


def do_logging(log_json=False):
    def logging_decorator_name(func):
        """Binds some request parameters to the thread-local context of structlog"""

        @functools.wraps(func)
        def wrapper_logging_bind_request(*args, **kwargs):

            flask.current_app.logger.debug(f"Test: {log_json}")
            current_user = get_user_to_log()

            with structlog.threadlocal.bound_threadlocal(
                resource=flask.request.path or "not applicable",
                request_args=flask.request.json if (flask.request.data and log_json) else None,
                user=current_user,
            ):

                value = func(*args, **kwargs)

                if hasattr(value, "status"):
                    structlog.threadlocal.bind_threadlocal(response=value.status)

                action_logger.info(f"{flask.request.endpoint}.{func.__name__}")

                return value

        return wrapper_logging_bind_request

    return logging_decorator_name
```
And then at the endpoints we do want to log json we do (the `log_json` could also of course be `True` by default if we decide that is better)
```
    @auth.login_required(role=["Super Admin", "Unit Admin", "Unit Personnel"])
    @do_logging()
    def post(self):
```

dds_web/api/errors.py
MatthiasZepper: I have now included a `structlog.threadlocal.clear_threadlocal()` after the log is called. 

In particular for the exceptions it may be desirable to be able to gradually assign values to the local context, which are added to the log event in case an exception occurs. Therefore, I think it is preferable to not clear the context before the event is logged. 

But once the log event is written, I agree, it is good practise to immediately clear out the context to not have forgotten assignments lingering around and obfuscating subsequent log events.

dds_web/api/files.py
MatthiasZepper: > ```shell
> dds project create -t test -d test -pi test
> dds data put -p <new-project-id-displayed-from-create> -s <file-you-want-to-upload>
> ```

YES, thank you! I always tried uploading into one of the existing projects! Now I finally know how that works!

> ```shell
>  @do_logging(log_json=True)
> ```

Cool, decorators with arguments. That is impressive!

However, I would suggest a different approach: Instead of logging everything or nothing, I have now created a utility function that will censor sensitive (or irrelevant) arguments from the requests prior to logging them. I have put the droplist/blocklist into the config, which to me seemed a better place than to hardcode it inside the function. 

Admittedly, it lacks the brevity of the other config entries (thus clutters the config) and will probably hardly change, so it might not be necessary to have it in the config?

dds_web/config.py
```diff
@@ -42,6 +42,17 @@ class Config(object):
...
+    # Request parameters dropped from the logfile
+    SENSITIVE_REQUEST_ARGS = [
+        "checksum",
+        "compressed",
+        "public_key",
+        "size",
+        "size_processed",
+        "salt",
+        "token",
+    ]
```
inaod568, line 54: This assumes though that the args are always written in this exact way. So if (for some reason) the variable is actually in caps, then this won't catch that that information is actually sensitive and still log it. So in that case we would need to check caps as well, and possibly also misspelled args. Shouldn't happen of course, just thinking different scenarios and worried that something will accidentally be logged. Also, someone could potentially post a huge amount of json to the server, which does not include these args, and that will let everything be logged. This applies to all endpoints though so we should probably think about if we really need json to be logged and, if so, I'd suggest specifying what actually _should_ be logged instead of what shouldn't.

dds_web/api/files.py
inaod568: My reply: https://github.com/ScilifelabDataCentre/dds_web/pull/771#discussion_r788402501

dds_web/api/errors.py
inaod568: Ah I saw you added it to within the [context manager](https://github.com/ScilifelabDataCentre/dds_web/pull/771/files#diff-bacb9c85e5e141775ed3b7a626bf14730d2ec362398022c166bead23297eecc8R91) -- when the `bound_threadlocal` is used as a context manager that itself should actually clean it up after. So there shouldn't be a need for the clearing within this. I thought it would be in the exception in this case, but if not then we shouldn't need this at all 👍🏻

inaod568: @MatthiasZepper https://www.structlog.org/en/stable/api.html#structlog.threadlocal.bound_threadlocal

inaod568: So we an remove the `structlog.threadlocal.clear_threadlocal()` correct?

MatthiasZepper: Depending on the behavior that we want. I initially felt that it might be helpful to explicitly bind variables to the context for a longer time such that they are retained for multiple log entries, but later agreed that the risk of having some forgotten binding around is to high, so I included the clear command.

The source of `bound_threadlocal` is:
```
def bound_threadlocal(**kw: Any) -> Generator[None, None, None]:
    context = get_threadlocal()
    saved = {k: context[k] for k in context.keys() & kw.keys()}

    bind_threadlocal(**kw)
    try:
        yield
    finally:
        unbind_threadlocal(*kw.keys())
        bind_threadlocal(**saved)
```

So it seems to save and restore only values that are explicitly assigned within the `bound_threadlocal`, but not keys that will be assigned within the context later.

```
bind_threadlocal(a=1, b=2, c=3)
get_threadlocal()   # {'a': 1, 'b': 2, 'c': 3}

with  bound_threadlocal(a=4, b=5, d=6):
        bind_threadlocal(c=5, e=7)
        # a and b from the context manager take precedence over the initial bindings
        get_threadlocal()   #{'a': 4, 'b': 5, 'c': 5, 'd': 6, 'e': 7}

get_threadlocal()  # {'a': 1, 'b': 2, 'c': 5, 'e': 7}
# c was not restored, d is not retained

clear_threadlocal()
get_threadlocal() # {}
```

dds_web/config.py
MatthiasZepper: The objections have been rendered irrelevant by the current decision to not log JSON and arbitrary arguments, but I would nevertheless like to respond briefly: 

_So in that case we would need to check caps as well, and possibly also misspelled args_
I fail to see how that is a problem. If we by mistake have an error in our CLI/API code, we should fix it to be consistent across our endpoints (and also would have to, because Marshmallow would also complain).

_and worried that something will accidentally be logged._ 
In log files that are only accessible to the Super Admins who have full database access. Admittedly, having useless log files that don't help to investigate potential issues are a larger concern to me than having a Super Admin knowing to much about their system.

_Also, someone could potentially post a huge amount of json to the server, which does not include these args, and that will let everything be logged_
It would have been possible to use `len()` to drop to large chunks of information. URLs however also have no limit and can be between >8k and sometimes up to <64k characters long (don't know about Flasks limit), so also logging the project arg could be misused to flood the log with junk information.

dds_web/api/errors.py
inaod568: Ok! Thanks for the demonstration! Lets keep the clear_threadlocal line.

dds_web/config.py
inaod568: > So in that case we would need to check caps as well, and possibly also misspelled args
I fail to see how that is a problem. If we by mistake have an error in our CLI/API code, we should fix it to be consistent across our endpoints (and also would have to, because Marshmallow would also complain).

True. Just thinking about potential issues.

> and worried that something will accidentally be logged.
In log files that are only accessible to the Super Admins who have full database access. Admittedly, having useless log files that don't help to investigate potential issues are a larger concern to me than having a Super Admin knowing to much about their system.

Yes, of course we need useful log files. But again, the action loggers purpose is to just see who or what has attempted what endpoint in what project. So having the possibility of logging json information just adds risk and defeats the purpose of this specific logger. The general logger, however, will log all information regarding potential errors where we can investigate potential issues when there are any. (@alneberg I believe I misspoke yesterday or Wednesday, so just to clarify: The logging occurs no matter what, even if the user is not authenticated. This is because we need to log when a request is denied and be able to figure out why, but also here the json is denied because the user has not been granted access to the system.)

> Also, someone could potentially post a huge amount of json to the server, which does not include these args, and that will let everything be logged
It would have been possible to use len() to drop to large chunks of information. URLs however also have no limit and can be between >8k and sometimes up to <64k characters long (don't know about Flasks limit), so also logging the project arg could be misused to flood the log with junk information.

Sure, that could happen. There's a lower risk of someone passing in a 64k character long project id though then someone passing in a lot of data in the request. But yes, we can add a limit to the length of the project ID being logged. 👍🏻  Lets think about it first though, don't do this now.

Anyway, remove the `SENSITIVE_REQUEST_ARGS`. 👍🏻

dds_web/web/user.py
```diff
@@ -354,9 +354,10 @@ def confirm_self_deletion(token):
     """Confirm user deletion."""
     s = itsdangerous.URLSafeTimedSerializer(flask.current_app.config.get("SECRET_KEY"))
 
+    # Get email from token
+    email = s.loads(token, salt="email-delete", max_age=604800)
+
```
inaod568, line 359: This should be in the try-except block, this is what generates the `itsdangerous.exc.SignatureExpired` exception etc that are caught in the except statements below.

dds_web/config.py
inaod568: The `SENSITIVE_REQUEST_ARGS` seem to still be there so just remove that

inaod568 (changes requested):

tests/test_user_delete.py
```diff
@@ -90,23 +90,25 @@ def test_del_route_invalid_token(client):
     """Use invalid signature"""
     client = tests.UserAuth(tests.USER_CREDENTIALS["delete_me_researcher"]).fake_web_login(client)
 
-    response = client.get(
-        tests.DDSEndpoint.USER_CONFIRM_DELETE + "invalidtoken", content_type="application/json"
-    )
+    with pytest.raises(itsdangerous.exc.BadSignature, match="No b'.' found in value"):
```
inaod568, line 93: Remove

tests/test_user_delete.py
```diff
@@ -90,23 +90,25 @@ def test_del_route_invalid_token(client):
...
 
 def test_del_route_expired_token(client):
     """Use expired token"""
     token = "InJlc2VhcmNodXNlcjFAbWFpbHRyYXAuaW8i.YbIcrg.BmxUW6fKsnC3ujO5z1E_5CYiit4"
     client = tests.UserAuth(tests.USER_CREDENTIALS["delete_me_researcher"]).fake_web_login(client)
 
-    response = client.get(
-        tests.DDSEndpoint.USER_CONFIRM_DELETE + token, content_type="application/json"
-    )
+    with pytest.raises(itsdangerous.exc.BadTimeSignature):
```
inaod568, line 106: Remove

inaod568 (changes requested):
Do this https://github.com/ScilifelabDataCentre/dds_web/pull/771/files#r789476226
And these new test changes 

This works for me

dds_web/web/user.py
MatthiasZepper: Fixed.

inaod568 (approved):
Great, thanks!! Approving!

Issue #719, opened 2021-11-23
Allow users to login with 2fa 


alneberg:
Currently one can register and setup 2fa but actual authentication with 2fa is still to be implemented.

inaod568:
HOTP via email with `cryptography` package

PR #775, opened 2022-01-20
Hotp mail

alneberg:
- New endpoint for api to request one-time authentication code (refferred to as HOTP) to be sent via email - Using basic authentication
 - Require valid HOTP to be supplied to the encrypted_token endpoint, otherwise authentication error
 - Remove signed token endpoint
 - New endpoint for web to request HOTP to be sent via email - requires valid username and password
 - Reuquire valid HOTP to be entered in the web login form
 - Moved `dds_web/api/errors.py` to `dds_web/errors.py` to avoid circular imports when using the errors in side `dds_web/database/models.py`

A new HOTP is issued by:
1. Incrementing the hotp counter 
2. Saving the current time

HOTP validation:
1. Check no more than 1 hour has passed since the HOTP was issued
2. verify HOTP code with current counter
3. If verified, increment counter

This should solve #719 and #773


inaod568 (commented):
Could we change slightly to something like this? 
1. Show login form 
2. User fills in form 
3. Submit triggers an email and displays a form which has the username and password as hidden fields, only showing the message (this has been sent to your email) and the field where the code should filled in
4. Submit triggers complete authentication and login. This still means that the authentication occurrs twice but the user doesn't have to fill it in twice. 

Another alternative is to have an extra column called "2fa_authenticated" and that the user gets logged in once the username and password is verified but the `2fa_authenticated` has to be set to True for the user to be able to access any other endpoints.

dds_web/utils.py
```diff
@@ -319,6 +320,28 @@ def page_query(q):
             break
 
 
+def create_one_time_password_email(user, hotp_value):
+    msg = flask_mail.Message(
+        "DDS One-Time Authentication Code",
+        sender=flask.current_app.config.get("MAIL_SENDER", "dds@noreply.se"),
```
talavis, line 326: I can't find `MAIL_SENDER` in `congif.py`. Either add it or set `MAIL_DEFAULT_SENDER` instead, not setting the sender here.

dds_web/api/__init__.py
```diff
@@ -28,8 +28,8 @@
 ####################################################################################################
 
 # Login/access ###################################################################### Login/access #
-api.add_resource(user.Token, "/user/token", endpoint="token")
 api.add_resource(user.EncryptedToken, "/user/encrypted_token", endpoint="encrypted_token")
+api.add_resource(user.RequestMail2fa, "/user/request_mail2fa", endpoint="request_mail2fa")
```
worukan, line 32: I would have it like 
`api.add_resource(user.SecondFactor, "/user/second_factor", endpoint="second_factor")`
and this endpoint would be used by a user to enter mfa value, not request mfa email.

dds_web/api/user.py
```diff
@@ -533,11 +519,23 @@ class Token(flask_restful.Resource):
...
     def get(self):
-        return flask.jsonify({"token": jwt_token(username=auth.current_user().username)})
+        args = flask.request.json
+        if args is None:
+            args = {}
 
+        data = token_schemas.TokenSchema().load(args)
+
+        return flask.jsonify(
+            {
```
worukan, line 529: I would keep the encrypted token as it is and only add 
`"message": "Please take this token to /user/second_factor to authenticate with MFA!",`

dds_web/api/user.py
```diff
@@ -533,11 +519,23 @@ class Token(flask_restful.Resource):
...
+                "token": encrypted_jwt_token(
+                    username=auth.current_user().username, sensitive_content=None
+                )
+            }
+        )
 
-class EncryptedToken(flask_restful.Resource):
-    """Generates encrypted token for the user."""
+
+class RequestMail2fa(flask_restful.Resource):
```
worukan, line 537: I would replace this class with the following

`class SecondFactor(flask_restful.Resource):
    """Take in and verify an authentication One-Time code entered by an authenticated user with basic credentials"""

    @auth.login_required
    def get(self):
        args = flask.request.json
        if args is None:
            args = {}

        token_schemas.TokenSchema().load(args)

        token_claims = dds_web.security.auth.decrypt_and_verify_token_signature(flask.request.headers["Authorization"].split()[1])

        return flask.jsonify(
            {
                "token": update_token_with_mfa(token_claims)
            }
        )`

requirements.txt
```diff
@@ -35,4 +35,5 @@ Flask-Bootstrap4==4.0.2
 PyQRCode==1.2.1
 Flask-Limiter==1.4
 factory_boy>=3.2.1
+sqlalchemy-utils
```
worukan, line 38: where is this used?

worukan:
The intersection for the token and MFA code to work together would be in auth.py
```
...
@auth.verify_token
def verify_token(token):
...
    if dds_web.utils.current_time() <= datetime.datetime.fromtimestamp(expiration_time):
        username = data.get("sub")
        if username:
            user = models.User.query.get(username)
            if user:
                # return user
                mfa_auth_time_string = data.get("mfa_auth_time")
                if mfa_auth_time_string:
                    mfa_auth_time = datetime.datetime.fromtimestamp(mfa_auth_time_string)
                    if mfa_auth_time >= dds_web.utils.current_time() - MFA_EXPIRES_IN:
                        return user

                if flask.request.path.endswith("/user/second_factor"):
                    return user

                # this should not be triggered more frequently than 15 mins
                #send_hotp_email(user)
                raise AuthenticationError(message="Two-factor authentication is required! "
                                                  "Please check your primary e-mail!")
        return None
    raise AuthenticationError(message="Expired token")
...
```

worukan:
Also, in order to trigger MFA, I would have it as 
```
@basic_auth.verify_password
def verify_password(username, password):
    """Verify that user exists and that password is correct."""
    user = models.User.query.get(username)
    if user and user.verify_password(input_password=password):

        # this should not be triggered more frequently than 15 mins
        send_hotp_email(user)

        return user
    return None
```

requirements.txt
alneberg: It's used in the `conftest.py`. But I have quite likely added it by mistake, I forgot we have a `requirements-test.txt` where this is already added. I'll remove it again.

alneberg: Removed now.

alneberg:
@worukan I see that your solution works as well. The benefit I can spot is that it would very easily allow for a separation between the token expiration and the mfa validity time.  I'ts not really hard from my implementation either though (one would just check the age of the token == time of mfa). 

The downside might be that it's slightly harder to add `TOTP` on top of yours since you directly trigger the hotp-mail upon requesting the token. With my solution a user would be able to send the `TOTP` request parameter directly to the token endpoint without first requesting the hotp token to be sent (since this is manually triggeded).

worukan:
@alneberg I don't think you need to worry about using TOTP. I didn't outline the generic solution right now to save time and focus on the existing solution. I imagine it would depend on checking the user's mfa setup in the db and take the necessary action whether sending an email (for HOTP) or doing nothing (for TOTP). The server should know who uses HOTP or TOTP and act accordingly, and initiate when a user needs to authenticate with an MFA method.

dds_web/api/user.py
worukan: For the previous code, update_token_with_mfa should be defined in dds_web/security/tokens.py as 

```
def update_token_with_mfa(token_claims):
    expires_in = datetime.datetime.fromtimestamp(token_claims.get("exp")) - dds_web.utils.current_time()
    return encrypted_jwt_token(username=token_claims.get("sub"),
                               sensitive_content=None,
                               expires_in=expires_in,
                               additional_claims={"mfa_auth_time": dds_web.utils.current_time().timestamp()})

```

dds_web/api/user.py
worukan: For clarity, I meant that changes to encrypted token can be reverted and only the suggested line in the previous comment can be added.

dds_web/templates/mail/authenticate.html
```diff
@@ -0,0 +1,26 @@
...
+    </tr>
+    <tr>
+        <td style=" font-family: arial, sans-serif; font-size: 14px; vertical-align: top;">
+            <p
+                style="font-family: arial, sans-serif; font-size: 14px; font-weight: bold; margin: 0; Margin-bottom: 5px;">
+                Here is the requested authentication code for SciLifeLab Data Delivery System (DDS)
+            </p>
+            <p
+                style="font-family: arial, sans-serif; font-size: 14px; font-weight: normal; margin: 0; Margin-top: 0px;">
+                Please, copy and paste the code below where you attempting to authenticate: </p>
```
worukan, line 19: Please, copy and paste the code below where you attempt to authenticate

dds_web/templates/mail/authenticate.txt
```diff
@@ -0,0 +1,4 @@
+Here is the requested authentication code for SciLifeLab Data Delivery System (DDS)
+Please, copy and paste the code below where you attempting to authenticate:
```
worukan, line 2: Please, copy and paste the code below where you attempt to authenticate

dds_web/templates/mail/authenticate.html
```diff
@@ -0,0 +1,26 @@
...
+            <p
+                style="font-family: arial, sans-serif; font-size: 14px; font-weight: normal; margin: 0; Margin-top: 0px; word-break: break-all">
+                {{token}}</p>
```
worukan, line 22: Which token is this in here and authenticate.txt as well? Or can it be renamed to one_time_value or similar?

dds_web/templates/user/confirm2fa.html
```diff
@@ -0,0 +1,47 @@
+{% extends "bootstrap/base.html" %}
+{% import "bootstrap/wtf.html" as wtf %}
+
+{% block content %}
+<title>2FA</title>
```
worukan, line 5: Not a significant issue, but would it be better to write something that a user can understand right away? 
Perhaps, DDS (either in acronym or full name) Multi-factor Authentication
I find it useful to have a clear title name for a browser tab which makes it easier to find it among many.

dds_web/templates/user/confirm2fa.html
```diff
@@ -0,0 +1,47 @@
...
+        {% for category, message in messages %}
+        <div class="alert alert-{{category}}" role="alert">
+            {{ message }}
+        </div>
+        {% endfor %}
+    </div>
+    {% endif %}
+    {% endwith %}
+    <p>Please finish the login authenticate yourself to trigger the sending of an authentication one-time code.
+        The one-time codes are valid for a short time (1 hour) after they have been requested.
```
worukan, line 21: This sentence needs a rephrase and we should change 1 hour to 15 minutes both in text and code logic.

dds_web/templates/user/confirm2fa.html
alneberg: Yes, I keep too many tabs to ever be able to read the title...

alneberg: Fixed!

dds_web/templates/user/confirm2fa.html
alneberg: That's a very strange sentence I agree. Ok, let's go with 15 min then.

alneberg: Should be adjusted now.

dds_web/templates/mail/authenticate.html
alneberg: Changed it to `one_time_value`

dds_web/templates/user/confirm2fa.html
```diff
@@ -0,0 +1,47 @@
...
+        {% for category, message in messages %}
+        <div class="alert alert-{{category}}" role="alert">
+            {{ message }}
+        </div>
+        {% endfor %}
+    </div>
+    {% endif %}
+    {% endwith %}
+    <p>Please finish the login by entering the one-time authentication code that was sent to you.
+        The one-time codes are valid for a short time (15 minutes) after they have been requested.
```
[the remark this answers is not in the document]
alneberg: Done

dds_web/templates/mail/authenticate.html
```diff
@@ -0,0 +1,26 @@
...
+        <td>
+            <img src="cid:Logo" alt="Logo" border="0"
+                style="border:0; outline:none; text-decoration:none; display:block; Margin-bottom: 15px; max-width: 400px; max-height: 87px;">
+        </td>
+    </tr>
+    <tr>
+        <td style=" font-family: arial, sans-serif; font-size: 14px; vertical-align: top;">
+            <p
+                style="font-family: arial, sans-serif; font-size: 14px; font-weight: bold; margin: 0; Margin-bottom: 5px;">
+                Here is the requested authentication code for SciLifeLab Data Delivery System (DDS)
```
worukan, line 15: I just noticed that _requested_ is in here as well. Let's change it to **issued**. It is in authenticate.txt as well.

dds_web/api/user.py
alneberg: Not sure I understand the row `expires_in = datetime.datetime.fromtimestamp(token_claims.get("exp")) - dds_web.utils.current_time()`.  What change does that make to the expiration time of the token?

alneberg: Nevermind I think I understand, it's because you add that time when you create the token in the other method.

alneberg:
@worukan do you also have an implementation for the `send_hotp_email` or should I just put the two pieces together that I already have? Also, would you like me to implement a way to prevent the email to be sent more often than once every 15 min?

worukan:
@alneberg 
I think we could place the following in security/auth.py
```
def send_hotp_email(user):
    hotp_value = user.generate_HOTP_token()
    msg = dds_web.utils.create_one_time_password_email(user, hotp_value)
    mail.send(msg)
```

I'm writing a small piece for the 15 mins frequency inside generate_HOTP_token. I will comment that in line in code soon.

dds_web/database/models.py
```diff
@@ -280,8 +285,9 @@ class User(flask_login.UserMixin, db.Model):
     username = db.Column(db.String(50), primary_key=True, autoincrement=False)
     name = db.Column(db.String(255), unique=False, nullable=True)
     _password_hash = db.Column(db.String(98), unique=False, nullable=False)
-    _otp_secret = db.Column(db.String(32))
-    has_2fa = db.Column(db.Boolean)
+    hotp_secret = db.Column(db.LargeBinary(20), unique=False, nullable=False)
+    hotp_counter = db.Column(db.BigInteger, unique=False, nullable=False, default=0)
+    hotp_requested_time = db.Column(db.DateTime, unique=False, nullable=True)
```
worukan, line 290: could we change this to `hotp_issue_time` ?

dds_web/api/schemas/token_schemas.py
```diff
@@ -0,0 +1,52 @@
...
+
+        user = auth.current_user()
+        if "HOTP" in data:
+            value = data.get("HOTP")
+            # Raises authenticationerror if invalid
+            user.verify_HOTP(value.encode())
+            data["mfa_verified"] = True
+            data["mfa_method"] = "HOTP"
+
+        return data
```
worukan, line 52: Does data need to be returned from validate_mfa? What is the use of it? Currently it seems to be not used.

dds_web/database/models.py
```diff
@@ -370,40 +375,42 @@ def verify_reset_token(token):
...
+            raise AuthenticationError("One-time authentication code has expired.")
+
+        try:
+            hotp.verify(token, self.hotp_counter)
+        except twofactor_InvalidToken:
+            raise AuthenticationError("Invalid one-time authentication code.")
+
+        # Token verified, increment counter to prohibit re-use
+        self.hotp_counter += 1
+        db.session.commit()
```
worukan, line 413: I think that lines 412 and 413 are not relevant and can be removed. Generate_HOTP_token already increments the counter before generating. Counter increment on line 412 doesn't happen on invalid one-time auth code and happens on successful verification. So on successful verification the counter is incremented twice.

dds_web/database/models.py
```diff
@@ -370,40 +375,42 @@ def verify_reset_token(token):
...
-        self.has_2fa = True
-
-    def totp_uri(self):
-        """Get uri for user otp_secret if the 2fa has not already been setup."""
-        if self.has_2fa:
-            return None
-        return pyotp.totp.TOTP(self.otp_secret).provisioning_uri(
-            name=self.username, issuer_name="Data Delivery System"
+    def generate_HOTP_token(self):
+        """Generate a one time password, e.g. to be sent by email.
```
worukan, line 379: It would be good to go over all code changes and replace occurrences of _password_ with **authentication code**. Also it is sometimes written with _one time_ and sometimes with **one-time**.

dds_web/api/schemas/token_schemas.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # This can be easily extended to require at least one MFA method
+        if "HOTP" not in data:
+            raise marshmallow.ValidationError("MFA method not supplied")
+
+        user = auth.current_user()
+        if "HOTP" in data:
+            value = data.get("HOTP")
+            # Raises authenticationerror if invalid
+            user.verify_HOTP(value.encode())
```
worukan, line 48: I have a suggestion for line 48. What do you think? Or could you make it even better?
The idea is if the entered otp value has expired or a wrong value is input and it has been more than 15 mins since otp has been issued, we will check and send a new email. As with the error, I didn't want to handle this in the model, so caught it only to raise again after sending email, if at all.

```
try:
    user.verify_HOTP(value.encode())
except AuthenticationError as error:
    dds_web.dds_web.security.auth.send_hotp_email(user)
    raise error
```

dds_web/database/models.py
```diff
@@ -370,40 +375,42 @@ def verify_reset_token(token):
...
-        """Verify the otp token."""
-        totp = pyotp.TOTP(self.otp_secret)
-        return totp.verify(token, valid_window=1)
+        hotp = twofactor_hotp.HOTP(self.hotp_secret, 8, hashes.SHA512())
+        return hotp.generate(self.hotp_counter)
+
+    def verify_HOTP(self, token):
+        """Verify the HOTP token.
+
+        raises AuthenticationError if token is invalid or has expired (older than 1 hour).
```
worukan, line 398: older than 15 minutes

dds_web/database/models.py
```diff
@@ -370,40 +375,42 @@ def verify_reset_token(token):
...
+        If the token is valid, the counter is incremented, to prohibit re-use.
```
worukan, line 399: Remember to update comment, if the increment will be removed from here

dds_web/web/user.py
```diff
@@ -147,35 +138,103 @@ def register():
...
+def cancel_2fa():
+    form = forms.Cancel2FAForm()
+    # Only checks the csrf I assume (doesn't hurt to validate)
+    if form.validate_on_submit():
+        # Reset 2fa cookie
+        redirect_to_login = flask.redirect(flask.url_for("auth_blueprint.login"))
+        redirect_to_login.set_cookie("2fa_initiated_token", "", expires=0)
+        return redirect_to_login
+
+    return flask.redirect(flask.url_for("auth_blueprint.login"))
```
worukan, line 160: lines 154-160 look a little bit strange. It seems like you could rewrite it even as 

```
form.validate_on_submit()
redirect_to_login = flask.redirect(flask.url_for("auth_blueprint.login"))
        redirect_to_login.set_cookie("2fa_initiated_token", "", expires=0)
        return redirect_to_login
```

because no matter the outcome of validation you redirect to the same endpoint and might as well unset the cookie in any case?
And it seems like validate only returns a boolean, not that it raises an exception. Does it help us in any way?

dds_web/web/user.py
```diff
@@ -193,12 +252,22 @@ def login():
...
-        flask.flash("Logged in successfully.")
+        # Correct credentials still needs 2fa
 
-        # Go to home page
-        return flask.redirect(next or flask.url_for("auth_blueprint.index"))
+        # Send 2fa token to user's email
+        hotp_value = user.generate_HOTP_token()
+        msg = dds_web.utils.create_one_time_password_email(user, hotp_value)
+        mail.send(msg)
+        flask.flash("One-Time Code has been sent to your primary email.")
```
worukan, line 261: Instead of lines 258-261, use 

```
if dds_web.dds_web.security.auth.send_hotp_email(user):
    flask.flash("One-Time Code has been sent to your primary email.")
```

In order for this to work, update in auth.py

```
def send_hotp_email(user):
    if not user.hotp_issue_time or (
        user.hotp_issue_time
        and (dds_web.utils.current_time() - user.hotp_issue_time > datetime.timedelta(minutes=15))
    ):
        hotp_value = user.generate_HOTP_token()
        msg = dds_web.utils.create_one_time_password_email(user, hotp_value)
        mail.send(msg)
        return True
    return False
```

inaod568 (commented):
@worukan 
Apart from my added comments, I also find it strange to have the 15 min limit for the mfa when we have the limiter at 10 requests per hour for the basic authentication.

dds_web/api/schemas/token_schemas.py
worukan: If you don't want this, it means that when a user's hotp verification fails and the hotp time expires, they have to do basic auth themselves again to be able to get a new hotp. This may not be obvious.

dds_web/web/user.py
```diff
@@ -147,35 +138,103 @@ def register():
...
     if next and not dds_web.utils.is_safe_url(next):
         return flask.abort(400)
 
+    if form.validate_on_submit():
+        try:
+            token = flask.request.cookies["2fa_initiated_token"]
+            user = dds_web.security.auth.verify_token(token)
+        except (KeyError, ddserr.AuthenticationError):
+            flask.flash("You have to first supply valid credentials", "danger")
+            return flask.redirect(flask.url_for("auth_blueprint.login", next=next))
```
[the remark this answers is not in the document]
worukan: I think we are not interested in the underlying error messages such as invalid token or expired token. These do not make sense from a browser user's perspective. If there is something wrong with the token or it has expired, they have to supply the valid credentials and they will be redirected to the login page to do that.

dds_web/database/models.py
alneberg: That's done by your PR already.

dds_web/api/schemas/token_schemas.py
inaod568: I thought that's how it should work, at least for the CLI? But if not, could you write a list on how it should work for the web and cli? When someone needs to do basic auth and mfa.

dds_web/web/user.py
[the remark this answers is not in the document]
inaod568: Sure, but for example if I were to log into a system with username and password, and then for some reason I wait too long to input the hotp, then I just get a message saying "You have to first supply valid credentials" - I would personally find this strange and think "but I did that, something must be wrong with this system". So I would like us to have a different message, maybe a concatenation of the underlying message and this message.

dds_web/api/schemas/token_schemas.py
alneberg: Then I can remove the "mfa_verified" and "mfa_method" as well right? I'll go for that at least

dds_web/database/models.py
alneberg: Without this, one could use the same HOTP token twice, as long as a new one has not been requested. While it might not be that easy to actually do it. It's very easy to use this to prevent it completely.

dds_web/database/models.py
alneberg: I changed this occurrence. Searching for "time password" only found this one. Searching for "password" obviously finds more, so harder to check.

dds_web/web/user.py
alneberg: I opted for Inas solution.

dds_web/web/user.py
alneberg: Yes I think this is a good addition!

dds_web/database/models.py
inaod568: Yes. Keep this.

dds_web/api/schemas/token_schemas.py
worukan: OK, so I had a look at how the new cli behaviour is and it can prompt the user with "Second factor authentication failed, would you like to try again?". So I have a new suggestion, instead of making a change here, let's take `send_hotp_email(user)` from line 105 to line 101 in auth.py. This way a new value will be generated if needed, the first try to verify hotp will fail (it will fail anyway because it has been more than 15 mins already), but at least they get the email as hotp verify fails and the cli will prompt to try again. How about this?

dds_web/web/user.py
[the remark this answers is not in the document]
worukan: perhaps "Supply valid credentials again due to timeout" or something like that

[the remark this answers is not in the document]
inaod568: Yes something like that would improve a lot!

dds_web/api/schemas/token_schemas.py
inaod568: Sounds good!

tests/test_basic_api.py
```diff
@@ -62,137 +58,159 @@ def test_auth_incorrect_username_and_password_check_statuscode_401_incorrect_inf
...
 
-def test_auth_token_wrong_secret_key_signed_token(client):
-    """Test that an encrypted token signed with the wrong key returns unauthorized"""
+# Token Authentication ###################################################### Token Authentication #
+def test_auth_token_mfa_missing(client):
+    pass
+
+
+def test_auth_token_mfa_inside_token_expired(client):
+    pass
```
inaod568, line 179: Remove? Or adding more later?

inaod568 (approved):
🎉

inaod568:
I'm approving after review and merging this now since @worukan had looked through and was ready to approve apart from two test related files.

PR #780, opened 2022-01-28
Use MAIL_DEFAULT_SENDER instead of MAIL_SENDER and MAIL_SENDER_ADDRESS

talavis:
* Set `MAIL_DEFAULT_SENDER` in `config.py` instead of setting `sender` for emails
  - If an email needs a custom sender, `sender` can still be used
* Remove the duplicate settings `MAIL_SENDER` and `MAIL_SENDER_ADDRESS`

### DIFF
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -193,7 +193,6 @@ class AddUser(flask_restful.Resource):
 
         msg = flask_mail.Message(
             subject,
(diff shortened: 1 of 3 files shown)


alneberg (approved):
Yes, looks good. I'll let Ina review/merge though.

inaod568 (approved):
Yes! Approved!

Issue #770, opened 2022-01-17
Add "disable user" functionality


inaod568:
Add possibility to activate/disable a user using `is_active` column in User table, etc.  
The `is_active` column comes with the `UserMixin` inheritance already implemented.

PR #784, opened 2022-01-28
Add User activation endpoint 

aanil:
Addresses https://github.com/ScilifelabDataCentre/dds_web/issues/770

* Added field `active` and property `is_active` to User model
* Invited user is activated at the time of registering.
* Authentication checks whether user is active.
* Super Admin can deactivate/reactivate all user and Unit admins can do the same for unit users. All other cases are forbidden.
* Users cannot deactivate/reactivate themselves.

### DIFF
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -456,16 +456,9 @@ class UserActivation(flask_restful.Resource):
             action == "deactivate" and not user.is_active
         ):
             raise ddserr.DDSArgumentError(message=f"User is already in desired state!")
-        import pdb
-
-        pdb.set_trace()
-        if action == "reactivate":
-            activate = True
-        else:
-            activate = False
 
         try:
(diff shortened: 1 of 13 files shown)


dds_web/api/user.py
```diff
@@ -425,6 +425,67 @@ def delete(self):
...
+        ):
+            raise ddserr.DDSArgumentError(message=f"User is already in desired state!")
+
+        if action == "reactivate":
+            activate = True
+        else:
+            activate = False
+
+        try:
+            user.active = activate
```
inaod568, line 466: I'd like to simplify with this
```
user.active = action == "reactivate"
```

inaod568 (commented):
It would be great if you could also add some `TODO` comment tags about having to update the users project access.

inaod568 (approved):
Works well! Approving!

Issue #789, opened 2022-02-01
Check if timestamp timezone bug in token and if so fix


inaod568:

MatthiasZepper:
After the ongoing refactoring of the token endpoints are completed, we should test if changing the timezone of the Docker containers has unwanted side-effects in terms of token validity and lifetime. 

Since we create from and check against the `utils.current_time()` this shouldn't be the case:

https://github.com/ScilifelabDataCentre/dds_web/blob/1083347d3eecd404645ad58c6db2913c2411b6bf/dds_web/api/user.py#L92

https://github.com/ScilifelabDataCentre/dds_web/blob/1083347d3eecd404645ad58c6db2913c2411b6bf/dds_web/security/auth.py#L79-L82

But we use the `.timestamp()` and `.fromtimestamp()` functions to de-serialize and serialize the datetime from and to JSON, which looses the timezone information contained into the datetime object.  

Therefore, I used the isodate package that [Microsoft engineers recommend when using the Azure SDK for Python](https://github.com/ScilifelabDataCentre/dds_web/blob/1083347d3eecd404645ad58c6db2913c2411b6bf/dds_web/) in my draft pull requests #790 and [227](https://github.com/ScilifelabDataCentre/dds_cli/pull/227).

PR #790, opened 2022-02-01
Embeds the tokens lifetime as isodate encoded timedelta

MatthiasZepper:
This PR suggests changes to the token's header to embed some metadata about the token to be openly readable by the CLI:

- `iat`: The _issued at_ value is a registered claim according to JWT specification and identifies the time at which the JWT was issued. This claim can be used to determine the age of the JWT. Its value MUST be a number containing a NumericDate value. 
- `exp`: The _expiration time_ value is a registered claim according to JWT specification and identifies the expiration time on or after which the JWT MUST NOT be accepted for processing. The processing of the “exp” claim requires that the current date/time MUST be before the expiration date/time listed in the “exp” claim. Implementers MAY provide for some small leeway, usually no more than a few minutes, to account for clock skew. Its value MUST be a number containing a NumericDate value.
- `csg`: The _consignee_ value is **not** a registered claim according to JWT specification, but was used to include the username.

### DIFF
--- a/dds_web/security/tokens.py
+++ b/dds_web/security/tokens.py
@@ -33,6 +33,7 @@ def encrypted_jwt_token(
             "alg": "A256KW",
             "enc": "A256GCM",
             "exp": (dds_web.utils.current_time() + expires_in).timestamp(),
+            "iat": (dds_web.utils.current_time()).timestamp(),
             "csg": username,
         },
         claims=__signed_jwt_token(
@@ -85,6 +86,7 @@ def __signed_jwt_token(
         header={
             "alg": "HS256",
             "exp": expiration_time.timestamp(),
+            "iat": (dds_web.utils.current_time()).timestamp(),
             "csg": username,
         },
         claims=data,
--- a/dds_web/security/tokens.py
+++ b/dds_web/security/tokens.py
@@ -32,8 +32,11 @@ def encrypted_jwt_token(
         header={
             "alg": "A256KW",
             "enc": "A256GCM",
+            # exp: This registered claim according to JWT specification identifies the time on or after which the JWT MUST NOT be accepted for processing.
             "exp": (dds_web.utils.current_time() + expires_in).timestamp(),
+            # iat: issued at value is a registered claim and identifies the time at which the token was issued.
             "iat": (dds_web.utils.current_time()).timestamp(),
+            # csg: consignee claim for transmitting the user name to which the token was issued. Is not registered according to JWT specification.
             "csg": username,
         },
         claims=__signed_jwt_token(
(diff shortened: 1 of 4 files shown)


inaod568:
Closed and opened because of mistake. 
Closed again because: Yesterday we discussed that we should not change to / use the `isodate` package because the expiration-bug was not reproducable, and also because we decided to not make changes in the tokens until the MFA PR had been merged. Kept it was a connected issue though to get back to it later.

MatthiasZepper:
Since the MFA PR has a higher priority, this was just opened as a draft, but you may of course close it temporarily (or even permanently) to prevent cluttering & confusion. 

I won't insist on having that feature. I thought it would be nice if the token would "tell" it's lifetime to the CLI instead of relying on some CLI config value that says "Tokens are valid for 48h" [see PR227 of the CLI](https://github.com/ScilifelabDataCentre/dds_cli/pull/227), but it was indeed just a minor byproduct of the investigation regarding #789. 

I used `isodate`, because to my knowledge `.timestamp()` can only convey fixed time points but no timedeltas, but I might be wrong about that.

inaod568:
I agree that this is a nice feature so I think the closing of this is definitely not permanent.

inaod568:
https://github.com/ScilifelabDataCentre/dds_cli/pull/227#issuecomment-1033479107

requirements.txt
```diff
@@ -12,6 +12,7 @@ Flask-Security-Too==4.1.0
 Flask-SQLAlchemy==2.5.1
 Flask-WTF==0.15.1
 immutabledict==2.2.0
+isodate
```
inaod568, line 15: Add a version, and fix the conflicts

dds_web/security/tokens.py
```diff
@@ -29,7 +29,13 @@ def encrypted_jwt_token(
     :param Dict or None additional_claims: Any additional token claims can be added. e.g., {"iss": "DDS"}
     """
     token = jwt.JWT(
-        header={"alg": "A256KW", "enc": "A256GCM"},
+        header={
+            "alg": "A256KW",
+            "enc": "A256GCM",
+            "exp": (dds_web.utils.current_time() + expires_in).timestamp(),
+            "iat": (dds_web.utils.current_time()).timestamp(),
```
inaod568, line 36: Would be great if we could att clarification comments regarding what the different headers stand for, for the future

inaod568:
Just adding a reminder/info here, from the `dds-development` channel: 
Add an `-y` to the docker file so that it says `apt-get upgrade -y`

inaod568 (approved):

PR #803, opened 2022-02-03
added projectkeys table from volkans PR

inaod568:
I just copied the new table and relationships from @worukans PR #796 
This way we get the `ProjectKeys` table into the dev branch and then we can build on it from there


worukan (approved):
ok

PR #772, opened 2022-01-17
For draft pr, not fully working yet

worukan:

PR #806, opened 2022-02-03
create new project with temporary keys etc

inaod568:
Copy pasted the code from @worukans PR #772, just some slight changes in order for it to work.
Added a temporary key etc, and also kept the old project keys in the same place as before - that way nothing breaks and we can remove it later.

### DIFF
--- a/dds_web/api/schemas/project_schemas.py
+++ b/dds_web/api/schemas/project_schemas.py
@@ -155,10 +155,11 @@ class CreateProjectSchema(marshmallow.Schema):
             )
             # Generate keys
             # NOTE: TEMPORARY:
+            # key=b"temp" * 8 is only a temporary KEK until we have the actual KEK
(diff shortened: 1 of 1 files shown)


worukan:
If we keep both the old and the new keys' code together, the issue is that the new public key will be lost, because there is one public_key attribute and it is written with the old one.

dds_web/api/schemas/project_schemas.py
```diff
@@ -135,8 +135,10 @@ def create_project(self, data, **kwargs):
                 public_id=data["public_id"], created_time=data["date_created"]
             )
 
-            # Generate keys
+            # NOTE: TEMPORARY
+            # Generate keys and add to project
             data.update(**key_gen.ProjectKeys(data["public_id"]).key_dict())
```
worukan, line 140: This is old key_gen code that generates a public and private key pair and its product is saved into the project model

worukan (approved):

PR #807, opened 2022-02-03
engine dispose for tests

alneberg:
I didn't have time to wait for the tests to finish locally. But I think this should fix it. Thanks to @talavis for finding the solution!


inaod568 (approved):
Nice, now it works locally for me too.

PR #808, opened 2022-02-04
Remove unused config values and folder binds

talavis:
* Remove unused config variables:
  - `LOCAL_TEMP_CACHE`
  - `DOWNLOAD_FOLDER`
  - `UPLOAD_FOLDER`
* Remove corresponding folder binds in `docker-compose.yml`
* Change the value of `MAX_CONTENT_LENGTH` to hex to make it clearer
* Remove reference to `.dds-cli.json` in `docker-compose.yml`

### DIFF
--- a/dds_web/config.py
+++ b/dds_web/config.py
@@ -25,7 +25,7 @@ class Config(object):
     SQLALCHEMY_TRACK_MODIFICATIONS = False
 
     # Data related config
-    MAX_CONTENT_LENGTH = 16777216
+    MAX_CONTENT_LENGTH = 0x1000000
     MAX_DOWNLOAD_LIMIT = 1000000000
 
     # Expected paths - these are the bind paths *inside* the container
--- a/dds_web/config.py
+++ b/dds_web/config.py
@@ -31,7 +31,6 @@ class Config(object):
     # Expected paths - these are the bind paths *inside* the container
     USE_LOCAL_DB = True
     LOGS_DIR = "/dds_web/logs"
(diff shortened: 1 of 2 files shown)


dds_web/config.py
```diff
@@ -25,15 +25,12 @@ class Config(object):
...
+    MAX_CONTENT_LENGTH = 0x1000000
     MAX_DOWNLOAD_LIMIT = 1000000000
 
     # Expected paths - these are the bind paths *inside* the container
     USE_LOCAL_DB = True
     LOGS_DIR = "/dds_web/logs"
-    UPLOAD_FOLDER = "/dds_web/uploads"
-    DOWNLOAD_FOLDER = "/dds_web/downloads"
-    LOCAL_TEMP_CACHE = "/dds_web/local_temp_cache"
     DDS_S3_CONFIG = "/code/dds_web/sensitive/s3_config.json"
```
alneberg, line 34: @inaod568 we can remove the `DDS_S3_CONFIG` as well right?

alneberg (approved):
Looks correct to me!

dds_web/config.py
talavis: That should mean that I can remove `s3_config.json? as well from the comment in `docker_compose.yml`?

alneberg: Yes it would. I think we simply moved the values from that one into the general config file.

inaod568: Yes 👍🏻

inaod568 (approved):

PR #809, opened 2022-02-04
Worukan security projectkeys module

inaod568:
To reduce the complexity of all coming PRs I've added the part of @worukans PR where the project keys are generated.
This should mean that the upload and download works for the person creating the project, so we can still test that, but it does mean that adding users to projects and downloading by anyone else becomes "broken" (they don't actually get access and can't download etc). In my opinion this is better and that we build on the rest, instead of having a single big PR with all features fixed. Huge risk of confusion and mistakes.

This code includes project private encryption with the password-derived key. This is wrong and is a temporary thing until the user keys are implemented. The passwords are hard coded which should not matter at this stage since all passwords are "password". Adding the cache in coming PRs and will then replace this part.

### DIFF
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -229,10 +229,10 @@ class GetPrivate(flask_restful.Resource):
     @logging_bind_request
     def get(self):
         """Get private key from database"""
-        flask.current_app.logger.debug("Getting the private key.")
-
         # Verify access to project
         project = project_schemas.ProjectRequiredSchema().load(flask.request.args)
+
+        flask.current_app.logger.debug("Getting the private key.")
 
         # Get Project keys row for current user and project
         project_keys_row = models.ProjectKeys.query.filter_by(
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -38,7 +38,7 @@ from dds_web.errors import (
 from dds_web.api.user import AddUser
 from dds_web.api.schemas import project_schemas
 from dds_web.api.schemas import user_schemas
-
+from dds_web.security import project_keys
 
 ####################################################################################################
 # ENDPOINTS ############################################################################ ENDPOINTS #
@@ -229,36 +229,28 @@ class GetPrivate(flask_restful.Resource):
     @logging_bind_request
     def get(self):
         """Get private key from database"""
-
-        project = project_schemas.ProjectRequiredSchema().load(flask.request.args)
-
-        # TODO (ina): Change handling of private key -- not secure
         flask.current_app.logger.debug("Getting the private key.")
 
-        app_secret = flask.current_app.config.get("SECRET_KEY")
-        passphrase = app_secret.encode("utf-8")
+        # Verify access to project
+        project = project_schemas.ProjectRequiredSchema().load(flask.request.args)
 
-        enc_key = bytes.fromhex(project.private_key)
-        nonce = bytes.fromhex(project.privkey_nonce)
-        salt = bytes.fromhex(project.privkey_salt)
+        # Get Project keys row for current user and project
+        project_keys_row = models.ProjectKeys.query.filter_by(
+            project_id=project.id, user_id=auth.current_user().username
+        ).one_or_none()
+        if not project_keys_row:
+            raise AccessDeniedError("Project key missing.")
 
-        kdf = scrypt.Scrypt(
-            salt=salt,
-            length=32,
-            n=2**14,
-            r=8,
-            p=1,
-            backend=backends.default_backend(),
+        # Return the decrypted project private key
+        return flask.jsonify(
+            {
+                "private": project_keys.obtain_project_private_key(
+                    user=auth.current_user(), project_key=project_keys_row
+                )
+                .hex()
+                .upper()
+            }
         )
-
-        key_enc_key = kdf.derive(passphrase)
-        try:
-            decrypted_key = decrypt(ciphertext=enc_key, aad=None, nonce=nonce, key=key_enc_key)
-        except Exception as err:
-            flask.current_app.logger.exception(err)
-            raise KeyNotFoundError
-
-        return flask.jsonify({"private": decrypted_key.hex().upper()})
 
 
 class UserProjects(flask_restful.Resource):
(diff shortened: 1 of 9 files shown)


tests/test_project_status.py
```diff
@@ -30,9 +30,9 @@
     "description",
     "pi",
     "public_key",
-    "private_key",
-    "privkey_salt",
-    "privkey_nonce",
+    # "private_key",
+    # "privkey_salt",
```
alneberg, line 34: Should these just be removed instead?

tests/test_project_status.py
```diff
@@ -90,6 +90,8 @@ def test_set_project_to_deleted_from_in_progress(module_client, boto3_session):
     assert file_in_db(test_dict=first_new_file, project=project.id)
 
     for field, value in vars(project).items():
+        print(field)
+        print(value)
```
alneberg, line 94: I assume these print statements should be removed

tests/test_project_status.py
inaod568: Done!

dds_web/security/project_keys.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def manage_project_key_among_users(existing_user, current_user, project_key):
+    return encrypt_project_key(existing_user, obtain_project_private_key(current_user, project_key))
+
+
+def obtain_project_private_key(user, project_key):
+    """Decrypt the project private key with the user specific KEK."""
+    # password = dds_web.cache.get(user.username)
+    password = "password"
```
alneberg, line 49: This one warrants a `TODO: please fix` don't you think?

inaod568: Sure 👍🏻

alneberg (approved):
Yes, I think it makes sense to merge incomplete code like this at this point as we need to enable other tasks and we're changing this code rapidly now. We just need to make sure we have a proper code review in the end.

inaod568:
@alneberg Agreed.

dds_web/api/project.py
```diff
@@ -229,37 +229,29 @@ class GetPrivate(flask_restful.Resource):
     @logging_bind_request
     def get(self):
         """Get private key from database"""
-
-        project = project_schemas.ProjectRequiredSchema().load(flask.request.args)
-
-        # TODO (ina): Change handling of private key -- not secure
         flask.current_app.logger.debug("Getting the private key.")
```
worukan, line 232: I think this was at a better place before, because if ProjectRequiredSchema is not satisfied, no further action is taken so not "Getting the private key."

inaod568: Yeah, fair enough! Changing

dds_web/api/project.py
```diff
@@ -229,37 +229,29 @@ class GetPrivate(flask_restful.Resource):
...
-
-        enc_key = bytes.fromhex(project.private_key)
-        nonce = bytes.fromhex(project.privkey_nonce)
-        salt = bytes.fromhex(project.privkey_salt)
+        # Get Project keys row for current user and project
+        project_keys_row = models.ProjectKeys.query.filter_by(
+            project_id=project.id, user_id=auth.current_user().username
+        ).one_or_none()
+        if not project_keys_row:
+            raise AccessDeniedError("Project key missing.")
```
worukan, line 242: If-checks make the code a bit crowded. There is a KeyNotFoundError implemented in obtain_project_private_key. We could perhaps rewrite that better, if you like. So I think lines 241-242 could be removed.

dds_web/database/models.py
```diff
@@ -355,6 +355,8 @@ def __init__(self, **kwargs):
         super(User, self).__init__(**kwargs)
         if not self.hotp_secret:
             self.hotp_secret = os.urandom(20)
+        if not self.kd_salt:
+            self.kd_salt = os.urandom(32)
```
worukan, line 359: I recommend removing lines 358-359, because a salt is paired with a password and there are places where a password is set or changed, and those places are where we set the salt. It is not something that exists on its own or for a lifetime.

dds_web/api/project.py
inaod568: @worukan I like the exception you added too but I'd like to keep this if statement for now at least. It's more explicit. It's possible we will remove this though but for now we can keep it.

dds_web/security/project_keys.py
```diff
@@ -0,0 +1,146 @@
...
+        flask.current_app.logger.exception(exception)
+        raise exception
+
+    # Derive the key
+    derived_key = argon2.low_level.hash_secret_raw(
+        secret=password.encode(),
+        salt=user.kd_salt,
+        time_cost=2,
+        # memory_cost=4194304, # This makes it crash locally
+        memory_cost=1000,  # Just changed to something so that it wouldn't crash
```
worukan, line 27: I recommend we create an issue that is for setting a config for development purposes. In that config, we could set a small value for the key derivation memory cost, disable hotp or other features that we don't use or use in a limited way on a day to day basis. If this breaks the build, perhaps we should empower our build environment, if possible.

inaod568: https://github.com/ScilifelabDataCentre/dds_web/issues/812#issue-1124288020

dds_web/database/models.py
inaod568: Ok removed this and put the salt generation in diffrent places in the code. We can come up with a better solution as we go

inaod568:
We're going to start with adding the user keys to the table and then build on that instead. Will open this again if we change the plan.

dds_web/security/project_keys.py
inaod568: @talavis Could you set a config value for development regarding the `memory_cost` mentioned here?

talavis: Just to mention: `memory cost` is in kibibytes, so `memory_cost=4194304`means that you want to use 4GiB of RAM. In the case of my docker setup, that's beyond my resource limits.

worukan: It is 4GiB of RAM. I too had the issue of running out of memory, but changed computer so no issue for me right now. I followed the documentation and the help of a calculator. How would you write it?

PR #814, opened 2022-02-04
Send email to researchuser when project is released

aanil:
*Email sent to all registered emails of research users when project is made available.

Haven't added links to project in the email template.

### DIFF
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -103,51 +103,7 @@ class AddUser(flask_restful.Resource):
         link = flask.url_for("auth_blueprint.confirm_invite", token=token, _external=True)
 
         # Compose and send email
-        unit_name = None
-        if auth.current_user().role in ["Unit Admin", "Unit Personnel"]:
-            unit = auth.current_user().unit
-            unit_name = unit.external_display_name
-            unit_email = unit.contact_email
-            sender_name = auth.current_user().name
-            subject = f"{unit_name} invites you to the SciLifeLab Data Delivery System"
-        else:
-            sender_name = auth.current_user().name
-            subject = f"{sender_name} invites you to the SciLifeLab Data Delivery System"
-
-        msg = flask_mail.Message(
-            subject,
-            recipients=[new_invite.email],
-        )
-
-        # Need to attach the image to be able to use it
-        msg.attach(
-            "scilifelab_logo.png",
-            "image/png",
-            open(
-                os.path.join(flask.current_app.static_folder, "img/scilifelab_logo.png"), "rb"
-            ).read(),
-            "inline",
-            headers=[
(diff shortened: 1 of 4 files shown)


aanil:
We should think about allowing Units to have their own personalised email templates that they can sent out when projects are released. These could be done by having a flag to suppress the automatic sending of emails, so that they can use their existing notification procedures(for eg. over Zendesk.).

dds_web/api/user.py
```diff
@@ -232,6 +195,76 @@ def add_user_to_project(existing_user, project, role):
...
+        if mail_type == "invite":
+            subject = f"{subject_subject} invites you to the SciLifeLab Data Delivery System"
+            recepients = [userobj.email]
+        elif mail_type == "project_release":
+            subject = f"Project made available by {subject_subject} in the SciLifeLab Data Delivery System"
+            recepients = [x.email for x in userobj.emails]
+            project_id = project.public_id
+            deadline = project.current_deadline.astimezone(datetime.timezone.utc).strftime(
+                "%Y-%m-%d %H:%M:%S %Z"
+            )
```
inaod568, line 227: Either add an `else` or have `if mail_type not in ["invite", "project_release"]: Raise <something>`

dds_web/api/user.py
```diff
@@ -232,6 +195,76 @@ def add_user_to_project(existing_user, project, role):
...
+        # Compose and send email
+        unit_name = None
+        project_id = None
+        deadline = None
+        if auth.current_user().role in ["Unit Admin", "Unit Personnel"]:
+            unit = auth.current_user().unit
+            unit_name = unit.external_display_name
+            unit_email = unit.contact_email
+            sender_name = auth.current_user().name
+            subject_subject = unit_name
```
inaod568, line 211: Doing this should only be possible for Unit Admin and Unit Personnel roles right? Or Super Admin later on I guess.

dds_web/api/user.py
aanil: Since the arguments for this function will only come from within the code, do we really need it?

dds_web/api/user.py
aanil: This was already in the code for the invite, I haven't changed that bit.

dds_web/api/user.py
inaod568: I think we should add it just in case, since this function can be called from outside, and it's just better to not leave anything to chance. 👍🏻

dds_web/api/user.py
inaod568: We can leave this for now, I need to check something.

dds_web/api/user.py
aanil: Done!

inaod568 (approved):

PR #816, opened 2022-02-05
Slight refactoring

inaod568:
* Removed `flask.jsonify`
* Added `api.representation` to automatically return json for all resources
* Altered some messages so that the lines are shorter
* Added `code` to all custom HTTPExceptions in the `errors.py` file
* In `api/user.py` I did some slight refactoring to reduce the number of if-else, I feel this made it a bit easier to follow along


tests/test_user_remove_association.py
```diff
@@ -60,8 +60,10 @@ def test_remove_not_associated_user_from_project(client, boto3_session):
         data=json.dumps(rem_user),
         content_type="application/json",
     )
-    assert response.status_code == http.HTTPStatus.OK
-    assert "User already not associated with this project" in response.json["message"]
+    assert response.status_code == http.HTTPStatus.BAD_REQUEST
+    assert (
+        f"{rem_user['email']} already not associated with this project" in response.json["message"]
```
alneberg, line 65: I think this can be rephrased while you're at it? Also surprised that the status code changed, I agree that it makes sense the way you have it now though.

inaod568: Sure. Better now?

PR #823, opened 2022-02-07
Add encryption to mariadb

### LINKED ISSUE #700: Make sure the database is encrypted
Talk to Hamza / Linus

talavis:
* Add config for using encryption-at-rest for mariadb
* Add keys to use for development
* Use the above files in `docker-compose.yml`

Key generation:
```
for i in {1..100} ; do echo "$i;"$(openssl rand -hex 32) >> keys.txt ; done
openssl rand -hex 128 > pass.key
openssl enc -aes-256-cbc -md sha1 -pass file:pass.key -in keys.txt -out keys.enc
```

If setup is correct and similar setup is used for real deployment, it should close #700


Dockerfiles/mariadb/db-encrypt/keys.txt
```diff
@@ -0,0 +1,100 @@
+1;58b22a41c9d095b8003995cf726f852a2c2bd44e4be5da2f52f643d7559c574d
```
worukan, line 1: I think this file is not needed as you have encrypted this into keys.enc.

Dockerfiles/mariadb/mariadb.conf
```diff
@@ -0,0 +1,17 @@
+[mariadb]
+# File Key Management
+plugin_load_add = file_key_management
+file_key_management_filename = /etc/mysql/encrypt/keys.enc
+file_key_management_filekey = FILE:/etc/mysql/encrypt/pass.key
+file_key_management_encryption_algorithm = AES_CBC
```
worukan, line 6: The recommended one is stated as AES_CTR. Is there a reason that prevents us from using it?

talavis: Nothing apart from my reading comprehension.

Dockerfiles/mariadb/mariadb.conf
```diff
@@ -0,0 +1,17 @@
...
+
+# InnoDB/XtraDB Encryption Setup
+innodb_default_encryption_key_id = 1
+innodb_encrypt_tables = FORCE
+innodb_encrypt_temporary_tables = ON
+innodb_encrypt_log = ON
+innodb_encryption_threads = 4
+
+encrypt-tmp-disk-tables = 1
+encrypt-tmp-files = 1
```
worukan, line 16: According to documentation, this is written as `encrypt_tmp_files=ON`
[https://mariadb.com/kb/en/server-system-variables/#encrypt_tmp_files]

Dockerfiles/mariadb/mariadb.conf
```diff
@@ -0,0 +1,17 @@
...
+file_key_management_encryption_algorithm = AES_CTR
+
+# InnoDB/XtraDB Encryption Setup
+innodb_default_encryption_key_id = 1
+innodb_encrypt_tables = FORCE
+innodb_encrypt_temporary_tables = ON
+innodb_encrypt_log = ON
+innodb_encryption_threads = 4
+
+encrypt-tmp-disk-tables = 1
```
worukan, line 15: According to documentation, this is written as `encrypt_tmp_disk_tables=ON`
[https://mariadb.com/kb/en/server-system-variables/#encrypt_tmp_disk_tables]
By the way there is an **if** in its description. Are we using aria for temp tables?

Dockerfiles/mariadb/mariadb.conf
```diff
@@ -0,0 +1,17 @@
+[mariadb]
+# File Key Management
+plugin_load_add = file_key_management
+file_key_management_filename = /etc/mysql/encrypt/keys.enc
+file_key_management_filekey = FILE:/etc/mysql/encrypt/pass.key
+file_key_management_encryption_algorithm = AES_CTR
+
+# InnoDB/XtraDB Encryption Setup
+innodb_default_encryption_key_id = 1
```
worukan, line 9: This is already set to 1 in defaults.
[https://mariadb.com/kb/en/innodb-system-variables/#innodb_default_encryption_key_id] 
Why do we have this defined here?
You have created 100 keys and it is always being set to use the first key. I couldn't understand everything yet, but reading documentation. Is this for development environment?

Dockerfiles/mariadb/mariadb.conf
talavis: A mix-up of CLI vs conf file

Dockerfiles/mariadb/mariadb.conf
talavis: A mix-up of CLI vs conf file

Dockerfiles/mariadb/db-encrypt/keys.txt
talavis: I agree, but decided to keep it to make it clear what's going on. These files are mostly intended as a reference to show what is needed in production, and thus I thought it made sense to include the unencrypted file to remind me about the needed formatting etc.

Dockerfiles/mariadb/mariadb.conf
talavis: It is for dev. That specific setting is a remnant of my testing and should be removed.

Dockerfiles/mariadb/mariadb.conf
talavis: We use Innodb, so not needed.

Dockerfiles/mariadb/db-encrypt/keys.txt
talavis: Also: technically only one key is needed, but I chose to follow their instructions by the letter when I set it up. It seems it chooses to use a different key for e.g. tmp files, so I guess five or so is probably a good number. The plugin does not support key rotation, so no need to keep them for that.

Dockerfiles/mariadb/mariadb.conf
talavis: I checked; the old spelling worked as well, I guess that's why I never noticed.

Dockerfiles/mariadb/db-encrypt/keys.txt
worukan: Yes I too saw that key rotation is not available and the key ids don't need to be contiguous. Perhaps you can reduce the number of keys?
I think documentation is a separate work, which can be kept either next to the actual thing or at a separate place. Since these are keys, I think it is better to keep it with its(your, not the online one) documentation elsewhere, so that you don't run into the issue of remembering whether it was needed or not to be in this place. This is a suggestion. You run this, so final choice is yours.

Dockerfiles/mariadb/mariadb.conf
```diff
@@ -0,0 +1,15 @@
...
+file_key_management_filename = /etc/mysql/encrypt/keys.enc
+file_key_management_filekey = FILE:/etc/mysql/encrypt/pass.key
+file_key_management_encryption_algorithm = AES_CTR
+
+# InnoDB Encryption Setup
+innodb_encrypt_tables = FORCE
+innodb_encrypt_temporary_tables = ON
+innodb_encrypt_log = ON
+innodb_encryption_threads = 4
+
```
worukan, line 13: Do you know and understand the following one?
[https://mariadb.com/kb/en/innodb-system-variables/#innodb_encryption_rotate_key_age]
It is given as `innodb_encryption_rotate_key_age = 1` in basic configuration
[https://mariadb.com/kb/en/innodb-encryption-overview/]
This is confusing as file_key_management plugin doesn't support key rotation. Does this mean that even if the plugin doesn't do it, innodb engine can do it???
What do you think?

talavis: Innodb supports it, but the file key plugin does not. E.g. the AWS plugin does support key rotation.

worukan: ok now that I look at the AWS plugin, it makes sense.

worukan (approved):
We have covered a lot of details together with its documentation. It looks good. You keep the plaintext keys at the place of your choice as it is meant for the development environment.

inaod568 (approved):

PR #827, opened 2022-02-08
Invite with encrypted token and test-TKEK

alneberg:
I've changed the invite to use an encrypted token where a "TKEK" variable can be defined. This PR mainly concerns handling of the encrypted token and storing it in the flask.session during the registration. The flask.session uses a signed cookie so it should not be possible to change the contents of it: https://flask.palletsprojects.com/en/2.0.x/api/?highlight=flash#sessions I've also added an extra check to make sure that the email supplied with the form is the same as the one in the token.

Left to be done with invites after this PR:
 - Create columns to store invites key pair
 - Create table ProjectInviteKeys
 - Create and use a real TKEK


inaod568:
> I've changed the invite to use an encrypted token where a "TKEK" variable can be defined. This PR mainly concerns handling of the encrypted token and storing it in the flask.session during the registration. The flask.session uses a signed cookie so it should not be possible to change the contents of it: https://flask.palletsprojects.com/en/2.0.x/api/?highlight=flash#sessions I've also added an extra check to make sure that the email supplied with the form is the same as the one in the token.
> 
> Left to be done with invites after this PR:
> 
> * Create columns to store invites key pair
> * Create table ProjectInviteKeys
> * Create and use a real TKEK

Created cards #828 #829 #830

dds_web/security/auth.py
```diff
@@ -69,21 +69,50 @@ def get_user_roles_common(user):
...
 def verify_token_no_data(token):
-    user, claims = __verify_general_token(token)
+    claims = __verify_general_token(token)
+    user = user_from_subject(claims.get("sub"))
     del claims
     gc.collect()
     return user
 
 
+def verify_invite_token_no_data(token):
```
inaod568, line 79: `verify_invite_token_no_data` -- just checking that I understand: this is refering to that it's checking that there is enough data in the invite token? I think a name change would be good in that case. Not a deal breaker though

inaod568:
@alneberg Only a few small comments, no dealbreakers. Looks good I think, so I'm just going to wait until Volkan has reviewed.

dds_web/security/auth.py
alneberg: No, it's to indicate that it doesn't return any of the sensitive data inside the token.

inaod568: Ah ok! Could we change to `verify_invite_token_no_sensdata`?

alneberg: Done!

dds_web/api/user.py
```diff
@@ -88,12 +90,29 @@ def invite_user(args):
...
 
         # Create link for invitation email
         link = flask.url_for("auth_blueprint.confirm_invite", token=token, _external=True)
 
+        # Quick search gave this as the URL length limit.
+        if len(link) >= 2048:
+            flask.current_app.logger.error(
+                "Invitation link was not possible to create due to length."
+            )
+            return {"message": "Invite failed due to server error", "status": 500}
```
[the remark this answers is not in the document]
alneberg: Changed!

dds_web/api/schemas/user_schemas.py
```diff
@@ -190,6 +202,8 @@ def make_user(self, data, **kwargs):
 
         db.session.add(new_user)
 
+        # Use the TKEK here
```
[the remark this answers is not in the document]
alneberg: Fixed!

dds_web/security/auth.py
```diff
@@ -112,17 +141,20 @@ def __verify_general_token(token):
...
+        return data
 
     raise AuthenticationError(message="Expired token")
 
 
+def user_from_subject(subject):
+    if subject:
+        user = models.User.query.get(subject)
+    if user and user.is_active:
+        return user
```
worukan, line 153: lines 152-153 should be indented once to the right

dds_web/security/auth.py
```diff
@@ -69,21 +69,50 @@ def get_user_roles_common(user):
...
 def verify_token(token):
-    user, data = __verify_general_token(token)
-    return handle_multi_factor_authentication(user, data.get("mfa_auth_time"))
+    claims = __verify_general_token(token)
+    user = user_from_subject(claims.get("sub"))
+
+    return handle_multi_factor_authentication(user, claims.get("mfa_auth_time"))
 
 
 def __verify_general_token(token):
```
worukan, line 113:
```
def __verify_general_token(token):
    """
    Verifies the format, signature and expiration time of an encrypted and signed JWT token.
    Raises AuthenticationError if token is invalid or absent, could raise other exceptions from dependencies.
    On successful verification, it returns a dictionary of the claims in the token.
    """
    try:
        data = (
            verify_token_signature(token)
            if token.count(".") == 2
            else decrypt_and_verify_token_signature(token)
        )
    except (ValueError, jwcrypto.common.JWException) as e:
        # ValueError is raised when the token doesn't look right (for example no periods)
        # jwcryopto.common.JWException is the base exception raised by jwcrypto,
        # and is raised when the token is malformed or invalid.
        flask.current_app.logger.exception(e)
        raise AuthenticationError(message="Invalid token")
    except Exception:
        if not token:
            raise AuthenticationError(message="No token")

    expiration_time = data.get("exp")
    # we use a hard check on top of the one from the dependency
    # exp shouldn't be before now no matter what
    if expiration_time and (
        dds_web.utils.current_time() <= datetime.datetime.fromtimestamp(expiration_time)
    ):
        return data

    raise AuthenticationError(message="Expired token")
```

worukan, line 113:
```
def __verify_general_token(token):
    """
    Verifies the format, signature and expiration time of an encrypted and signed JWT token.
    Raises AuthenticationError if token is invalid or absent, could raise other exceptions from dependencies.
    On successful verification, it returns a dictionary of the claims in the token.
    """
    if not token:
        raise AuthenticationError(message="No token")
    try:
        data = (
            verify_token_signature(token)
            if token.count(".") == 2
            else decrypt_and_verify_token_signature(token)
        )
    except (ValueError, jwcrypto.common.JWException) as e:
        # ValueError is raised when the token doesn't look right (for example no periods)
        # jwcryopto.common.JWException is the base exception raised by jwcrypto,
        # and is raised when the token is malformed or invalid.
        flask.current_app.logger.exception(e)
        raise AuthenticationError(message="Invalid token")

    expiration_time = data.get("exp")
    # we use a hard check on top of the one from the dependency
    # exp shouldn't be before now no matter what
    if expiration_time and (
        dds_web.utils.current_time() <= datetime.datetime.fromtimestamp(expiration_time)
    ):
        return data

    raise AuthenticationError(message="Expired token")
```

worukan: Updated the comments at the beginning and added no token handling.

dds_web/security/auth.py
```diff
@@ -69,21 +69,50 @@ def get_user_roles_common(user):
...
 def verify_token_no_data(token):
-    user, claims = __verify_general_token(token)
+    claims = __verify_general_token(token)
+    user = user_from_subject(claims.get("sub"))
     del claims
     gc.collect()
     return user
 
 
+def verify_invite_token_no_sensdata(token):
```
worukan, line 79: Remove this function!

dds_web/security/auth.py
```diff
@@ -69,21 +69,50 @@ def get_user_roles_common(user):
...
+    email, TKEK = verify_invite_token(token)
+    del TKEK
+    gc.collect()
+    return email
+
+
+def verify_invite_token(token):
```
worukan, line 86:
```
def __base_verify_token_for_invite(token):
    claims = __verify_general_token(token)
    if claims.get("sub"):
        raise AuthenticationError(message="Invalid token")
    return claims


def verify_invite_token(token):
    claims = __base_verify_token_for_invite(token)
    email = claims.get("inv")
    if email:
        return email, models.Invite.query.filter(models.Invite.email == email).first()
    return None, None


def matching_email_with_invite(token, email):
    claims = __base_verify_token_for_invite(token)
    return claims.get("inv") == email
```

worukan: Replace the function with these three.

worukan: An invite token should not have a sub, because it doesn't belong to a user yet. Sensitive content is meant to be for one variable which could be a password, key or other sensitive content. There is additional_claims option for a token. We can come up with our own claims here. If in the future there seems to be a reason to put more than one sensitive content in a token, it can go into additional_claims. I don't consider an email sensitive content, but nevertheless every part of our encrypted token is encrypted so if we put an email in there it will be protected. inv does not seem to be a reserved claim name. We could add the email as inv. So an invite token would look something like (I skipped the exp for the sake of example):
```
encrypted_jwt_token(
                username="",
                sensitive_content="BOGUS",
                additional_claims={"inv": "existing_invite_email@mailtrap.io"},
            )
```

dds_web/web/user.py
```diff
@@ -67,38 +67,35 @@ def index():
 @logging_bind_request
 def confirm_invite(token):
```
worukan, line 68:
```
def confirm_invite(token):
    """Confirm invitation."""
    # Verify token and, on success, get row from invite table
    email, invite_row = dds_web.security.auth.verify_invite_token(token)

    # Check the invite exists
    if not invite_row:
        if email and dds_web.utils.email_in_db(email=email):
            flask.flash("Registration has already been completed.")
            return flask.make_response(flask.render_template("user/userexists.html"))
        else:
            flask.flash("This invitation link has expired or is invalid.", "danger")
            return flask.redirect(flask.url_for("auth_blueprint.index"))

    # Save encrypted token to be reused at registration
    # token is in the session already if the user refreshes the page
    if "invite_token" not in flask.session:
        # New visit or session has expired
        flask.session["invite_token"] = token

    form = forms.RegistrationForm()

    # Prefill fields - unit readonly if filled, otherwise disabled
    # These should only be used for display to user and not when actually registering
    # the user, then the values should be fetched from the database again.
    form.unit_name.render_kw = {"disabled": True}
    if invite_row.unit:  # backref to unit
        form.unit_name.data = invite_row.unit.name
        form.unit_name.render_kw = {"readonly": True}

    form.email.data = email
    suggested_username = email.split("@")[0]

    if dds_web.utils.valid_chars_in_username(
        suggested_username
    ) and not dds_web.utils.username_in_db(suggested_username):
        form.username.data = suggested_username

    return flask.render_template("user/register.html", form=form)
```

worukan: The beginning has changed, added an email check, updated flash text, compiler says use "not in" by following pep guidelines

dds_web/web/user.py
```diff
@@ -122,19 +119,41 @@ def confirm_invite(token):
 )
 def register():
```
worukan, line 120:
```
def register():
    """Handles the creation of a new user"""
    form = dds_web.forms.RegistrationForm()

    # Two reasons are possible for the token to be None.
    # The most likely reason is that the session has expired (given by PERMANENT_SESSION_LIFETIME config variable)
    # Less likely is that the confirm_invite was not called before posting the registration form.
    if flask.session.get("invite_token") is None:
        flask.current_app.logger.info(
            "No token has been found in session when posting to register."
        )
        flask.flash(
            "Error in registration process, please go back and use the link in the invitation email again."
        )
        return flask.redirect(flask.url_for("auth_blueprint.index"))

    # Validate form - validators defined in form class
    if form.validate_on_submit():
        # Create new user row by loading form data into schema
        try:
            new_user = user_schemas.NewUserSchema().load(form.data)
        except Exception as err:
            # This should never happen since the form is validated
            # Any error catched here is likely a bug/issue
            flask.current_app.logger.warning(err)
            flask.flash("Error in registration process, please try again.")
            return flask.redirect(flask.url_for("auth_blueprint.index"))

        flask.flash("Registration successful!")
        return flask.make_response(flask.render_template("user/userexists.html"))

    # Go to registration form
    return flask.render_template("user/register.html", form=form)
```

worukan: We don't need the token here neither disclose the claims and do extra handling of them.

dds_web/api/schemas/user_schemas.py
```diff
@@ -119,6 +120,12 @@ class NewUserSchema(marshmallow.Schema):
         required=True,
         validate=marshmallow.validate.And(marshmallow.validate.Email(), utils.email_not_taken),
     )
+    token_email = marshmallow.fields.Email(
+        required=True,
+    )
+    TKEK = marshmallow.fields.String(
+        required=True,
+    )
```
worukan, line 128: lines 123-128 should be removed.

dds_web/api/schemas/user_schemas.py
```diff
@@ -5,6 +5,7 @@
 ####################################################################################################
 
 # Installed
+import flask
 import marshmallow
 import sqlalchemy
```
worukan, line 10: `import dds_web`

dds_web/api/schemas/user_schemas.py
```diff
@@ -132,10 +139,7 @@ def verify_username(self, value):
 
         if utils.username_in_db(username=value):
             raise marshmallow.ValidationError(
-                message=(
-                    f"The username '{value}' is already taken by another user. "
-                    "Try something else."
-                )
+                message=(f"The username '{value}' is already taken by another user. ")
```
worukan, line 142: compiler says parenthesis are not needed, so
`message=f"The username '{value}' is already taken by another user. "`

dds_web/api/schemas/user_schemas.py
```diff
@@ -151,11 +155,19 @@ def verify_new_email(self, value):
     def verify_and_get_invite(self, data, **kwargs):
```
worukan, line 155:
```
    def verify_and_get_invite(self, data, **kwargs):
        """Verifies that the email is in the invite table and in that case saves the invite info."""

        invite = models.Invite.query.filter(
            models.Invite.email == sqlalchemy.func.binary(data.get("email"))
        ).one_or_none()
        if not invite:
            raise ddserr.InviteError(
                message="No invite has been found for this email at schema validation"
            )

        data["invite"] = invite
```

worukan: Removed token related things. Updated error message.

dds_web/api/schemas/user_schemas.py
```diff
@@ -151,11 +155,19 @@ def verify_new_email(self, value):
...
+            raise ddserr.InviteError(message="Form email and token email not the same")
+
         invite = models.Invite.query.filter(
             models.Invite.email == sqlalchemy.func.binary(data.get("email"))
         ).one_or_none()
         if not invite:
-            raise ddserr.InviteError
+            raise ddserr.InviteError(message="No invite found for this email at schema validation")
 
         data["invite"] = invite
```
worukan, line 173:
```
    def make_user(self, data, **kwargs):
        """Deserialize to an User object"""
        token = flask.session.get("invite_token")
        if not dds_web.security.auth.matching_email_with_invite(token, data.get("email")):
            raise ddserr.InviteError(message="Form email and token email are not the same")

        common_user_fields = {
            "username": data.get("username"),
            "password": data.get("password"),
            "name": data.get("name"),
        }

        # Create new user
        invite = data.get("invite")
        if invite.role == "Researcher":
            new_user = models.ResearchUser(**common_user_fields)
            # Currently no project associations
        elif invite.role in ["Unit Admin", "Unit Personnel"]:
            new_user = models.UnitUser(**common_user_fields)

            new_user.is_admin = invite.role == "Unit Admin"

            invite.unit.users.append(new_user)
        elif invite.role == "Super Admin":
            new_user = models.SuperAdmin(**common_user_fields)

        # Create new email and append to user relationship
        new_email = models.Email(email=data.get("email"), primary=True)
        new_user.emails.append(new_email)
        new_user.active = True

        db.session.add(new_user)

        # TODO Use the TKEK here via the token.
        # Idea is to handle it in auth.py together with project_user_keys.py
        flask.session.pop("invite_token", None)

        # Delete old invite
        db.session.delete(invite)

        # Save and return
        db.session.commit()

        return new_user
```

worukan: Added token handling to the beginning, commented next to the tkek comment and brought token popping to here because we will be done with the invite token here.

dds_web/api/user.py
```diff
@@ -88,12 +90,32 @@ def invite_user(args):
...
+
+        sensitive_content = json.dumps({"invited_email": new_invite.email, "TKEK": TKEK})
+
+        token = encrypted_jwt_token(
+            new_invite.email,
+            sensitive_content,
+            expires_in=datetime.timedelta(
+                hours=flask.current_app.config["INVITATION_EXPIRES_IN_HOURS"]
+            ),
+        )
```
worukan, line 104: Replace lines 96-104 with

```
token = encrypted_jwt_token(
            username="",
            sensitive_content=TKEK,
            expires_in=datetime.timedelta(
                hours=flask.current_app.config["INVITATION_EXPIRES_IN_HOURS"]
            ),
            additional_claims={"inv": new_invite.email},
        )
```

tests/test_user_confirminvite.py
```diff
@@ -4,13 +4,16 @@
 from dds_web import db
 from dds_web.database import models
 import dds_web.errors as ddserr
+from dds_web.security.tokens import encrypted_jwt_token
 import itsdangerous
 import pytest
+import datetime
 
 
 def get_email_token(email):
```
worukan, line 13:
```
def get_email_token(email):
    return encrypted_jwt_token(
        username="",
        sensitive_content="BOGUS",
        expires_in=datetime.timedelta(hours=24),
        additional_claims={"inv": email},
    )
```

tests/test_user_confirminvite.py
```diff
@@ -21,20 +24,27 @@ def test_no_token(client):
...
 def test_expired_token(client):
-    with pytest.raises(itsdangerous.exc.BadTimeSignature):
-        response = client.get(
-            tests.DDSEndpoint.USER_CONFIRM
-            + "ImZpcnN0X3Rlc3RfZW1haWxAbWFpbHRyYXAuaW8i.YW2HiQ.zT4zcM-yt_5S6NfCn2VoYDQSv_g",
-            content_type="application/json",
-        )
+    response = client.get(
+        tests.DDSEndpoint.USER_CONFIRM
+        + "ImZpcnN0X3Rlc3RfZW1haWxAbWFpbHRyYXAuaW8i.YW2HiQ.zT4zcM-yt_5S6NfCn2VoYDQSv_g",
```
worukan, line 41: I guess you might need to rewrite this one.

worukan:
I have tested many of the conditionals. For example, no token in session or emails do not match, etc. I think I didn't change any of the logic/flow that has been designed into the pr. It has been only simplification and clear separation of concerns.

dds_web/api/schemas/user_schemas.py
inaod568: Yep, it's because it's a shorter message.

dds_web/security/auth.py
inaod568: Doesn't really matter since the if statement checks if user is set

alneberg: I do think it matters, I think user is not defined otherwise. I'll indent it at least.

dds_web/api/user.py
```diff
@@ -89,12 +91,33 @@ def invite_user(args):
             raise ddserr.InviteError(message=valerr.messages)
 
         # Create URL safe token for invitation link
-        s = itsdangerous.URLSafeTimedSerializer(flask.current_app.config["SECRET_KEY"])
-        token = s.dumps(new_invite.email, salt="email-confirm")
+        TKEK = "bogus"
+        # TODO change to real TKEK.
+
+        sensitive_content = json.dumps({"invited_email": new_invite.email, "TKEK": TKEK})
```
worukan, line 97: this just slipped in. let's remove it (unused)

dds_web/security/auth.py
```diff
@@ -112,17 +131,20 @@ def __verify_general_token(token):
...
-        if user and user.is_active:
-            return user, data
-
-        return None, data
+        return data
 
     raise AuthenticationError(message="Expired token")
 
 
+def user_from_subject(subject):
```
worukan, line 139: can we change the access modifier of this to private and update the calls (only in this file)?

dds_web/api/user.py
alneberg: Very observant of you! Fixed!

dds_web/security/auth.py
alneberg: Yes! Done!

inaod568 (approved):

worukan (approved):
Good work and collaboration!

PR #833, opened 2022-02-08
Add back accidently removed old solution to check download eligibility

aanil:
Added back old solution.

### DIFF
--- a/dds_web/api/files.py
+++ b/dds_web/api/files.py
@@ -357,6 +357,11 @@ class FileInfo(flask_restful.Resource):
 
         input_ = {**flask.request.args, **{"requested_items": flask.request.json, "url": True}}
 
+        # Check eligibility to download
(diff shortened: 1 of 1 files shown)


inaod568 (approved):
Nice, thanks, works well again!

MatthiasZepper:
Well done. 

I had extended the `verify_project_access` function and modified the `ProjectContentSchema` instead. This now means that file tree listing on the `dds ls` command still works even though the project is not available for download.

Do we want (in particular researchers) to be able list the files inside projects even if they are not available to download to them?

inaod568:
> Well done.
> 
> I had extended the `verify_project_access` function and modified the `ProjectContentSchema` instead. This now means that file tree listing on the `dds ls` command still works even though the project is not available for download.
> 
> Do we want (in particular researchers) to be able list the files inside projects even if they are not available to download to them?

I think we can skip this for now. They have the possibility of asking for access again, so it's definitely not a must have.

PR #838, opened 2022-02-09
Add coverage check to tests

talavis:
* Generate a coverage xml file
* Upload the file to codecov

**Note that this will an "empty" folder called coverage which will be mounted during test runs**


inaod568:
@talavis Cool!!

PR #840, opened 2022-02-09
Web session valid for one hour

inaod568:
Not super important since the web is barely used, but I thought it was nice to add. The time can be discussed later and we can also add that the session is refreshed automatically on redirect. Not applicable now though.

### DIFF
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -6,6 +6,7 @@
 
 # Standard library
 import logging
+import datetime
 
 # Installed
 import click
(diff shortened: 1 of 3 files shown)


talavis (approved):

dds_web/__init__.py
```diff
@@ -6,6 +6,7 @@
 
 # Standard library
 import logging
+import datetime
```
talavis, line 9: minor thing, but maybe sort them alphabetically?

inaod568: Agreed, but this is either a separate task for (much) later, or during other prioritized tasks 👍🏻

PR #847, opened 2022-02-09
Update readme

inaod568:


README.md
```diff
@@ -19,7 +20,9 @@ and/or projects resulting in the production of large amounts of data, e.g. seque
 
 The backend interface is built using [Flask](https://flask.palletsprojects.com/en/2.0.x/).
 
+---
 ## Development
+<br>
```
alneberg, line 25: Not convinced about using `<br>` in markdown is a good idea, but it looks good on the github rendering so ok then.

alneberg:
Looks good! I think the mention of " compiled CSS files ready to go" is outdated. I guess we hope to add it back in the future though.

alneberg (approved):

README.md
inaod568: I thought the page was too crowded and a bit difficult to get an overview of without any spaces, but I don't know any other way to do spaces in markdown.

PR #849, opened 2022-02-10
Session instead of cookie

alneberg:
1. Use session cookie instead of regular cookie for 2fa initiated token
2. Check that token exists already at the GET request of `confirm_2fa`
3. Added check for case when session token is valdi but user does not exist (happens during testing if user is deleted in-between login and confirm_2fa).

### DIFF
--- a/dds_web/web/user.py
+++ b/dds_web/web/user.py
@@ -201,7 +201,7 @@ def confirm_2fa():
         )
         return flask.redirect(flask.url_for("auth_blueprint.login", next=next))
 
-    # Valid invite token, but user does not exist (should never happen)
+    # Valid 2fa initiated token, but user does not exist (should never happen)
     if not user:
         flask.session.pop("2fa_initiated_token", None)
         flask.flash("Error: Internal error.", "danger")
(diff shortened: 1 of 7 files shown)


dds_web/web/user.py
```diff
@@ -215,11 +224,10 @@ def confirm_2fa():
...
         # Correct username, password and hotp code --> log user in
         flask_login.login_user(user)
-        # Remove token from cookies and make it expired
         flask.flash("Logged in successfully.")
-        redirect_to_index = flask.redirect(next or flask.url_for("auth_blueprint.index"))
-        redirect_to_index.set_cookie("2fa_initiated_token", "", expires=0)
-        return redirect_to_index
+        # Remove token from session
+        flask.session.pop("2fa_initiated_token", None)
+        return flask.redirect(next or flask.url_for("auth_blueprint.index"))
```
inaod568, line 230: Check `is_safe_url` in `utils.py` too

inaod568: Aah ok! Missed that, great, thanks!

inaod568 (approved):

PR #853, opened 2022-02-10
Delete user keys too when Deleting/Archiving project

aanil:
* Added it to tests 
* Added stuff to try to increase coverage


dds_web/api/project.py
```diff
@@ -329,6 +329,9 @@ def delete_project_contents(project):
         """Remove project contents"""
         DBConnector(project=project).delete_all()
 
+        for project_key in project.project_user_keys:
+            db.session.delete(project_key)
+
```
inaod568, line 334: Maybe we should have this after the data is removed?

inaod568 (approved):

PR #854, opened 2022-02-10
Removing a user from a project removes the associated project keys with the user

worukan:
Written a test to check it, Fixed the code accordingly, Updated the conftest due to running issues with the project setup


worukan:
Reused test written by @aanil

inaod568 (approved):

Issue #825, opened 2022-02-07
Allow invite of `Project Owner`


alneberg:
At the moment, such an invite will fail since that role is not checked for in the NewUser schema in the API. 

1. There’s no check for the Project Owner  “role” in the InviteUserSchema  (in dev at least) so specifying a Project OWner  in dds user add [email] --role  will fail. 
2. When adding that functionality, we also need to add a check that there’s a project specified in the invite, otherwise a “project owner ” doesn’t make sense.
3. In the `NewUserSchema` there is no check for the role `Project Owner`:
```
        # Create new user
        invite = data.get("invite")
        if invite.role == "Researcher":
            new_user = models.ResearchUser(**common_user_fields)
            # Currently no project associations
        elif invite.role in ["Unit Admin", "Unit Personnel"]:
            new_user = models.UnitUser(**common_user_fields)

            new_user.is_admin = invite.role == "Unit Admin"

            invite.unit.users.append(new_user)
        elif invite.role == "Super Admin":
            new_user = models.SuperAdmin(**common_user_fields)
```
--> even though the role is saved in the invite, when the role is retrieved during registration --> the registration fails

Issue #798, opened 2022-02-03
Add user to project -- fix workflow


inaod568:
At the moment, if `dds user add --project` is called and the account does not exist, there's an invite sent out, but the invite has no information about any project and therefore the user is only invited to the system in general. We need to add information in the invite procedure which keeps track of that the user is getting invited and should be getting access to one or more projects.
It's possible that we can add a column to the `Invite` table, and in that case we need to be able to specify multiple projects. Look into what the best solution is.

inaod568:
I'm closing this issue due to some confusion and opening a new one which is more relevant at this stage.

PR #855, opened 2022-02-10
key generation for invite

worukan:
This work has started. The goal is to establish key pairs for an invite, handle the temporary key and be able to run a flow that starts with creating an invite and reaching to a user.

- Invite has to be setup with keys properly
- invite token has to be received properly and unwrapped to verify the temporary key in it
- create a user with the key setup transferred from the invite

Heads up @inaod568 @alneberg @MatthiasZepper 
If anyone has already been working or thinking on overlapping parts, let's synchronize!


MatthiasZepper (approved):
Very nicely solved! I like it. 

It is still mind-boggling to me that one can implicitly change values inside a function that are not returned (like assigning `owner.private_key = encrypted key` in  `__encrypt_owner_private_key` and this still persists outside the scope of the function although the modified owner object is not returned, but you seem to know what you are doing...

dds_web/database/models.py
```diff
@@ -68,6 +68,38 @@ class ProjectUserKeys(db.Model):
...
+    # ---
+    invite_id = db.Column(
+        db.Integer, db.ForeignKey("invites.id", ondelete="CASCADE"), primary_key=True
+    )
+    invite = db.relationship("Invite", back_populates="project_invite_keys")
+    # ---
+
+    # Additional columns
+    key = db.Column(db.LargeBinary(300), nullable=False, unique=True)
+
```
MatthiasZepper, line 101: I think the `ProjectInviteKeys` need an additional owner column analogous to the `ProjectUser] table. I would actually suggest to duplicate this information also in the `ProjectUserKeys`, because then we could easily check if a Researcher is allowed to share (Project Owner) the project key or not.

worukan: Thank you for bringing up an important detail!
I double checked a few places. Here are some things to raise:

- According to the role responsibilities agreement that we have established, a research account cannot invite someone to a project. The minimum required level of role for inviting to a project is project owner.
- The implementation to invite to a project does not take the previous point into account. I wrote a new test case for it and it fails, meaning that it is now possible that a researcher can invite someone to a project. I think we need to create a bug for this. I can immediately fix it in that, if agreed. @inaod568 @MatthiasZepper @alneberg 
- Considering that we can fix this in the invite function, we do not need to keep track of who has invited who under normal operation. However I highly recommend that we log who has invited who to which project at which date and time as part of our audit log. This can be investigated from the log if needed.
- So all in all, I don't think we need to add a new column to the model.

MatthiasZepper: >     * According to the role responsibilities agreement that we have established, a research account cannot invite someone to a project. The minimum required level of role for inviting to a project is project owner.

That is correct. A project owner status is required to grant further users access to a project.

I was a bit sloppy in my previous comment, because I just wrote _Researcher_ (the return value of `User.role`), even though I meant `ResearchUser` with regard to the database models.

The logical problem is, that the role of a Unit Admin is a property of the user, such that we can return either a _Unit Personnel_ or _Unit Admin_ value for `User.role` on `UnitUsers`(database model), but for the researchers it is more complicated as they can be the owners of just some projects they have access to.

>     * The implementation to invite to a project does not take the previous point into account. I wrote a new test case for it and it fails, meaning that it is now possible that a researcher can invite someone to a project. I think we need to create a bug for this. I can immediately fix it in that, if agreed. @inaod568 @MatthiasZepper @alneberg

If you wish to fix that, I am happy to leave it to you. 

In a way I was aware of this and wanted to sort that out, but by means of by adding that owner column to the keys. I have no real understanding how to straightforwardly solve it otherwise.  

>     * Considering that we can fix this in the invite function, we do not need to keep track of who has invited who under normal operation. However I highly recommend that we log who has invited who to which project at which date and time as part of our audit log. This can be investigated from the log if needed.

Agree. With the structlog package in place, this should be fairly straightforward. See example for the deletion here:

https://github.com/ScilifelabDataCentre/dds_web/blob/cfa6bfc001205ee645dff3cc1361edc99f17cafc/dds_web/api/user.py#L519-L523

>     * So all in all, I don't think we need to add a new column to the model.

Not convinced yet, but I also don't insist on it.

worukan: As far as I can follow everyone's point:
`InviteUserSchema` in `invite_user` loads the **role** argument when someone tries to invite someone. This is persisted into the db as `invite.role`, so if a project owner is invited `invite.role = "Project Owner"`. Then we generate key pair for invite and invite.id is associated with the generated keys in `ProjectInviteKeys`. Invited user goes to confirm_invite with the token, we make sure that the invite is in the db and fetch it. At this point you have the invite with its role already set and there is an association to the generated keys for itself, so as I see it the role is immediately accessible.
Please help me understand if I'm missing someone's point. We could perhaps do a spontaneous chat about it, if needed?

inaod568: Aahh! I follow your meaning now. Yes, this should be already handled then! @alneberg @MatthiasZepper

alneberg: We want to be able to invite a user to multiple projects at once where the user is an owner of some but not an owner of others. I think we have called "Project Owner" a user role a bit too sloppily. It's not a user role, it's a user-project specific permission. But anyway. Inviting to a project is not even implemented yet so this can all wait until later. I think it's important to get these changes merged and we can build from here instead of creating huge PRs.

worukan: Yes we talked about inviting to multiple projects this week. It is a separate issue to solve. And yes again that right now we don't pass in a project arg to the invite so it is not helpful to set role to project owner while inviting. I think we need to create specific issues for these and write clear a description for each.

inaod568: https://github.com/scilifelabdatacentre/dds_web/issues/860

inaod568: Feel free to add more info if you want to.

MatthiasZepper: I fully agree with being everything said. 

The case that _an invited e-mail needs to have owner privileges for some, but not all projects_  was indeed the biggest issue I had during my work. Even worse, Phil had specifically requested that one needs to be able to specify multiple projects in one invocation of the `dds user add` command in our BioInfo meeting on Monday.

Since #798 was basically a "Fix everything" issue, I felt responsible for all of it. Solving the problem of _Fine-grained project ownership to an invitee_ was basically what I felt would be the purpose of #825?

Anyway: I am more than happy to see this PR merged and then build on top of it whatever still remains to be done! Because the problem of being confronted with an overarching task was exactly what caused me a lot of headache in the past days.

worukan:
All tests pass locally. There is a problem building the test environment in here at the time of commenting.

inaod568:
> All tests pass locally. There is a problem building the test environment in here at the time of commenting.

@worukan I'm guessing it's the same issue as Matthias had. Linus suggested to add the `-y` to the dockerfile, did that remotely now so hopefully it sorts it out now.

inaod568:
I'd add a test case regarding my previous comment.

dds_web/database/models.py
inaod568: @MatthiasZepper I agree that the full task of inviting a user to a project, also being able to invite a user to multiple projects (has been on the table for a while), also adding existing users to projects and having different procedures depending on which role the invited user is and the current user has, and to add the project keys to the invite -- that's a big task. However, the task that you were assigned to/agreed to do (#798) has a description that I don't see as a "fix everything" and I also have tried to explain what I mean during at least two of our meetings last week. In this case it turned out that it was better to do certain other tasks before approaching the task you started with, which sometimes is what happens. If you feel that you have a tough task ahead of you and are unsure about what the goal is, please bring this up and discuss this as soon as possible so that we can sort out any confusion (maybe theres a misinterpretation on my part, or yours) and possibly even change which task you work on. 👍🏻

worukan:
Thank you. I try to write as much understandable code as possible, but please take note of anything unclear or needs clarification/description in a new issue. I can come back to those after we deliver for testing.

dds_web/api/user.py
```diff
@@ -118,7 +118,11 @@ def invite_user(args):
 
         # Append invite to unit if applicable
         if new_invite.role in ["Unit Admin", "Unit Personnel"]:
-            auth.current_user().unit.invites.append(new_invite)
+            if "Unit" in auth.current_user().role:
+                # Give new unit user access to all projects of the unit
+                auth.current_user().unit.invites.append(new_invite)
+                for project in auth.current_user().unit.projects:
```
MatthiasZepper, line 124: Doesn't looping though `auth.current_user().unit.projects` implicitly assume that every unit user holds the keys for every project ever created by a particular unit? 

Judged by the code, currently this procedure doesn't seem to be tolerant on missing keys (at least the `KeyNotFoundError` seems to be raised). What happens to the keys when a project is deleted or archived - should they persist nonetheless?

aanil: Maybe the unit user can be given access to just active projects? So none that are 'Archived' or 'Deleted'?

inaod568: @MatthiasZepper @aanil The keys should be removed when the project is deleted or archived. So there won't be any keys for those projects.

inaod568: @aanil implemented the deletion of the keys last week https://github.com/ScilifelabDataCentre/dds_web/blob/ad24b040bab6450ab936b5a713d01fc38192e43c/dds_web/api/project.py#L339-L341

dds_web/security/project_user_keys.py
```diff
@@ -39,16 +40,16 @@ def __decrypt_with_rsa(ciphertext, private_key):
...
-def __encrypt_project_private_key_for_user(user, project_private_key):
-    user_public_key = serialization.load_der_public_key(user.public_key)
-    if isinstance(user_public_key, asymmetric.rsa.RSAPublicKey):
-        return __encrypt_with_rsa(project_private_key, user_public_key)
-    exception = Exception("User public key cannot be loaded!")
+def __encrypt_project_private_key(owner, project_private_key):
+    public_key = serialization.load_der_public_key(owner.public_key)
+    if isinstance(public_key, asymmetric.rsa.RSAPublicKey):
+        return __encrypt_with_rsa(project_private_key, public_key)
+    exception = Exception("Public key cannot be loaded for encrypting the project private key!")
```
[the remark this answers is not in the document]
inaod568: Agreed. Added a TODO comment.

dds_web/api/user.py
inaod568: @MatthiasZepper @aanil https://github.com/scilifelabdatacentre/dds_web/issues/874 
Not a must have though

inaod568: Added a check to see if the project is active.

alneberg (approved):

dds_web/database/models.py
MatthiasZepper: > Just a side note, @MatthiasZepper a single invocation of `dds user add` can still entitle several requests to the API so it doesn't have to mean that the endpoint needs to be able to accept multiple projects at once.

True. That is actually the much easier way to go and solve that in the CLI. I can do that.

PR #859, opened 2022-02-11
Add more tests for user

aanil:
* Add some more tests for user.py

### DIFF
--- a/tests/test_user_add.py
+++ b/tests/test_user_add.py
@@ -1,3 +1,4 @@
+import http
 import json
 import sqlalchemy
 from dds_web import db
@@ -41,7 +42,7 @@ def test_add_user_with_researcher(client):
         data=json.dumps(first_new_user),
         content_type="application/json",
     )
-    assert response.status == "403 FORBIDDEN"
+    response.status_code == http.HTTPStatus.FORBIDDEN
     invited_user = models.Invite.query.filter_by(email=first_new_user["email"]).one_or_none()
     assert invited_user is None
 
@@ -53,7 +54,7 @@ def test_add_user_with_unituser_no_role(client):
         data=json.dumps(first_new_email),
         content_type="application/json",
     )
-    assert response.status == "400 BAD REQUEST"
+    assert response.status_code == http.HTTPStatus.BAD_REQUEST
     invited_user = models.Invite.query.filter_by(email=first_new_email["email"]).one_or_none()
     assert invited_user is None
 
@@ -65,7 +66,7 @@ def test_add_user_with_unitadmin_with_extraargs(client):
         data=json.dumps(first_new_user_extra_args),
         content_type="application/json",
     )
-    assert response.status == "400 BAD REQUEST"
+    assert response.status_code == http.HTTPStatus.BAD_REQUEST
     invited_user = models.Invite.query.filter_by(
         email=first_new_user_extra_args["email"]
     ).one_or_none()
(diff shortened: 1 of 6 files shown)


tests/test_user_add.py
```diff
@@ -41,7 +43,7 @@ def test_add_user_with_researcher(client):
         data=json.dumps(first_new_user),
         content_type="application/json",
     )
-    assert response.status == "403 FORBIDDEN"
+    response.status_code == http.HTTPStatus.FORBIDDEN
```
alneberg, line 46: I think it's nice to keep the `assert` before the comparison?

tests/test_user_add.py
```diff
@@ -137,7 +169,7 @@ def test_add_user_with_unitpersonnel_permission_denied(client):
         data=json.dumps(new_unit_admin),
         content_type="application/json",
     )
-    assert response.status == "403 FORBIDDEN"
+    response.status_code == http.HTTPStatus.FORBIDDEN
```
alneberg, line 172: same here, keep the `assert`

tests/test_user_add.py
```diff
@@ -241,7 +318,7 @@ def test_existing_user_change_ownership_same_permissions(client):
         data=json.dumps(user_same_ownership),
         content_type="application/json",
     )
-    assert response.status == "403 FORBIDDEN"
+    response.status_code == http.HTTPStatus.FORBIDDEN
```
alneberg, line 321: here as well

tests/test_user_add.py
```diff
@@ -255,4 +332,4 @@ def test_add_existing_user_with_unsuitable_role(client):
         data=json.dumps(user_with_unsuitable_role),
         content_type="application/json",
     )
-    assert response.status == "403 FORBIDDEN"
+    response.status_code == http.HTTPStatus.FORBIDDEN
```
alneberg, line 335: And here

tests/test_user_info.py
```diff
@@ -79,3 +79,16 @@ def test_get_info_superadmin_user(client):
...
+    """Show the amount of GB hours and the total cost for a unit"""
+    # Preliminary test - to be expanded later
+    response = client.get(
+        tests.DDSEndpoint.USAGE,
+        headers=tests.UserAuth(tests.USER_CREDENTIALS["unituser"]).token(client),
+        content_type="application/json",
+    )
+    assert response.status_code == http.HTTPStatus.OK
+    assert "project_usage" in response.json.keys()
+    assert "total_usage" in response.json.keys()
```
alneberg, line 94: I know it's daunting, but we should perhaps test whether the data returned is valid? This is very nice for coverage though... 😂

tests/test_user_add.py
aanil: Done!

tests/test_user_add.py
aanil: Done!

tests/test_user_info.py
aanil: Yeah, but the returned data is all 0. I'm not sure if it should be mocked.

alneberg: Ah, true. Might need to add some simple sizes to the files in the db then?

tests/test_user_add.py
aanil: 👍

tests/test_user_add.py
aanil: 👍

tests/test_user_info.py
aanil: The issue is that the upload time and the time at which the test is run is so close together that the function rounds it to zero.

inaod568 (approved):

PR #861, opened 2022-02-11
Moved functions from db connector

inaod568:
The majority of all functions in the `db_connector` were used by a single other function, so I moved them and renamed the `db_connector.py` to `db_tools.py`. If we want we can of course move them back to this file, just not have them as static functions in that class, but in the way it is now we're following the same pattern as for for example inviting users etc.


dds_web/api/files.py
```diff
@@ -276,18 +350,83 @@ def delete(self):
...
+
+        not_removed_dict, not_exist_list = ({}, [])
+
+        with ApiS3Connector(project=project) as s3conn:
+
+            # Delete each file
+            for x in files:
+                # Delete from db
+                try:
+                    name_in_bucket = self.delete_one(project=project, filename=x)
```
aanil, line 371: It should be `filename=files[x]` here as files is a dict and not a list.

dds_web/api/files.py
```diff
@@ -276,18 +350,83 @@ def delete(self):
...
+                    if not name_in_bucket:
+                        raise DatabaseError(
+                            message="Remote file name not found.", pass_message=True
+                        )
+                except FileNotFoundError:
+                    db.session.rollback()
+                    not_exist_list.append(x)
```
aanil, line 378: Same here.

dds_web/api/files.py
inaod568: No `files` is a list. 
`flask.request.json` doesn't need to be a dict. In this case it's a list of files being passed in. 
```
flask.current_app.logger.debug(f"Files: {files}")
```
gives the following output
```
dds_backend   | [2022-02-13 18:34:14,535] files [DEBUG] Files: ['testfile.txt']
```

dds_web/api/files.py
inaod568: https://github.com/ScilifelabDataCentre/dds_web/pull/861#discussion_r805398118

dds_web/api/files.py
aanil: 👍

dds_web/api/db_tools.py
```diff
@@ -0,0 +1,41 @@
+"""API DB Connector module."""
```
alneberg, line 1: I guess this is left from the old file.

dds_web/api/db_tools.py
```diff
@@ -0,0 +1,41 @@
...
+# Own modules
+from dds_web.database import models
+from dds_web import db
+from dds_web.errors import (
+    DatabaseError,
+    UserDeletionError,
+)
+
+####################################################################################################
+# CLASSES ################################################################################ CLASSES #
```
alneberg, line 21: Functions?

dds_web/api/files.py
```diff
@@ -220,49 +220,123 @@ def get(self):
...
         # Check project not empty
-        with DBConnector(project=project) as dbconn:
-            # Get number of files in project and return if empty
-            num_files = project.num_files
-            if num_files == 0:
-                return {
-                    "num_items": num_files,
-                    "message": f"The project {project.public_id} is empty.",
+        if project.num_files == 0:
+            raise EmptyProjectException(project=project)
```
alneberg, line 224: I don't think listing files on an empty project should raise BAD_REQUEST. I think of it as doing `ls` on an empty directory. I think the previous behaviour returning info that is empty is better.

MatthiasZepper: Agree. In particular, since Researchers can list on projects with the _In Progress_ status.

aanil: Should Researchers be able to list project contents of projects with `In Progress` status? It means the project is not ready to be accessed by the Researcher after all.

MatthiasZepper: My gut feeling was _No_ and this is what I also was about to point out during the review of your PR [when Ina already merged it](https://github.com/ScilifelabDataCentre/dds_web/pull/833#issuecomment-1032651676). But I acknowledge that it isn't a big deal.

inaod568: True. Added a check.

inaod568: @MatthiasZepper It's in this PR now

MatthiasZepper: @inaod568: Thank you very much! Great!

alneberg (approved):
Great! I think this is an improvement and I haven't been able to break it yet. :)

aanil (approved):
👍

PR #863, opened 2022-02-11
a researcher (non-project owner) should not be able to invite someone…

### LINKED ISSUE #862: a researcher (non-project owner) should not be able to invite someone to a project
Noticed it thanks to @MatthiasZepper 
written a test case (not committed anywhere at the time of writing this) which fails now. Needs its small fix.

1. Check if current user is a Researcher
2. If so, check that a project is specified
3. Get row from ProjectUsers
4. Check if the row has `owner=True`
5. If so -- user can invite another Researcher to the specified project.

worukan:
… to a project, written a test case to cover this
Closes #862

### DIFF
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -5,11 +5,9 @@
 ####################################################################################################
 
 # Standard library
-import datetime
 import os
 import smtplib
 import time
-import json
 import datetime
 
 # Installed
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -43,7 +43,7 @@ action_logger = structlog.getLogger("actions")
 # ENDPOINTS ############################################################################ ENDPOINTS #
 ####################################################################################################
 class AddUser(flask_restful.Resource):
-    @auth.login_required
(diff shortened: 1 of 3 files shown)


dds_web/api/user.py
```diff
@@ -43,7 +43,7 @@
 # ENDPOINTS ############################################################################ ENDPOINTS #
 ####################################################################################################
 class AddUser(flask_restful.Resource):
-    @auth.login_required
+    @auth.login_required(role=["Super Admin", "Unit Admin", "Unit Personnel", "Project Owner"])
```
alneberg, line 46: Frankly I think this is incorrect. I don't think "Project Owner" is a valid role as I've explained already.

worukan: I have followed the agreed roles which are pinned down. We could remove that as well, if @inaod568 is fine with it. Someone can then update the agreement as well.

alneberg (approved):
Nice! Yes, including project owner as a role like this works. We need to keep this in mind for all endpoints. A quick search in the code base showed no issues with the endpoints we have currently.

inaod568 (approved):
Nice, works well!

Issue #877, opened 2022-02-14
Create a new endpoint for providing project access to a user


worukan:
When a user needs to reset password, project user keys have to be re-established. Roles from Project Owner and above can invite to project which means that they can `share_project_private_key`. The flow for this activity has to be defined clearly. When can this endpoint be used? How is the interaction initiated? How would for example a unit personnel know to use this and when? Do we need to store the information about a reset password and/or how and who to propagate this information? A researcher can be involved in multiple projects requiring interaction with multiple units/project owners. Do we need to send email about the user to all involved?
Thinking of _if it is a unit user, it would be only unit projects, and if it is a research user, it would be the projectusers entries_ , we could check which projects are defined in the db but not having projectuserkeys entries in the table, if we immediately clean up that table at password reset. Does this make sense?
This is related to #876 in a way that first reset password happens and then someone has to use this endpoint to share project keys to the user.
In terms of security parts, we already have everything ready, I think.

MatthiasZepper:
See more extensive suggestion written in #876. 

To initiate the endpoint, a sufficiently privileged user would enter `dds project access fix` or alike and project access would be reestablished for the specified e-mails and/or projects.

I would not delete the keys right away, but move them to a _Lost/Orphaned Keys_ table. Then the `fix` command could replace those keys after which they are deleted for good.

inaod568:
Merged with #885

Issue #876, opened 2022-02-14
Reset password flow to work with key derivation


worukan:
There will be a key that will be derived from a user's password. The derived key will protect the user's private key. This means that if a password is lost, all the keys and related components have to be re-generated. This is a similar situation with invite user. I suggest we prepare a flow for reset password similar to the invite-register user flow. We could again come up with a new encrypted token such as encrypted_token(user, temp_key_for_reset, expiry_time_for_reset_token, a reset token claim such as "rst": "pwd" or something better). We need to go through all the steps needed/expected from reset password request to setting the new password. I think this task can be done by clearly outlining the flow, what is present and what is missing in here textually and/or write code which will at first work without security parts or at least with the existing ones.

MatthiasZepper:
I think, the flows could look like that:

**Self-initiated/Anonymous reset: Only for Researchers**
1.  In the CLI:  `dds user password-reset -e primary-email@user.se`  (checks that e mail is primary e-mail linked to a ResearchUser. Request to be denied for UnitUsers)
2. Require MFA of the user to be reset for confirmation. (We can ask for some additional information like the public-project-id of the most recent project the user has had access to)
3. Send out confirmation link via e-mail
4. Confirmation link opens a form in the browser to choose a new password.
5. New replacement key pair for the user is created and the Users Private Key is encrypted with the new password-derived key. 
6. (Old) encrypted project private keys for this user are moved to another table "Orphaned/lost keys". 
7. For each active project of that user, an e-mail is sent out to the respective Project Owners. It contains a Copy and Paste ready command `dds project access fix -e primary-email@user.se -p ngi666` to reestablish access. 
7b. _In case the respective user is the sole ProjectOwner, that e-mail should be instead sent to the unit user who last modified the status of that project._
8. Running the `dds project access fix` command by the ProjectOwners respectively Unituser should basically comb through the Orphaned/lost keys table for projects to which the these have access but the specified user not anymore. Then the access is granted anew by encrypting the Projects' Private Keys with the new Public Key of that User and subsequently all the recovered LostKeys are removed from that table.  

**Unituser reset: Requires a logged in user from the same unit (asked a colleague to help out)**
1.  In the CLI:  `dds user password-reset -e primary-email@user.se`  (checks that e mail is primary e-mail linked to a UnitUser of the same Unit)
2. Require MFA of the user triggering the reset and MFA of the user whose reset is triggered simultaneously.
[...]

PR #885, opened 2022-02-15
ProjectAccess endpoint

inaod568:
Added a skeleton of a ProjectAccess endpoint. Created a decorator to make sure that the user is specified -- this can be used in other places later too. Added comment regarding what should be added. 

This is a part of #877 and implements a small part of what is described in [this comment](https://github.com/ScilifelabDataCentre/dds_web/issues/876#issuecomment-1040086353) in #876


dds_web/api/project.py
```diff
@@ -473,3 +479,43 @@ def get(self):
...
+
+class ProjectAccess(flask_restful.Resource):
+    """Renew project access for users."""
+
+    @auth.login_required(role=["Unit Admin", "Unit Personnel", "Project Owner"])
+    @logging_bind_request
+    @dbsession
+    @user_required
+    @project_optional
+    @renew_access_required
```
alneberg, line 492: I must say I'm not a fan of using decorators this way. I think it makes following the code logic very difficult. But that's entirely up to you.

inaod568: What are you suggesting that I change? We need the functionality for multiple endpoints, but we could change them to function calls if you prefer that

dds_web/api/dds_decorators.py
```diff
@@ -31,6 +40,99 @@
...
+        current_user_role = get_user_roles_common(user=auth.current_user())
+        other_user_role = get_user_roles_common(user=user)
+
+        # Check if Researcher and if so is project owner or not
+        if other_user_role == "Researcher" and project:
+            project_user_row = models.ProjectUsers.query.filter_by(
+                project_id=project.id, user_id=user.username
+            ).one_or_none()
+            if project_user_row and project_user_row.owner:
+                other_user_role = "Project Owner"
```
alneberg, line 80: I think this is taken care of by the `get_user_roles_common` no? Since `project` should still be in `flask.request.arg`?

dds_web/api/project.py
```diff
@@ -482,3 +487,100 @@ def get(self):
...
+            raise AccessDeniedError(message="You cannot renew your own access.")
+
+        # Get roles
+        current_user_role = get_user_roles_common(user=auth.current_user())
+        other_user_role = get_user_roles_common(user=user)
+
+        # Check access
+        if not (
+            (
+                current_user_role in "Unit Admin"
```
alneberg, line 551: I think this needs to be an `==` instead of `in`?

alneberg (approved):
Just a tiny bug and resolving conflicts left. 👍

PR #886, opened 2022-02-15
Reactivate project access

inaod568:
Reactivate project access if user has lost access during reactivation.


alneberg:
I think it would be better to do the reactivate project access as a separate endpoint rather than mixing it with the activate deactivate user. They're not completely the same. For example, reactivate project access should be possible to be done by any unit user or a project owner for the project in question, while reactivating a user should be only for admins. But the code looks good in general!

dds_web/api/user.py
```diff
@@ -452,8 +447,29 @@ def post(self):
...
+
+            for proj in list_of_projects:
+                if proj.is_active:
+                    project_keys_row = models.ProjectUserKeys.query.filter_by(
+                        project_id=proj.id, user_id=user.username
+                    ).one_or_none()
+                    if not project_keys_row:
+                        share_project_private_key(
+                            from_user=current_user, to_another=user, project=proj
+                        )
```
alneberg, line 468: I believe you can copy paste this into the #885 no?

alneberg:
Ok, I see as an addition to also having it in #885 it's perfectly fine to also check for project reactivation here as well.

dds_web/api/user.py
inaod568: @alneberg It's very possible that we can use this or extremely similar code in #885, but I decided to just make the skeleton in that first of all. The person who works on the `ProjectAccess` endpoint could potentially create a new function in that class so that we can just call it from this function too.

inaod568: I can add a comment in the next card for this so that noone re-invents the wheel

inaod568:
Waiting for #885

inaod568:
#885 is done so this is ready

alneberg (approved):
Works well, even though we don't yet have a use-case for this.

Issue #860, opened 2022-02-11
Invite user to multiple projects at the same time


inaod568:
At this time, https://github.com/ScilifelabDataCentre/dds_web/pull/855 and https://github.com/ScilifelabDataCentre/dds_web/pull/863 add to and change the functionality of the invites and how inviting new users  / adding existing users to projects should work. With these changes, a user can invite a user to one project at a time.

**Needed functionality: ** To invite a _new user_ to multiple projects at one time. 

At some point we should also be able to invite a user to multiple projects with different roles, e.g. a user could be a Researcher for one and a Project Owner for another. Adding this functionality is not a high priority but a should have. 

However, adding users to multiple projects at a time is **required**.

inaod568:
@MatthiasZepper Is this the card you're working on at the moment? Or something else?

alneberg:
Fixed with #911.

PR #888, opened 2022-02-15
Issue860 invites multiple projects

### LINKED ISSUE #887: Add functionality to add project to _invite_ of new user
This should entail (on the invite side, not registration side)

* Adding project as argument (if it's specified) here: https://github.com/ScilifelabDataCentre/dds_web/blob/b573bec26d44a3fd159950d638025e2b1d453086/dds_web/api/user.py#L62
* Checking if project if specified and in that case sharing the project private here: https://github.com/ScilifelabDataCentre/dds_web/blob/b573bec26d44a3fd159950d638025e2b1d453086/dds_web/api/user.py#L127 

Add comments below if you feel you need to add anything, or if you feel anything is incorrect about that I wrote above.

MatthiasZepper:
This PR addresses one part of #860 and closes #887: Attaching project associations with new or pending invites. 

Since we intended to be as atomic as possible with the PRs, the scope of this one is only to create the respective records in the database. Moving Keys and Projects from the Invite tables to the analogous User tables will be addressed in another PR.

Another ToDo for the CLI is actually to loop through multiple projects and users.

### DIFF
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -51,7 +51,7 @@ class AddUser(flask_restful.Resource):
         project = flask.request.args.get("project", None)
         if project:
             project = project_schemas.ProjectRequiredSchema().load({"project": project})
-            role = args.get("role", None)
+            role = flask.request.json.get("role", None)
 
         args = flask.request.json
         # Check if email is registered to a user
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -45,33 +45,42 @@ class AddUser(flask_restful.Resource):
     @auth.login_required(role=["Super Admin", "Unit Admin", "Unit Personnel", "Project Owner"])
     @logging_bind_request
     def post(self):
-        """Create an invite and send email."""
+        """Associate existing users or unanswered invites with projects or create invites"""
 
+        # A project may or may not be specified
         project = flask.request.args.get("project", None)
+        if project:
+            project = project_schemas.ProjectRequiredSchema().load({"project": project})
+            role = args.get("role", None)
+
         args = flask.request.json
         # Check if email is registered to a user
         existing_user = user_schemas.UserSchema().load(args)
+        unanswered_invite = user_schemas.UnansweredInvite().load(args)
 
-        if existing_user and not project:
-            raise ddserr.DDSArgumentError(
-                message="User exists! Specify a project if you want to add this user to a project."
-            )
-
-        if not existing_user:
-            # Send invite if the user doesn't exist
-            invite_user_result = self.invite_user(args)
-            return invite_user_result, invite_user_result["status"]
-
-        else:
-            # If there is an existing user, add them to project.
-            if project:
-                add_user_result = self.add_user_to_project(existing_user, project, args.get("role"))
+        if existing_user or unanswered_invite:
+            if project and role:
+                whom = existing_user or unanswered_invite
+                add_user_result = self.add_to_project(whom, project, role)
                 flask.current_app.logger.debug(f"Add user result?: {add_user_result}")
                 return add_user_result, add_user_result["status"]
+            else:
+                raise ddserr.DDSArgumentError(
+                    message="This user was already added to the system. Specify the project you wish to give access to."
+                )
+
+        else:
+            # Send invite if the user doesn't exist
+            if project:
+                invite_user_result = self.invite_user(args, project)
+            else:
+                invite_user_result = self.invite_user(args)
+
+            return invite_user_result, invite_user_result["status"]
 
     @staticmethod
     @logging_bind_request
-    def invite_user(args):
+    def invite_user(args, project=None):
         """Invite a new user"""
 
         try:
@@ -125,6 +134,15 @@ class AddUser(flask_restful.Resource):
                         share_project_private_key(auth.current_user(), new_invite, project)
         else:
             db.session.add(new_invite)
+            if project:
+                project.invites.append(
+                    models.ProjectInvites(
+                        project_id=project.id,
+                        invite_id=new_invite.id,
+                        owner=new_invite.role == "Project Owner",
+                    )
+                )
+                share_project_private_key(auth.current_user(), new_invite, project)
 
         db.session.commit()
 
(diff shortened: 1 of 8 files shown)


dds_web/database/models.py
```diff
@@ -100,6 +100,39 @@ class ProjectInviteKeys(db.Model):
     key = db.Column(db.LargeBinary(300), nullable=False, unique=True)
 
 
+class ProjectInvites(db.Model):
```
alneberg, line 103: I think we could do it without this table, and instead just add "owner" to the ProjectInviteKeys instead or is there some issue with primary keys there?

dds_web/api/user.py
```diff
@@ -165,30 +183,44 @@ def add_user_to_project(existing_user, project, role):
...
-        for rusers in project.researchusers:
-            if rusers.researchuser is existing_user:
+
+        if isinstance(whom, models.ResearchUser):
+            link = project.researchusers
+        else:
+            link = project.invites
+
+        for rusers in link:
+            if rusers.researchuser is whom:
```
alneberg, line 194: I think it would be safer to use `==` instead of `is` here. `is` checks that the two variables actually reference the exact same object, so I'm not sure if two different instances of the same user would be evaluating to true here.

dds_web/database/models.py
alneberg: I can see how I caused this confusion. Sorry for that. I think it should be fairly quick to delete it though?

inaod568: Yeah I agree, I think it's better to add this to the `ProjectInviteKeys` table.

inaod568: Otherwise we have `Invite`, `ProjectInviteKeys` and `ProjectInvites`

inaod568: It does mean that we would have the `owner` column for all unit users too though, but since we wouldn't need to check that in that case it doesn't actually matter

MatthiasZepper: Sorry, but this is a clear **NO** from my side!. It was _my_ suggestion last week to have the owner column on the keys instead of the Project/User association and this was repeatedly turned down by you. I even stressed that turning my proposal down would require this additional association table.  

I yesterday went almost mad solving all the little tiny issues arising from this way of building it. It caused me ~6h of despair to address the "Project Owner problem" and how to resolve the joins. 

Having the owner column on the keys for the invites but on the ProjectUser association table for the Users would imply changing a ton of function that need to process instances of both `Invite` and `User` objects. If we are consistent in this regard, we can be basically use the same code.

dds_web/api/user.py
MatthiasZepper: Sure, but this is essentially code from @inaod568, so I kept the `is` as it is now:  
https://github.com/ScilifelabDataCentre/dds_web/blob/b573bec26d44a3fd159950d638025e2b1d453086/dds_web/api/user.py#L171-L181

dds_web/database/models.py
alneberg: I see that I didn't make myself clear regarding this before. I agreed that we shouldn't push for the owner column solution on the previous PR on the invite keys, but instead do it in a new PR. Again, sorry for not being clear on this point. However, if you now think that the the ProjectInvites table is a better solution, I'm all ears.

MatthiasZepper: No, I am not at all saying that the `ProjectInvites` table is the better solution (I still like my initial proposal more), but it is the solution that we have decided on and that I thus implemented. 

The only way to make this whole thing even more complicated is solving the same problem in two different ways for `Invites` and `Users`.

dds_web/api/user.py
```diff
@@ -45,33 +45,42 @@ class AddUser(flask_restful.Resource):
...
-        if existing_user and not project:
-            raise ddserr.DDSArgumentError(
-                message="User exists! Specify a project if you want to add this user to a project."
-            )
-
-        if not existing_user:
-            # Send invite if the user doesn't exist
-            invite_user_result = self.invite_user(args)
-            return invite_user_result, invite_user_result["status"]
+        if existing_user or unanswered_invite:
```
alneberg, line 61: Should we have a check to make sure we don't have both `existing_user` and `unanswered_invite`. That shouldn't be allowed to happen, but doesn't hurt to check for it I guess?

dds_web/api/user.py
```diff
@@ -45,33 +45,42 @@ class AddUser(flask_restful.Resource):
...
+        # A project may or may not be specified
         project = flask.request.args.get("project", None)
+        if project:
+            project = project_schemas.ProjectRequiredSchema().load({"project": project})
+            role = flask.request.json.get("role", None)
+
         args = flask.request.json
         # Check if email is registered to a user
         existing_user = user_schemas.UserSchema().load(args)
+        unanswered_invite = user_schemas.UnansweredInvite().load(args)
```
alneberg, line 60:
```suggestion
        if existing_user and unanswered_invite:
            raise ddserr.DatabaseError(message="Email exists for user and invite at the same time")
```

alneberg: This is untested and probably needs black formatting.

dds_web/api/user.py
alneberg: Added suggestion above.

inaod568 (commented):
Looked through and tested. After you've added the parts that Johannes has commented on I'm ok to approve. 
Have added this card for the continuation:
https://github.com/scilifelabdatacentre/dds_web/issues/895

dds_web/api/user.py
MatthiasZepper: Thank you for the suggestion. In the Unanswered Invite Schema, I have now made sure that the Invite Object is only returned if no user with that email exists:
```
        # double check if there is no existing user with this email
        userexists = utils.email_in_db(email=data.get("email"))

        if userexists and invite:
            raise ddserr.DatabaseError(message="Email exists for user and invite at the same time")

        # if the user exists already, the invite object must not be returned to prevent sign-up
        return invite if not userexists else None
```

I feel this is the better place for it, even though we indeed don't use that schema elsewhere, so we could have it also in the API endpoint code directly.

alneberg: Yes, that's probably a better place for it! 👍

inaod568 (approved):

PR #890, opened 2022-02-16
Database migrations

### LINKED ISSUE #583: Add Database Migrations
Add migrations to create databse schema and to version control the schema. The main benefit is that migrations can alter a database which is already populated without having to recreate the entire database every time.

talavis:
Add support for database migrations via `flask-migrations`/`Alembic`.

Please check that the instructions in README.md are clear enough.

Closes #583 and #892


inaod568 (approved):
Nice. Tried it out and it works as expected.

README.md
```diff
@@ -77,11 +77,30 @@ These values are publicly visible on GitHub and **should not be used in producti
 > :exclamation: <br> 
 > At the time of writing, upload within projects created in the development database will most likely not work. <br>
 > To use the upload functionality with the `CLI`, first create a project.
+
 ### Database changes
 
-If the database is modified, you will need to rebuild the containers from scratch.
+> :heavy_exclamation_mark: Do your database changes in `models.py` while the containers are running (`docker-compose up`). Do not restart them to regenerate the database.
```
alneberg, line 83: Maybe we should change the development flow so that we don't use the `db.create_all()` but instead use alembic to create tables on starting the containers? We can create a card for this though

README.md
```diff
@@ -77,11 +77,30 @@ These values are publicly visible on GitHub and **should not be used in producti
...
+If you modify the database models (e.g. tables or indexes), you must create a migration for the changes. We use `Alembic` (via `flask-migrate`) which compares our database models with the running database to generate a suggested migration.
+
+Run the command `flask db migrate -m <commit message/name>` in the running backend:
+
+```bash
+docker exec dds_backend flask db migrate -m <commit message/name>
+```
 
-First, remove all docker containers and volumes
+This will create a migration in the folder `migrations/versions`. Confirm that the changes in the file match the changes you did, otherwise change the `upgrade` and `downgrade` functions as needed. Keep an eye out for changes to the `apscheduler` tables and indexes, and make sure they are not included in the migration. Once the migration looks ok, test it by running `flask db upgrade` in the backend:
```
alneberg, line 93: One could add here that it's perfectly fine to delete the new version generated and rerun the migrate command to generate a new file.

README.md
```diff
@@ -77,11 +77,30 @@ These values are publicly visible on GitHub and **should not be used in producti
...
+
+```bash
+docker exec dds_backend flask db upgrade
+```
+
+Finally, confirm that the database looks correct after running the migration and commit the migration file to git. Note that you need to run `black` on the generated migration file.
```
alneberg, line 99: During my basic testing it was necessary to run the `flask db upgrade` without shutting down the backend in between generating the migration. If I did shut down the backend in between, I was not able to get it running again without reverting the changes in `models.py`, removing the versions file and running `docker compose down` before starting it again.

alneberg: I think removing the two `db.create_all()` occurencies in the `__init__.py` fixed this issue. Then one needs to add the `flask db upgrade` in the test docker-compose command as well.

README.md
inaod568: Yes, I think that could be good. But yes, that's another card: https://github.com/scilifelabdatacentre/dds_web/issues/892

README.md
inaod568: Hmm strange. Doing the exact procedure that @talavis wrote here worked for me. 
1. Run `docker-compose up`
2. Change in db
3. `docker exec dds_backend flask db migrate -m <commit message/name>` while running
4. Check in migration file and remove the rows about apscheduler
5. `docker exec dds_backend flask db upgrade`

I do agree with https://github.com/ScilifelabDataCentre/dds_web/pull/890/files#r808367032 though -- has been discussed for a while to decouple the db creation from the system startup.

README.md
talavis: The old migrations file also has to be restored to avoid the reference to the new file. I decided to word it as
> If you want to start over, restore the content of `migrations/versions` and start from autogeneration method 2.

README.md
talavis: `db.create_all()` is removed and most of the instructions are reworded.

README.md
talavis: See my other comment

talavis (commented):
`db.create_all()` has been removed and most of the migration text in the readme is reworded. Also updated the production text.

alneberg:
Sorry for keeping this waiting. I think it would be nice to add the upgrade command to the test docker-compose file as well. And also I believe the initial migration needs to be regenerated or complemented with the latest additions to the models.

inaod568 (approved):
Nice! 
Works well when you get the hang of it.

Issue #882, opened 2022-02-15
Add tests to assert that mails have been sent


alneberg:
Simplest would be to mock the mail.send method and asserting that it has been called. Needed for endpoints with:
 - [x] invite 
 - [x] project release
 - [x] token authentication
 - [x] delete user self

alneberg:
#912 ticked the final box for this issue.

PR #891, opened 2022-02-16
Reset password

### LINKED ISSUE #880: Look over reset password token sent
A small part of the #876

alneberg:
Sends an email similarly formatted  as the other emails apart from a red button instead of the regular one. It also includes an all caps warning. The email is sent to only the email specified in the form, not specifically the primary nor all emails for a specific user.  The link in the email contains an encrypted JWT which is valid for 1 hour.

If a new password is entered successfully, the password is updated, the project_user_keys are deleted and a new user key pair is generated.

I also included a few tests for checking that invite emails have been sent when releasing a project, when users are added and when an invite is created. This relates to #876 and closes #880, #882.

Things left to do which is omitted from this PR:
 - ~~Encrypt the new user private key using the users new password~~
 - Add a landing page after resetting password, describing which units that should be contacted to regain access to active projects.


dds_web/security/auth.py
```diff
@@ -84,6 +84,18 @@ def verify_token_no_data(token):
...
+    claims = __verify_general_token(token)
+    user = __user_from_subject(claims.get("sub"))
+    if user:
+        rst = claims.get("rst")
+        del claims
+        gc.collect()
+        if rst and rst == "pwd":
+            return user
+    raise AuthenticationError(message="Invalid token")
+
```
worukan, line 97: If you have this function, then you have to update verify token with the following. Otherwise someone can use the system with a reset token.

```
@auth.verify_token
def verify_token(token):
    """Verify token used in token authentication."""
    claims = __verify_general_token(token=token)

    if claims.get("rst"):
        raise AuthenticationError(message="Invalid token")

    user = __user_from_subject(subject=claims.get("sub"))

    return __handle_multi_factor_authentication(
        user=user, mfa_auth_time_string=claims.get("mfa_auth_time")
    )
```

dds_web/web/user.py
```diff
@@ -319,7 +318,15 @@ def request_reset_password():
     form = forms.RequestResetForm()
     if form.validate_on_submit():
         email = models.Email.query.filter_by(email=form.email.data).first()
-        dds_web.utils.send_reset_email(email_row=email)
+        token = dds_web.security.tokens.encrypted_jwt_token(
+            username=email.user.username,
+            sensitive_content="",
```
worukan, line 323: sensitive_content=None,

dds_web/web/user.py
```diff
@@ -339,18 +346,29 @@ def reset_password(token):
...
+        # Delete project user keys for user
+        for project_user_key in user.project_user_keys:
+            db.session.delete(project_user_key)
+        db.session.commit()
+
+        # Update user password
         user.password = form.password.data
         db.session.commit()
+
+        dds_web.security.project_user_keys.generate_user_key_pair(user)
```
worukan, line 369: Please don't generate_user_key_pair here, instead
```
user.kd_salt = None
user.nonce = None
user.public_key = None
user.private_key = None

user.password = form.password.data
db.session.commit()
```

tests/test_user_reset_password.py
```diff
@@ -0,0 +1,332 @@
...
+    # Redirection status code
+    assert response.status_code == http.HTTPStatus.OK
+    # Incorrect token should redirect and not lead to form
+    assert flask.request.path == tests.DDSEndpoint.LOGIN
+
+
+def get_valid_reset_token(username, expires_in=3600):
+    return encrypted_jwt_token(
+        username=username,
+        sensitive_content="",
```
worukan, line 81: sensitive_content=None

dds_web/web/user.py
alneberg: So when would it be generated?

dds_web/security/auth.py
alneberg: Yes of course! I'll add a test for it as well.

alneberg: Fixed!

dds_web/web/user.py
alneberg: Fixed!

tests/test_user_reset_password.py
alneberg: FIxed!

dds_web/web/user.py
worukan: It's in my key derivation PR. Keys are generated if the lines that I wrote above come in that order. Key generation is triggered from `user.password = form.password.data`

dds_web/templates/mail/password_reset.html
```diff
@@ -0,0 +1,55 @@
...
+        </td>
+    </tr>
+    <tr>
+        <td style=" font-family: arial, sans-serif; font-size: 14px; vertical-align: top;">
+            <p
+                style="font-family: arial, sans-serif; font-size: 14px; font-weight: bold; margin: 0; Margin-bottom: 5px;">
+                WARNING! Password Reset</p>
+            <p
+                style="font-family: arial, sans-serif; font-size: 14px; font-weight: normal; margin: 0; Margin-bottom: 10px;">
+                If you proceed with resetting your password, you will loose access to all data.</p>
```
aanil, line 18:
```suggestion
                If you proceed with resetting your password, you will lose access to all data.</p>
```

dds_web/templates/mail/password_reset.txt
```diff
@@ -0,0 +1,5 @@
+Warning! If you proceed with resetting your password, you will loose access to all data.
```
aanil, line 1:
```suggestion
Warning! If you proceed with resetting your password, you will lose access to all data.
```

dds_web/web/user.py
alneberg: I can change that then. I will need to wait for your PR to have the tests pass though.

alneberg: Ok, fixed now!

inaod568 (approved):
It looks good and it works (apart from the password derivation part). I'm fine with merging this and then fixing the points you added in your description. @alneberg That's what you were suggesting, correct?

inaod568:
I tested my PR (https://github.com/ScilifelabDataCentre/dds_cli/pull/236) with this and it works from what I can tell

PR #910, opened 2022-02-18
Rearrangement and clean up of the token

worukan:
Before submitting a pr:
- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG


dds_web/security/tokens.py
```diff
@@ -17,7 +17,11 @@
 
 # Functions ############################################################################ FUNCTIONS #
 def encrypted_jwt_token(
-    username, sensitive_content, expires_in=datetime.timedelta(hours=48), additional_claims=None
+    username,
+    sensitive_content,
+    expires_in=datetime.timedelta(hours=48),
+    additional_claims=None,
+    fully_authenticated=False,
```
MatthiasZepper, line 24: I like that solution a lot! It allows to keep the invite/deletion/password reset tokens short, while the regular tokens contain the additional information for a better user experience.

dds_web/security/tokens.py
```diff
@@ -27,18 +31,21 @@ def encrypted_jwt_token(
...
     """
+    jwe_protected_header = {
+        "alg": "A256KW",
+        "enc": "A256GCM",
+    }
+    if fully_authenticated:
+        # exp claim in this (integrity) protected JWE header is provided only to let the
+        # cli know the precise expiration time of the encrypted token. It has no impact
+        # on the actual enforcement of the expiration of the token.
+        jwe_protected_header["exp"] = (dds_web.utils.current_time() + expires_in).timestamp()
```
MatthiasZepper, line 45: Given that you have come up with this really clever way of balancing token length vs. additional metadata for a better user experience: Why are you still so restrictive with regard to putting anything else into the header than the expiration time?

tests/test_token.py
```diff
@@ -41,7 +46,7 @@ def test_extract_encrypted_token_sensitive_content_no_token(client):
 def test_encrypted_and_signed_token(client):
     username = "researchuser"
     expires_in = datetime.timedelta(minutes=1)
-    expiry_datetime = datetime.datetime.utcnow() + expires_in
+    expiry_datetime = dds_web.utils.current_time() + expires_in
     additional_claim = {"iss": "DDS"}
```
MatthiasZepper, line 50: I now that you didn't change that and it is only a test, but since I just noticed: 

Would DDS be a valid `iss` claim? Shouldn't that be [a URI in URN format](https://en.wikipedia.org/wiki/Uniform_Resource_Name)?

dds_web/security/tokens.py
worukan: Because it is plaintext and putting more information makes the token longer. We have to carefully think what we really need and if we need to put it in the header or even in the encrypted part of the token.

tests/test_token.py
worukan: This is a bogus test string. One can understand that it is possible to include such additional claims by either looking at a test like this or reading the docstring written for the function.

dds_web/security/tokens.py
MatthiasZepper: I acknowledge the problem of token length, but think there is a difference with regard to tokens that are part of an URL and send out via e-mail and the "regular" MFA tokens that we cache in the CLI. For the latter, I don't see why we have to be _that_ strict in terms of length that we even need to scrap the username, but you have the say in the security aspects.

worukan: We don't want to reveal the usernames unless we really have to.

MatthiasZepper: In which regard do we reveal the username here? In order to authenticate, the user needs to provide it first, no? The `.dds_cli_token` file should only be readable by the user, Johannes specifically wrote a check to set the correct file permissions. 

Since the username is obfuscated in the token and needs to be deserialized with `jwt.JWT(jwt=token)` and `token.token.jose_header` first, I would consider this even slightly better than if we store the entered username locally as you suggested for Anands PR.

worukan: what do you mean by **obfuscated**? 
Deserialize is like you copy-paste the token string at jwt.io and see the plaintext header. It is nothing fancy.

inaod568: @MatthiasZepper https://github.com/ScilifelabDataCentre/dds_cli/pull/244#issuecomment-1047054729

MatthiasZepper: I know that it is not encrypted, but it is also not clear text and will therefore not show up when e.g. searching with `grep `for a particular username on a multi-user system.

I would agree that e.g. on a university cluster with hundreds of users, finding readable token file for a particular user owing to wrongly set file system permissions like this would be a security nightmare.

But I agree with you and Ina that we must not have this discussion now. It is a waste of precious time for you guys. If the two of you prefer to have a separate username retrieval endpoint in the API to save <210 Bytes in token length, I will accept that I was outvoted.

inaod568 (approved):

MatthiasZepper (approved):

PR #911, opened 2022-02-18
Invite researchers to projects 

inaod568:
Before submitting a pr:
- [ ] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG


dds_web/api/schemas/user_schemas.py
```diff
@@ -219,6 +219,12 @@ def make_user(self, data, **kwargs):
         # Verify and transfer invite keys to the new user
         if verify_and_transfer_invite_to_user(token, new_user, data.get("password")):
             for project_invite_key in invite.project_invite_keys:
+                if isinstance(new_user, models.ResearchUser):
+                    project_user = models.ProjectUsers(
+                        project_id=project_invite_key.project_id, owner=project_invite_key.owner
+                    )
```
worukan, line 225: I don't know how you have this working. You have written a similar line like this before that I tried to use and it didn't work because for example now you are trying to create ProjectUsers object which is missing the primary key user_id. That just didn't work for me and honestly I wouldn't expect it to work. Just sharing my observation.

inaod568: That's very strange that it's not working..
The `project_user` is added in the line 226 below, that's how it's working. At least for me I get a row in the ProjectUser table

dds_web/api/user.py
```diff
@@ -47,62 +48,103 @@ class AddUser(flask_restful.Resource):
...
+
+        # Verify role or current and new user
+        if current_user_role == "Super Admin" and project:
             return {
-                "message": invite_err.description,
                 "status": ddserr.InviteError.code.value,
+                "message": (
+                    "Super Admins do not have project data access and can therefore "
+                    "not invite users to specific projects."
+                ),
```
worukan, line 121: Why is this error written different from the others? Any special reason?
For example, why not `raise InviteError(message="Super Admins do not have project data access and can therefore not invite users to specific projects.")`

dds_web/api/user.py
```diff
@@ -47,62 +48,103 @@ class AddUser(flask_restful.Resource):
...
+            if new_user_role in ["Super Admin", "Unit Admin", "Unit Personnel"]:
+                raise ddserr.AccessDeniedError()
+        elif current_user_role == "Researcher":
+            raise ddserr.AccessDeniedError()
+
+        # Create invite row
+        new_invite = models.Invite(
+            email=email,
+            role=(
+                "Researcher" if new_user_role in ["Researcher", "Project Owner"] else new_user_role
```
worukan, line 140: `role=("Researcher" if new_user_role == "Project Owner" else new_user_role),`

dds_web/api/user.py
inaod568: The raises are actually incorrect (or actually it's better to raise and then catch but this is another refactoring task), the `invite_user` function needs to be called multiple times from for exemple the CreateProject endpoint. And it shouldn't raise anything for a specific user getting invited.

worukan (approved):
Good job!

PR #913, opened 2022-02-19
Change non sensitive

inaod568:
Before submitting a pr:
- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG


migrations/versions/a5a40d843415_changed_sensitive.py
```diff
@@ -0,0 +1,47 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "projects",
+        sa.Column(
+            "is_sensitive", mysql.TINYINT(display_width=1), autoincrement=False, nullable=True
+        ),
+    )
+    op.drop_column("projects", "non_sensitive")
+    op.create_table(
+        "apscheduler_jobs",
```
alneberg, line 36: I think you need to remove the `apscheduler_jobs` parts in both migrations.

migrations/versions/a5a40d843415_changed_sensitive.py
```diff
@@ -0,0 +1,47 @@
...
+# revision identifiers, used by Alembic.
+revision = "a5a40d843415"
+down_revision = "d117e6299dc9"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("projects", sa.Column("non_sensitive", sa.Boolean(), nullable=True))
```
talavis, line 21: Shouldn't the value be converted from the old one to the new one, rather than be dropped completely?

migrations/versions/a5a40d843415_changed_sensitive.py
```diff
@@ -0,0 +1,47 @@
...
+        "apscheduler_jobs",
+        sa.Column("id", mysql.VARCHAR(length=191), nullable=False),
+        sa.Column("next_run_time", mysql.DOUBLE(asdecimal=True), nullable=True),
+        sa.Column("job_state", sa.BLOB(), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+        mysql_default_charset="utf8mb4",
+        mysql_engine="InnoDB",
+    )
+    op.create_index(
+        "ix_apscheduler_jobs_next_run_time", "apscheduler_jobs", ["next_run_time"], unique=False
```
talavis, line 45: Like Johannes said
```suggestion
```

migrations/versions/c3d3e960fc1b_sensitive_not_nullable.py
```diff
@@ -0,0 +1,39 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "c3d3e960fc1b"
+down_revision = "a5a40d843415"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
talavis, line 20: Seems like this migration does nothing?

migrations/versions/a5a40d843415_changed_sensitive.py
```diff
@@ -0,0 +1,47 @@
...
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "projects",
+        sa.Column(
+            "is_sensitive", mysql.TINYINT(display_width=1), autoincrement=False, nullable=True
+        ),
+    )
+    op.drop_column("projects", "non_sensitive")
```
talavis, line 34: Same as above: shouldn't we keep the values (reversed)?

migrations/versions/c3d3e960fc1b_sensitive_not_nullable.py
talavis: The autogenerate function probably couldn't detect what you did. I suggest you remove this file and add that requirement to the earlier migration

tests/test_project_creation.py
```diff
@@ -83,7 +83,7 @@ def test_create_project_with_credentials(client, boto3_session):
     assert (
         created_proj
         and created_proj.date_created > time_before_run
-        and not created_proj.is_sensitive
+        and not created_proj.non_sensitive
```
talavis, line 86:
```suggestion
        and created_proj.non_sensitive
```
Not sure what it should be here (only looking at the diff), but shouldn't it be that way to maintain the meaning?

talavis (changes requested):
See my comments. The important things:
* remove apscheduler from the migrations
* remove the migration that does nothing
* keep the values when the columns are dropped/added
* make sure that "sensitive_not_nullable" (I guess default=x) is included in the first migration

tests/test_project_creation.py
talavis: Or maybe it's simply checking the defaults?

migrations/versions/a5a40d843415_changed_sensitive.py
```diff
@@ -0,0 +1,47 @@
...
+        sa.Column("id", mysql.VARCHAR(length=191), nullable=False),
+        sa.Column("next_run_time", mysql.DOUBLE(asdecimal=True), nullable=True),
+        sa.Column("job_state", sa.BLOB(), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+        mysql_default_charset="utf8mb4",
+        mysql_engine="InnoDB",
+    )
+    op.create_index(
+        "ix_apscheduler_jobs_next_run_time", "apscheduler_jobs", ["next_run_time"], unique=False
+    )
```
talavis, line 46:
```suggestion
    )
```

migrations/versions/a5a40d843415_changed_sensitive.py
inaod568: So... how do I do that?

migrations/versions/c3d3e960fc1b_sensitive_not_nullable.py
inaod568: > add that requirement to the earlier migration

How do I do that? 

Also I wasn't sure that setting it to nullable should be recognized but I guess that makes sense

talavis (approved):
Not tested, but looks ok. Maybe take a look at whether .all() is really needed.

inaod568:
Possibly, but it feels strange since I'm used to flask_sqlalchemy and those queries. 
Feels better to be consistent, `all()`/`first()`/`one()` etc is used throughout the system in all other cases so since this works I'll keep it.

PR #925, opened 2022-02-22
Fix empty requests

inaod568:
Fix handling of errors occurring when  empty requests are made, when incorrect info is put into Schemas etc. 

Before submitting a pr:
- [x] Tests passing
- [ ] Black formatting
- [ ] Migrations for any changes to the database schema
- [x] Rebase/merge the `dev` branch
- [ ] Note in the CHANGELOG

Issue #924, opened 2022-02-22
--no-mail flag


inaod568:
On request from NGI, make it possible to skip sending an email to users when a project has become available. This is just so that the unit can send the mail themselves.

PR #926, opened 2022-02-22
Issue924 no mail flag

MatthiasZepper:
Before submitting a pr:
- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG

Introduced a `--no-mail` flag in the CLI respectively a `send_email: True/False` json parameter to fix [issue 924](https://github.com/scilifelabdatacentre/dds_web/issues/924) ([#925](https://github.com/ScilifelabDataCentre/dds_web/pull/925))

### DIFF
--- a/tests/test_project_status.py
+++ b/tests/test_project_status.py
@@ -367,7 +367,7 @@ def test_set_project_to_available_valid_transition(module_client, test_project):
     assert db_deadline == calc_deadline
 
 
-def test_set_project_to_available_no_mail(client):
+def test_set_project_to_available_no_mail(client, boto3_session):
     """Set status to Available for test project, but skip sending mails"""
 
     response = client.post(
@@ -393,8 +393,8 @@ def test_set_project_to_available_no_mail(client):
                 json={"new_status": "Available", "deadline": 10, "send_email": False},
             )
         # assert that no mail is being sent.
-        assert mock_mail_send.call_count == 0
         assert mock_mail_func.called == False
+        assert mock_mail_send.call_count == 0
 
     assert response.status_code == http.HTTPStatus.OK
     assert "An e-mail notification has not been sent." in response.json["message"]
--- a/tests/test_project_status.py
+++ b/tests/test_project_status.py
@@ -367,12 +367,12 @@ def test_set_project_to_available_valid_transition(module_client, test_project):
     assert db_deadline == calc_deadline
 
 
-def test_set_project_to_available_no_mail(client, boto3_session):
+def test_set_project_to_available_no_mail(module_client, boto3_session):
     """Set status to Available for test project, but skip sending mails"""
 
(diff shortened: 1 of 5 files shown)


inaod568 (approved):
Nice! Clean and simple

MatthiasZepper:
> Nice! Clean and simple

Thanks. Need to still add a test for the other endpoint though. So far only the `AddUser` one is tested - and I also need to sign still.

inaod568:
> > Nice! Clean and simple
> 
> Thanks. Need to still add a test for the other endpoint though. So far only the `AddUser` one is tested - and I also need to sign still.

Ok! I'll mark it as draft until you've added the additional test(s) 👍🏻

inaod568 (approved):
Looks good! I think we need to add even more tests for the no-mail flag, but I'm adding that as a card for later!

tests/test_project_status.py
```diff
@@ -365,6 +367,39 @@ def test_set_project_to_available_valid_transition(module_client, test_project):
     assert db_deadline == calc_deadline
 
 
+def test_set_project_to_available_no_mail(client):
```
alneberg, line 370:
```suggestion
def test_set_project_to_available_no_mail(client, test_project):
```

aanil: I think it should be boto3_session as that is what we use in the project creation tests.
```suggestion
def test_set_project_to_available_no_mail(client, boto3_session):
```

tests/test_project_status.py
```diff
@@ -365,6 +367,39 @@ def test_set_project_to_available_valid_transition(module_client, test_project):
...
+        ) as mock_mail_func:
+            response = client.post(
+                tests.DDSEndpoint.PROJECT_STATUS,
+                headers=tests.UserAuth(tests.USER_CREDENTIALS["unituser"]).token(client),
+                query_string={"project": public_project_id},
+                json={"new_status": "Available", "deadline": 10, "send_email": False},
+            )
+        # assert that no mail is being sent.
+        assert mock_mail_send.call_count == 0
+        assert mock_mail_func.called == False
```
alneberg, line 397:
```suggestion
            # assert that no mail is being sent.
            assert mock_mail_func.called == False

        assert mock_mail_send.call_count == 0
```

alneberg: I think it makes sense for the `mock_mail_func` to be asserted inside the `with` statement and not outside of it.

tests/test_project_status.py
MatthiasZepper: Thanks a lot! I now went with `def test_set_project_to_available_no_mail(client, boto3_session)`

tests/test_project_status.py
MatthiasZepper: Absolutely! Wrong indentation here.... 😳 

I basically sabotaged my own test! 😖

PR #935, opened 2022-02-24
create new superadmin

inaod568:
Adding config to have a superadmin email added too.

- [ ] Tests passing
- [ ] Black formatting
- [ ] Note in the CHANGELOG


dds_web/web/user.py
```diff
@@ -204,7 +204,7 @@ def confirm_2fa():
     # Valid 2fa initiated token, but user does not exist (should never happen)
```
alneberg, line 204: I guess you could change this message if you expect it to actually happen.

alneberg (approved):
Looks good to me.

migrations/versions/666003748d14_change_active_nullable_and_default.py
```diff
@@ -0,0 +1,43 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    session = Session(bind=op.get_bind())
+    all_user_rows = session.query(models.User).all()
+    for user in all_user_rows:
+        if user.active is None:
+            user.active = False
+    session.commit()
+    op.alter_column("users", "active", existing_type=mysql.TINYINT(display_width=1), nullable=False)
```
talavis, line 30: Not sure: should the new default value be mentioned here, or is that only in the model?

talavis (approved):
It seems to work fine

migrations/versions/666003748d14_change_active_nullable_and_default.py
inaod568: It seems like it. Not adding / looking into this now though

PR #936, opened 2022-02-24
TOTP authentication

alneberg:
A user can request to switch from OTP via email to using TOTP. This request is verified via email. After this, only TOTP can be used for MFA. Note: I needed to use the current setting for TOTP due to restrictions in (at least my version of) Google Authenticator app:

```python
twofactor_totp.TOTP(os.urandom(20), 6, hashes.SHA1(), 30)
```

I tried using 8 digits instead of 6 and also using `hashes.SHA256` or `hashes.SHA512` but those settings are ignored by the phone app when the QRcode is scanned.

When verifying a OTP, three time-frames are checked: 30 seconds earlier, now and 30 seconds later. This is to be slightly more tolerating against clock drift. Due to this and to avoid a code to be re-used, a new verification cannot be done within 90 seconds of a previously successful one for a specific user.

Left to do for this PR:
- [x] Add tests
- [x] Adjust the web confirm_2fa
- [x] Add the no-cache feature that was in place before from @inaod568 
- [x] Fix styling of email for activating TOTP
- [x] Fix styling of form for activating TOTP
- [x] Fix weird error when attempting to authenticate with HOTP when TOTP is activated
- [x] Create informative error message when authenticating with TOTP too soon after previous success
- [x] Better interface for activating TOTP in the CLI
- [x] localhost instead of dds_backend in the activation email?
- [x] Make sure no HOTP email is sent when TOTP is activated

Left to do (perhaps not within this PR):
- [x] Make it possible to switch back to OTP via email
- [x] Possible to clean up TOTP authentication in the CLI?

The regular things:

- [ ] Tests passing
- [x] Black formatting
- [ ] Note in the CHANGELOG

<a href="https://gitpod.io/#https://github.com/ScilifelabDataCentre/dds_web/pull/936"></a>


dds_web/database/models.py
```diff
@@ -486,6 +494,80 @@ def verify_HOTP(self, token):
         self.hotp_issue_time = None
         db.session.commit()
 
+    @property
+    def totp_initiated(self):
+        """To check if activation of TOTP has been initiated, not to be confused
+        with user.totp_enabled, that indicates if TOTP is successfully enabled for the user."""
+        return self._totp_secret is not None
```
inaod568, line 501: Do we need both the property `totp_initiated` and the column `totp_enabled`?

dds_web/api/__init__.py
```diff
@@ -67,6 +67,9 @@ def output_json(data, code, headers=None):
 api.add_resource(user.DeleteUserSelf, "/user/delete_self", endpoint="delete_user_self")
 api.add_resource(user.RemoveUserAssociation, "/user/access/revoke", endpoint="revoke_from_project")
 api.add_resource(user.UserActivation, "/user/activation", endpoint="user_activation")
+api.add_resource(
+    user.RequestTOTPActivation, "/user/request_activate_totp", endpoint="request_totp_activation"
```
inaod568, line 71: Suggestion: `/user/totp/request` or something

dds_web/api/user.py
```diff
@@ -784,13 +784,75 @@ class SecondFactor(flask_restful.Resource):
     @handle_validation_errors
     def get(self):
 
-        token_schemas.TokenSchema().load(flask.request.json)
+        try:
+            token_schemas.TokenSchema().load(flask.request.json)
+        except marshmallow.ValidationError as err:
+            raise ddserr.AuthenticationError(message=err.messages)
```
inaod568, line 790: Isn't this already handled in the `@handle_validation_errors`?

dds_web/database/models.py
```diff
@@ -486,6 +494,80 @@ def verify_HOTP(self, token):
...
+    def setup_totp_secret(self):
+        """Generate random 160 bit as the new totp secret and return provisioning URI
+        We're using SHA1 (Google Authenticator seems to only use SHA1 and 6 digit codes)
+        so secret should be at least 160 bits
+        https://cryptography.io/en/latest/hazmat/primitives/twofactor/#cryptography.hazmat.primitives.twofactor.totp.TOTP
+        """
+        self._totp_secret = os.urandom(20)
+        db.session.commit()
+
+    def get_totp_secret(self):
```
inaod568, line 512: Suggestion:
```
@property
def totp_secret(self):
```
?

dds_web/api/user.py

dds_web/api/__init__.py
aanil: Done!

dds_web/database/models.py
aanil: Done!

dds_web/database/models.py
```diff
@@ -491,6 +499,92 @@ def verify_HOTP(self, token):
...
+        to allow for some clock drift.
+
+        raises AuthenticationError if token is invalid, has expired or
+        if totp has been successfully verified within the last 90 seconds.
+        """
+        # can't use totp successfully more than once within 90 seconds.
+        # Time frame chosen so that no one can use the same token more than once
+        # No need to use epoch time here.
+        current_time = dds_web.utils.current_time()
+        if self.totp_last_verified is not None and (
```
inaod568, line 559:
```suggestion
        if self.totp_last_verified and (
```

requirements.txt
```diff
@@ -40,6 +40,7 @@ MarkupSafe==2.0.1
 marshmallow==3.14.1
 marshmallow-sqlalchemy==0.27.0
 packaging==21.3
+Pillow==9.0.1
```
inaod568, line 43: Is this used?

inaod568 (changes requested):
First review, need to dig into it more and then give a more thorough review.

1. We should probably think of a way to log the user out / force reauthentication when they have changed the 2fa method
2. It's good that there's a check for if the totp is already activated, but we should add a way of resetting it at some point. I don't mean move to hotp again but reseting the totp

requirements.txt
inaod568: Required by `qrcode`

dds_web/api/user.py
```diff
@@ -982,6 +984,79 @@ def get(self):
         return {"token": update_token_with_mfa(token_claims)}
 
 
+class RequestTOTPActivation(flask_restful.Resource):
+    """Request to switch from HOTP to TOTP for second factor authentication."""
+
+    @auth.login_required
```
inaod568, line 990:
```suggestion
    @auth.login_required
    @json_required
```

inaod568, line 990:
```suggestion
    @auth.login_required
    @json_required
```

dds_web/api/user.py
```diff
@@ -982,6 +984,79 @@ def get(self):
...
+                f"mail/request_activate_totp.html",
+                link=link,
+            )
+
+            AddUser.send_email_with_retry(msg)
+            return {
+                "message": "Please check your email and follow the attached link to activate two-factor with authenticator app."
+            }
+        else:
+            if not user.totp_enabled:
```
inaod568, line 1052: This is just a personal preference but it would make more sense to me to have it like this instead

```
if user.totp_enabled:
    user.deactivate_totp()
    return ...

return ....
```

inaod568, line 1052: This is just a personal preference but it would make more sense to me to have it like this instead

```
if user.totp_enabled:
    user.deactivate_totp()
    return ...

return ....
```

inaod568, line 1052: This is just a personal preference but it would make more sense to me to have it like this instead

```
if user.totp_enabled:
    user.deactivate_totp()
    return ...

return ....
```

dds_web/api/user.py
```diff
@@ -982,6 +984,79 @@ def get(self):
         return {"token": update_token_with_mfa(token_claims)}
 
 
+class RequestTOTPActivation(flask_restful.Resource):
+    """Request to switch from HOTP to TOTP for second factor authentication."""
+
+    @auth.login_required
+    def put(self):
```
inaod568, line 991: Not sure about "put" since we're not making any DB changes etc

dds_web/web/user.py
```diff
@@ -148,6 +152,81 @@ def register():
...
+    form = forms.ActivateTOTPForm()
+
+    dds_web.security.auth.verify_activate_totp_token(token, current_user=user)
+
+    if user.totp_enabled:
+        flask.flash("Two-factor authentication via authenticator app is already enabled.")
+        return flask.redirect(flask.url_for("pages.home"))
+
+    # Don't change secret on page reload
+    if not user.totp_initiated:
```
inaod568, line 174: Should there be an else for this too? Or is the idea that it's fine if it's already initiated?

inaod568: Also, when is totp_initiated set to True?

dds_web/database/models.py
```diff
@@ -491,6 +499,92 @@ def verify_HOTP(self, token):
...
+        to allow for some clock drift.
+
+        raises AuthenticationError if token is invalid, has expired or
+        if totp has been successfully verified within the last 90 seconds.
+        """
+        # can't use totp successfully more than once within 90 seconds.
+        # Time frame chosen so that no one can use the same token more than once
+        # No need to use epoch time here.
+        current_time = dds_web.utils.current_time()
+        if self.totp_last_verified and (
```
inaod568, line 559: Should it be `and`? Not sure I follow?

dds_web/security/auth.py
```diff
@@ -266,7 +278,8 @@ def __handle_multi_factor_authentication(user, mfa_auth_time_string):
             if mfa_auth_time >= dds_web.utils.current_time() - MFA_EXPIRES_IN:
                 return user
 
-        send_hotp_email(user)
+        if not user.totp_enabled:
+            send_hotp_email(user)
```
inaod568, line 282: This returns True or False, so then it continues to check the path below, but if it doesn't pass that if statement then it will say "Please check your primary email" which would be weird if totp is enabled,

dds_web/database/models.py
alneberg: Yeah, I think it's a check to make sure `totp_last_verified` is not null, that is, if totp hasn't been used yet. So the check is: "have TOTP been used before and was it too recently used"

inaod568:
@zishanmirza commented the following in the Slack dds-development channel:

> In the web, there is no test if the user already has a TOTP already setup or requesting activation. (edited) 
> Deactivation of TOTP is not tested.

dds_web/web/user.py
```diff
@@ -148,6 +152,102 @@ def register():
     return flask.render_template("user/register.html", form=form)
 
 
+@auth_blueprint.route("/activate_totp/<token>", methods=["GET", "POST"])
+@limiter.limit(
+    dds_web.utils.rate_limit_from_config,
+    methods=["GET", "POST"],
```
inaod568, line 158: Is this part needed?

inaod568: Only the `methods=["GET", "POST"]`  - it's defined in the route

inaod568: Also not a big thing though, just a question

inaod568 (changes requested):
I think it looks great and I haven't found an issue. 
We need to add the following though:
1. Tests
    a. Request for activation of TOTP
    b. Activation of TOTP
    c. Request for activation of HOTP 
    d. Activation of HOTP 
2.. Row in changelog

dds_web/web/user.py
aanil: Wouldn't it be for the limiter in this case? The  `confirm_2fa()` method is written the same way.

inaod568: Yeah my question regards both of them. All the other endpoints aren't written like that so I'm assuming it's not needed.

inaod568 (approved):
Looks great!
Noticed an edge case (no need to fix it now, adding it to the board): If I request HOTP as the 2FA method when I already have that, a 2fa email is still sent. Should probably change that.

Issue #949, opened 2022-02-28
Adressing logging for so far uncaught exceptions


MatthiasZepper:
We currently have an issue with some uncaught exceptions, which will not produce helpful logging messages for trouble-shooting. For that reason, I will try to modify the logging accordingly to address those.

inaod568:
In that case it's the general logger right? Not the action logger.

inaod568:
@MatthiasZepper 
Ah thanks for the slack link! 
Now I'm up to speed. Yes, what you and Phil suggested would work.

inaod568:
Fixed with #958

PR #958, opened 2022-02-28
Action log entry for uncaught exceptions.

MatthiasZepper:
Addresses issue  #949. After a lot of trial and error with manual error and Traceback/StackSummary formatting, I went with the default formatting provided by `structlog`. 

The log entry for an uncaught exception will look like this then:
```

{"resource": "/api/v1/proj/list", "project": null, "user": "unituser_1", "exception": "Traceback (most recent call last):\n  File \"/code/dds_web/api/dds_decorators.py\", line 162, in wrapper_logging_bind_request\n    value = func(*args, **kwargs)\n  File \"/code/dds_web/api/project.py\", line 289, in get\n    total_bhours_db = 1 / 0.0\nZeroDivisionError: float division by zero", "event": "Uncaught exception in api_blueprint.list_projects.get", "logger": "actions", "level": "error", "timestamp": "2022-02-28T16:24:20.761019Z"}
```

It is now only an addition to the decorator, thus misses out on all non-decorated endpoints (only a few though). On [Stack Overflow](https://stackoverflow.com/questions/8050775/using-pythons-logging-module-to-log-all-exceptions-and-errors/68293938#68293938) I also found a demo to wrap the system’s except hook - this would be an alternative approach. Any opinions on that?

- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,3 +28,4 @@ Please add a _short_ line describing the PR you make, if the PR implements a spe
 * Invite Unit Admin (temporary way) ([#938](https://github.com/ScilifelabDataCentre/dds_web/pull/938))
 * Add support for getting IPs from X-Forwarded-For ([#952](https://github.com/ScilifelabDataCentre/dds_web/pull/952))
 * Relax requirements for usernames (wider length range, `.` and `-`) ([#943](https://github.com/ScilifelabDataCentre/dds_web/pull/943))
+* Extended the `dds_web.api.dds_decorators.logging_bind_request` decorator to catch all not yet caught exceptions and make sure they will be logged ([#958](https://github.com/ScilifelabDataCentre/dds_web/pull/958)).
--- a/dds_web/api/dds_decorators.py
+++ b/dds_web/api/dds_decorators.py
@@ -6,6 +6,8 @@
 
 # Standard library
 import functools
+import sys
+import traceback
 
 # Installed
 import boto3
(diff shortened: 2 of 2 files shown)


MatthiasZepper:
@inaod568: Thanks for merging. 

With regard to the `handle_validation_errors` decorator, I see your point. Although I am still confused about the way how combinations of multiple decorators work, I think just changing the order of the decorators will do in that case? 

In case you are adressing that, you might want to change the `isinstance(err, LoggedHTTPException)` in my PR. It should probably better read  `issubclass(type(err), LoggedHTTPException)`, because I just realized that we never raise the LoggedHTTPException directly. It is just the superclass of the respective HTTP exceptions...

inaod568:
@MatthiasZepper Ah! I didn't notice this was an issue actually. I'll try it out and in the case there's an issue and we need to fix it I'll create a card.

inaod568:
@MatthiasZepper 

```
                raise DatabaseError(message="test")
                return value

            except Exception as err:
                flask.current_app.logger.debug(f"Is LoggedHTTPException?: {isinstance(err, LoggedHTTPException)}")
                if not isinstance(err, LoggedHTTPException):
                    flask.current_app.logger.debug("It's in here")
```

Log: 
```
[2022-03-01 20:29:58,544] dds_decorators [DEBUG] Is LoggedHTTPException?: True
```

So it's fine because of the inheritance 👍🏻 
but it

MatthiasZepper:
> So it's fine because of the inheritance 👍🏻 but it

Thanks for the in-depth investigation!

PR #975, opened 2022-03-02
Custom registration form error messages

inaod568:
Added custom messages. Currently not showing up, for them to actually be visible we need to add `novalidate` to the html form. 

- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG

### DIFF
--- a/dds_web/templates/user/register.html
+++ b/dds_web/templates/user/register.html
@@ -1,82 +1,83 @@
 <html>
 
-<head>
-    <title>Registration form</title>
-</head>
+    <head>
+        <title>Registration form</title>
+    </head>
 
-<body>
-    <h1>Create your DDS account</h1>
-    {% with messages = get_flashed_messages(with_categories=true) %}
-    {% if messages %}
-    <div class=flashes>
-        {% for category, message in messages %}
-        <div class="alert alert-{{category}}" role="alert">
-            {{ message }}
+    <body>
+        <h1>Create your DDS account</h1>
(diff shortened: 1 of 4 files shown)


ewels:
Nice! Much of this validation we can also do in the front-end if we want, both with native browser controls and maybe some simple javascript. Can be nice to get immediate validation errors before submitting the form.

Any chance the things like username / password lengths etc could be put in some kind of config variable to make them easier to reuse?

ewels:
Sorry, not in the output ([these work as html attributes](https://www.w3schools.com/tags/att_input_minlength.asp) so can just be printed into the HTML template), I more just mean in the Python so that we don't have to keep 3 lots of validation parameters in sync..

inaod568:
@ewels Yeah started looking into that too but I think I'll create a card for this for now. But definitely a good point, should add some config for username, password etc.

inaod568:
@ewels https://github.com/scilifelabdatacentre/dds_web/issues/978

dds_web/templates/user/register.html
```diff
@@ -1,82 +1,82 @@
...
-        Unit name:
-        {{ form.unit_name }}
-        {% endif %}
-        <br>
-        <!-- Submit button -->
-        {{ form.submit }}
+            <!-- Unit -->
+            {% if unit %}
+            Unit name:
+            {{ unit }}
```
aanil, line 73: 👍

aanil (approved):
👍 Looks ok!

PR #994, opened 2022-03-04
Run pip install before injecting into the container

talavis:
As suggested by @ewels the `pip install` command will be run during container creation instead of having to run it after starting a shell in the container.

- [x] Tests passing
- [x] Black formatting
- [ - ] Note in the CHANGELOG


inaod568 (approved):
Works like a charm!

PR #1013, opened 2022-03-07
Fix the dark mode switch behaviour

ewels:
Rewrote the Javascript that handles the dark mode switch:

- Use javascript, not jQuery
- If no cookie is set:
    - Set the switch status on page load if OS is dark mode
    - Listen to changes to the OS theme and toggle the switch automatically
- On click, set theme and cookie
- Resolved some styling differences between auto-dark-mode and manual-dark-mode
    - In the `dds-auto.css` file the light CSS is always loaded, then we load dark mode on top if request by the OS.
    - I played with making both conditional, but I'm worried that it could have weird side effects for (a) super old / weird browsers and (b) stuff like printing.
    - Instead I just fixed the inconsistencies I could find, like `.bg-light`
 
Javascript could probably be a bit more efficiently written, but seems to work.

- [x] Tests passing
- [x] Black formatting
- [ - ] Note in the CHANGELOG

### DIFF
--- a/dds_web/static/js/dds.js
+++ b/dds_web/static/js/dds.js
@@ -23,22 +23,57 @@
   });
 })();
 
+//
+// Dark mode switch
+//
+
+// OS set to dark mode
+if (window.matchMedia && window.matchMedia("(prefers-color-scheme: dark)").matches) {
+  // Only continue if we have no cookie set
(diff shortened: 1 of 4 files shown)


inaod568 (approved):
Looks good and works 👍🏻

talavis (approved):

PR #1017, opened 2022-03-07
Delete bucket

inaod568:
Delete the bucket and its contents when deleting / archiving the project. The bucket needs to be empty in order to delete it. 

- [x] Tests passing
- [x] Black formatting
- [ - ] Note in the CHANGELOG

### DIFF
--- a/dds_web/api/api_s3_connector.py
+++ b/dds_web/api/api_s3_connector.py
@@ -9,6 +9,7 @@ import logging
 import traceback
 import pathlib
 import json
+import gc
 
 # Installed
 import botocore
@@ -94,6 +95,10 @@ class ApiS3Connector:
 
         # Delete bucket
         bucket.delete()
+
+        # Clean up
+        del bucket
+        gc.collect()
 
     @bucket_must_exists
     def remove_multiple(self, items, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


dds_web/api/api_s3_connector.py
```diff
@@ -84,11 +85,21 @@ def get_s3_info(self):
...
+        # Get bucket object
         bucket = self.resource.Bucket(self.project.bucket)
+
+        # Delete objects first
         bucket.objects.all().delete()
 
+        # Delete bucket
+        bucket.delete()
+
+        # Clean up
```
talavis, line 99: Is this really needed?

inaod568: The clean up? It doesn't hurt, the bucket doesn't exist so there's no need to keep it in memory

talavis: The effect should be the same if you just set it to None, or even just leave it.

inaod568: Yes, basically the same. So I will keep this. Plus this is the method we use in multiple other places in the code.

talavis (approved):

PR #1021, opened 2022-03-08
Issue1007 input validation

MatthiasZepper:
- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG


talavis (commented):
Maybe it makes even more sense to also have it for the name field for users (in the web form as well)?

inaod568 (approved):
Looks great! And no I don't think we need to do anything in the CLI. So this is good! 

_If_ there are any complaints in the future that we don't allow certain characters we can _maybe_ change that then.

inaod568:
> Maybe it makes even more sense to also have it for the name field for users (in the web form as well)?

Yeah we probably need to add this in more places but there are quite a few schemas and forms, so I think it's good to keep the PRs as small as possible. So this is fine for this PR 👍🏻

inaod568:
@MatthiasZepper @talavis https://github.com/ScilifelabDataCentre/dds_web/issues/1022

PR #1029, opened 2022-03-09
Fix regex for file rm/ls

### LINKED ISSUE #1018: Deletion not happening
- [x] Make sure the bucket is deleted when deleting and archiving project @inaod568 
- [x] Regex in RemoveDir endpoint not matching anything that it should match --> not deleting @inaod568

talavis:
Escape the folder name before adding using it in a regex. Avoids missinterpretation of e.g. `[]`.

Reverts #1019 since that change did nothing except making the query a bit uglier in the mariadb log.

Closes #1018 

- [X] Tests passing
- [X] Black formatting
- [-] Migrations for any changes to the database schema
- [x] Note in the CHANGELOG


inaod568 (approved):
Tried uploading and downloading, deleting and listing. The commands that I've tested works with both special characters and without

PR #1045, opened 2022-03-10
Catch KeyNotFoundError

i-oden:
The CLI displays `INTERNAL SERVER ERROR` if a unit user does not have access to a specific project and tries to give another user access. This PR returns the specific errors (should only be one atm) and which projects were affected.

To test: 
1. Use the Super Admin account to add a new Unit Personnel
2. Use the new Unit Personnel account to give another user access to a project 

Needs to be used with https://github.com/ScilifelabDataCentre/dds_cli/pull/379, otherwise it will just say that all is well every time. 

- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG

### DIFF
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -9,7 +9,6 @@ import os
 import smtplib
 import time
 import datetime
-from tkinter.tix import Tree
 
 # Installed
 import flask


CHANGELOG.md
```diff
@@ -45,3 +45,4 @@ Please add a _short_ line describing the PR you make, if the PR implements a spe
 ## Sprint (2022-03-09 - 2022-03-23)
 
 - Introduce a separate error message if someone tried to add an unit user to projects individually. ([#1039](https://github.com/ScilifelabDataCentre/dds_web/pull/1039))
+- Catch KeyNotFoundError when user tries to give access to a project they themselves do not have access to
```
inaod568, line 48:
```suggestion
- Catch KeyNotFoundError when user tries to give access to a project they themselves do not have access to ([#1045](https://github.com/ScilifelabDataCentre/dds_web/pull/1045))
```

dds_web/api/user.py
```diff
@@ -196,46 +196,52 @@ def invite_user(email, new_user_role, project=None, unit=None):
...
-            if project:
-                share_project_private_key(
-                    from_user=auth.current_user(),
-                    to_another=new_invite,
-                    project=project,
-                    from_user_token=dds_web.security.auth.obtain_current_encrypted_token(),
-                    is_project_owner=new_user_role == "Project Owner",
-                )
+            else:
+                db.session.add(new_invite)
```
MatthiasZepper, line 231: Mind that adding the invite to the Database session now happens inside your `try` statement. If a `KeyNotFoundError` is raised, the invite will not be stored in the database (but the mail was already send before)

dds_web/api/user.py
```diff
@@ -196,46 +196,52 @@ def invite_user(email, new_user_role, project=None, unit=None):
...
+                        from_user_token=dds_web.security.auth.obtain_current_encrypted_token(),
+                        is_project_owner=new_user_role == "Project Owner",
+                    )
+        except ddserr.KeyNotFoundError as keyerr:
+            return {
+                "message": "You do not have access to the specified project.",
+                "status": ddserr.AccessDeniedError.code.value,
+            }
 
         db.session.commit()
```
MatthiasZepper, line 246: Do we need to wrap that also into a try statement for `sqlalchemy.exc.SQLAlchemyError` or `sqlalchemy.exc.IntegrityError`?

MatthiasZepper (changes requested):
Not entirely sure, but I think the logic with the try statements, database actions and mail sending should be refactored slightly.

dds_web/api/user.py
```diff
@@ -196,46 +196,52 @@ def invite_user(email, new_user_role, project=None, unit=None):
...
-                    msg = f"{str(new_invite)} was successful, but specification for {str(project)} dropped. Unit Users have automatic access to projects of their unit."
+        try:
+            # Append invite to unit if applicable
+            if new_invite.role in ["Unit Admin", "Unit Personnel"]:
+                # TODO Change / move this later. This is just so that we can add an initial Unit Admin.
+                if auth.current_user().role == "Super Admin":
+                    if unit:
+                        unit_row = models.Unit.query.filter_by(public_id=unit).one_or_none()
+                        if not unit_row:
+                            raise ddserr.DDSArgumentError(message="Invalid unit publid id.")
```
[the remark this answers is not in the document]
inaod568:
```suggestion
                            raise ddserr.DDSArgumentError(message="Invalid unit publid ID.")
```

dds_web/api/user.py
```diff
@@ -196,46 +196,52 @@ def invite_user(email, new_user_role, project=None, unit=None):
...
-                            to_another=new_invite,
-                            from_user_token=dds_web.security.auth.obtain_current_encrypted_token(),
-                            project=unit_project,
-                        )
-
-                if not project:  # specified project is disregarded for unituser invites
-                    msg = f"{str(new_invite)} was successful."
-                else:
-                    msg = f"{str(new_invite)} was successful, but specification for {str(project)} dropped. Unit Users have automatic access to projects of their unit."
+        try:
```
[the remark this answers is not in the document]
inaod568: > I think the try statement in this part needs to be more granular and split into multiple parts only wrapping the respective share_project_private_key() invocations?
>
>Because if it ever happens that a Unit Admin has not full access to every project of a unit, they now can't invite new Unit Personnel successfully anymore because the try will fail.

True, I'll move the try except to wrap the `share_project_private_key` only.

> Maybe it would be better to include the try statement into share_project_private_key() itself and return True or False as success messages that can be evaluated by the endpoints to generate error messages?

No I don't think we should mess with the share_project_private_key functionality and what it returns. It's good that it raises the key exception, and then we should handle the exception where we need to. Some instances may have different needs.

inaod568:
I need to change the tests. The unit user that's being used does not have a key in the table - that's why the tests are failing.

MatthiasZepper (approved):
Very nice. I like the solution with the `goahead `variable!

inaod568:
I don't understand this https://github.com/ScilifelabDataCentre/dds_web/runs/5512442090?check_suite_focus=true#step:3:667

When I run it myself it works. I'm missing something and I'm not going to be able to figure it out today. If anyone else does, let me know. 🙏🏻

MatthiasZepper:
> I don't understand this https://github.com/ScilifelabDataCentre/dds_web/runs/5512442090?check_suite_focus=true#step:3:667
> 
> When I run it myself it works. I'm missing something and I'm not going to be able to figure it out today. If anyone else does, let me know. 🙏🏻

The unit admin you are trying to invite has the wrong role configured in the JSON: `new_unit_admin = {"email": "new_unit_admin@mailtrap.io", "role": "Super Admin"}`. Therefore the `if new_invite.role in ["Unit Admin", "Unit Personnel"]:` fails and the `DDSArgumentError` is never raised.

However, I don't understand why the test passed locally then?

inaod568:
I've added some more tests so if you could have a look again @MatthiasZepper.
If I could get a review from @talavis too that would be great.

dds_web/api/user.py
```diff
@@ -6,9 +6,12 @@
 
 # Standard library
 import os
+from pyexpat import model
```
talavis, line 9: Auto-import*3?

inaod568, line 9:
```suggestion
```

dds_web/api/user.py
```diff
@@ -6,9 +6,12 @@
...
+from re import T
```
inaod568, line 9:
```suggestion
```

dds_web/api/user.py
```diff
@@ -6,9 +6,12 @@
...
 import smtplib
 import time
 import datetime
+from tkinter.tix import Tree
```
inaod568, line 12:
```suggestion
```

talavis (approved):

MatthiasZepper (approved):

PR #1072, opened 2022-03-21
Update text

### LINKED ISSUE #1071: Change "There is no account with that email. To get an account, you need an invitation."
This should be changed to proceeding anyway and displaying
```
If the email is connected to a user within the DDS, you should receive an email with the password reset instructions. 
```
or something similar

zishanmirza:
This patch updates the text.

Fixes #1071.

- [X] Tests passing
- [X] Black formatting
- [ - ] Note in the CHANGELOG


inaod568 (approved):
Looks good. I think we should consider changing the name of this function at some point since it's now only handling the reset password occurrence. But now since it's not used anywhere else this is ok.

PR #1075, opened 2022-03-22
Add check to delete user

inaod568:
Temporary fix until we implement a way to renew the invites. 
This deletes the invite row if you specify is_invite. 

The "The invite has been deleted" independent of whether or not it exists is deliberate.

Signed-off-by: Ina Odén Österbo <inaod568@users.noreply.github.com>

- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG

### DIFF
--- a/dds_web/api/user.py
+++ b/dds_web/api/user.py
@@ -779,14 +779,16 @@ class DeleteUser(flask_restful.Resource):
         try:
             unanswered_invite = user_schemas.UnansweredInvite().load({"email": email})
             if unanswered_invite:
-                if not current_user_role == "Super Admin" or (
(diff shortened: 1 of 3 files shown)


zishanmirza (approved):

dds_web/api/user.py
```diff
@@ -775,6 +786,36 @@ def delete_user(user):
             db.session.rollback()
             raise ddserr.DatabaseError(message=str(err))
 
+    @staticmethod
```
aanil, line 789: Will this function be used by other classes in the future?

inaod568: Yeah that's what I was thinking

PR #1108, opened 2022-03-30
Change to Flask-APScheduler

inaod568:
**It looks like I have changed a lot of files.... I just ran `black .` and this was the result.**
The important parts (where I've actually changed stuff): 
- `scheduled_tasks.py`
- `__init__.py`
- `utils.py`
- `requirements.py`

Change the scheduled tasks to use Flask-Apscheduler instead -- this means that the setup structure is like the migrations, database etc etc. (`init_app(app)`) 
Have also removed the scheduler stuff from the `utils.py` module, and now the scheduled tasks are in a new file, and set up with the help of decorators. 

There may be some more config needed but we'll notice that as we add the tasks. Also, the first task here is just an example and something that @valyo has started on implementing, but since there are more than one that's working on cronjobs etc now, I want to merge this before anyone starts so that it all works. 

The previous solution currently has issues with the app context.

- [ ] Tests passing
- [ ] Black formatting
- [ ] Note in the CHANGELOG

<a href="https://gitpod.io/#https://github.com/ScilifelabDataCentre/dds_web/pull/1108"></a>


inaod568:
Closing this. Look at https://github.com/ScilifelabDataCentre/dds_web/pull/1109 instead

PR #1109, opened 2022-03-30
Flask-Apscheduler

inaod568:
Same as https://github.com/ScilifelabDataCentre/dds_web/pull/1108 without the mess.

- [ ] Tests passing
- [ ] Black formatting
- [ ] Note in the CHANGELOG


valyo (approved):

aanil (approved):
👍

talavis (commented):
I can't get it working in Gunicorn (which is used for production). I have a task that triggers correctly when running via `flask` (standard docker-compose), but it does not trigger when I run in Gunicorn. If we can't get it working in Gunicorn, I do not see a reason to use it at all.

valyo:
> I can't get it working in Gunicorn (which is used for production). I have a task that triggers correctly when running via `flask` (standard docker-compose), but it does not trigger when I run in Gunicorn. If we can't get it working in Gunicorn, I do not see a reason to use it at all.

are you sure that it just doesn't seem that way after Ina's latest commit, which commented out all print/debug statements?

talavis:
> > I can't get it working in Gunicorn (which is used for production). I have a task that triggers correctly when running via `flask` (standard docker-compose), but it does not trigger when I run in Gunicorn. If we can't get it working in Gunicorn, I do not see a reason to use it at all.
> 
> a naive question, but are you sure that it just doesn't seem that way after Ina's latest commit, which commented out all print/debug statements?

```
@scheduler.task("interval", id="print_test", seconds=2, misfire_grace_time=1)
def print_test():
        from datetime import datetime
        with structlog.threadlocal.bound_threadlocal(
            who={"user": "print_test", "role": str(datetime.now())},
            by_whom={"user": "print_test", "role": "scheduler"},
        ):
            action_logger.info("apscheduler print_test")
```
Gives me the following when I start with `flask`:
```
{"who": {"user": "print_test", "role": "2022-03-30 09:29:20.873854"}, "by_whom": {"user": "print_test", "role": "scheduler"}, "event": "apscheduler print_test", "logger": "actions", "level": "info", "timestamp": "2022-03-30T09:29:20.873915Z"}
```
While I only get an empty file if I start with Gunicorn.

requirements.txt
```diff
@@ -19,6 +19,7 @@ email-validator==1.1.3
 factory-boy==3.2.1
 Faker==12.0.0
 Flask==2.0.2
+Flask-APScheduler==1.12.3
```
talavis, line 22:
```suggestion
Flask-APScheduler==1.12.1
```
https://github.com/viniciuschiele/flask-apscheduler/issues/198

Confirmed to work.

talavis (commented):
With our current Gunicorn setup, there will be two schedulers running in parallell, meaning that each task is run twice each timestep. Using my earlier example:
```
{"who": {"user": "print_test", "role": "2022-03-30 10:49:47.573795"}, "by_whom": {"user": "print_test", "role": "scheduler"}, "event": "apscheduler print_test", "logger": "actions", "level": "info", "timestamp": "2022-03-30T10:49:47.573850Z"}
{"who": {"user": "print_test", "role": "2022-03-30 10:49:47.575475"}, "by_whom": {"user": "print_test", "role": "scheduler"}, "event": "apscheduler print_test", "logger": "actions", "level": "info", "timestamp": "2022-03-30T10:49:47.575616Z"}
{"who": {"user": "print_test", "role": "2022-03-30 10:49:49.577357"}, "by_whom": {"user": "print_test", "role": "scheduler"}, "event": "apscheduler print_test", "logger": "actions", "level": "info", "timestamp": "2022-03-30T10:49:49.577386Z"}
{"who": {"user": "print_test", "role": "2022-03-30 10:49:49.580273"}, "by_whom": {"user": "print_test", "role": "scheduler"}, "event": "apscheduler print_test", "logger": "actions", "level": "info", "timestamp": "2022-03-30T10:49:49.580313Z"}
```

talavis (changes requested):
See my suggestion. We need to discuss what to do with the gunicorn setup later.

talavis (approved):

talavis (commented):
It seems that the changes after 1.12.1 fixes the bug we see in our tests. But that fix seems to instead cause the issues in Gunicorn. Yay.

requirements.txt
```diff
@@ -19,6 +19,7 @@ email-validator==1.1.3
 factory-boy==3.2.1
 Faker==12.0.0
 Flask==2.0.2
+Flask-APScheduler==1.12.1
```
talavis, line 22:
```suggestion
Flask-APScheduler==1.12.3
```
Going back to 1.12.3 again. `FLASK_ENV` must be set to `production`; `development` will not work.

PR #1116, opened 2022-03-31
Scheduled job to change status to expired

valyo:
Can probably be done more intelligently but I got an issue with query 
`models.Project.current_deadline <= current_time()`
I can continue trying if/when there's nothing better to do

- [ ] Tests passing
- [ ] Black formatting
- [ ] Note in the CHANGELOG

### DIFF
--- a/dds_web/scheduled_tasks.py
+++ b/dds_web/scheduled_tasks.py
@@ -14,15 +14,12 @@ def set_available_to_expired():
     from dds_web.database import models
     from dds_web.errors import DatabaseError
     from dds_web.api.project import ProjectStatus
-
     from dds_web.utils import current_time, page_query
 
     with scheduler.app.app_context():
 
-        for project in page_query(
-            models.Project.query.filter(models.Project.is_active == 1)
-        ):
-      
+        for project in page_query(models.Project.query.filter(models.Project.is_active == 1)):
+
             if "Available" in project.current_status and project.current_deadline <= current_time():
                 scheduler.app.logger.debug("Handling expiring project")
                 scheduler.app.logger.debug(
--- a/dds_web/scheduled_tasks.py
+++ b/dds_web/scheduled_tasks.py
@@ -1,11 +1,11 @@
 import flask_apscheduler
 import flask
 
-# Apscheduler
+## Apscheduler
 scheduler = flask_apscheduler.APScheduler()
 
 
-@scheduler.task("interval", id="available_to_expired", seconds=50, misfire_grace_time=1)
+@scheduler.task("interval", id="available_to_expired", seconds=5, misfire_grace_time=1)
 def change_status_to_expired():
     # print("Task: Change project status from Available to expired.", flush=True)
     from dds_web.database import models
@@ -14,11 +14,15 @@ def change_status_to_expired():
 
     with scheduler.app.app_context():
         scheduler.app.logger.debug("This means that the app context works!")
-        test = models.ProjectStatuses.query.all()
+        # test = models.Project.query.all()
         # print(test, flush=True)
 
         for project in page_query(
-            models.ProjectStatuses.query.filter(models.ProjectStatuses.deadline >= current_time())
+            # models.ProjectStatuses.query.filter(models.ProjectStatuses.is_active >= current_time())
+            models.Project.query.filter(models.Project.is_active == 1)
         ):
-
-            scheduler.app.logger.debug("Project: %s has stautus %s - Expires: %s", project.project_id, project.status, project.deadline)
+            if "Available" in project.current_status:
+                scheduler.app.logger.debug("Project: %s has status %s - Expires: %s", project.title, project.current_status, project.current_deadline)
+                # scheduler.app.logger.debug(project.id)
+                # scheduler.app.logger.debug(project.current_deadline)
+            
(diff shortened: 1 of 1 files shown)


dds_web/scheduled_tasks.py
```diff
@@ -1,24 +1,56 @@
...
+                            + (
+                                ": Database malfunction."
+                                if isinstance(err, sqlalchemy.exc.OperationalError)
+                                else ": Server Error."
+                            )
+                        ),
+                    ) from err
+                scheduler.app.logger.debug("Project: %s has status Archived now!", project.title)
+            else:
+                scheduler.app.logger.debug("Nothing to do for Project: %s", project.title)
```
inaod568, line 56: Commiting once per project is fine but then you shouldn't raise the exception and instead save a list of the different errors that occurr and then log them after the scheduled task is done. So for example I would add a dict before the for loop(s) `errors = {}` and then in the unit loop `errors[unit.name] = {}`, and then in this except statement I'd do
```
errors[unit.name][project.public_id] = {"message": str(err), alt_message=("Status was not updated" + etc....)}
continue
```
And then after both for loops are finished 
```
for unit, projects in errors.items():
    scheduler.app.logger.error(f"Unit '{unit}' projects encountered issues during expiration process:")
    for project in projects:
        # log the project errors ("message" probably) for each project and note which unit it belongs to
        # or 
```

inaod568 (changes requested):

dds_web/scheduled_tasks.py
valyo: What I noticed from testing with DB down is that the apscheduler is already catching the error when trying to access its own table apscheduler_jobs and raising an error 
`dds_backend     | Error getting due jobs from job store 'default': (pymysql.err.OperationalError) (2003, "Can't connect to MySQL server on 'db' ([Errno -2] Name or service not known)")`

This basically means that the jobs will not be triggered and there won't be errors coming from them. On the other hand DB can go down (or connection can get lost) after the jobs are triggered, which means that we still want to catch and log them.
I am thinking correctly here!

inaod568: I'm not entirely sure I'm following, but do you mean that if the commit for one project fails and therefore raises an exception, the for loop will still continue with the rest of the projects? This is what I meant with my comment, sorry if that wasn't clear. If there is a SQLAlchemyError when commiting one project status change, it could be that there's a problem with exactly that project and none of the others, so then the project status change should continue in the other ones.

valyo: the changes are pushed now, not sure if I've done the error handling properly but let's see

dds_web/scheduled_tasks.py
```diff
@@ -1,24 +1,74 @@
 import flask_apscheduler
 import flask
 
-# Apscheduler
+## Apscheduler
 scheduler = flask_apscheduler.APScheduler()
 
 
+# @scheduler.task("cron", id="available_to_expired", minute=5, hour=2, misfire_grace_time=1)
```
inaod568, line 8: I think we should plan the cron jobs for just after midnight. Multiple cronjobs should be running concurrently. We may need to tweak the queries a bit to block multiple queries happening at the same time for the same row.
```suggestion
# @scheduler.task("cron", id="available_to_expired", hour=0, minute=1, misfire_grace_time=3600)
```

dds_web/scheduled_tasks.py
```diff
@@ -1,24 +1,74 @@
...
+            errors[unit.name] = {}
+
+            days_in_expired = unit.days_in_expired
+
+            for project in page_query(
+                models.Project.query.filter(
+                    sqlalchemy.and_(
+                        models.Project.is_active == 1, models.Project.unit_id == unit.id
+                    )
+                )
```
inaod568, line 35: Same thing here with the blocking
```suggestion
                db.session.query(models.Project).filter(
                    sqlalchemy.and(
                        models.Project.is_active == 1, models.Project.unit_id == unit.id
                    )
                ).with_for_update()
```

dds_web/scheduled_tasks.py
```diff
@@ -1,24 +1,74 @@
...
+            ):
+
+                if (
+                    project.current_status == "Available"
+                    and project.current_deadline >= current_time()
```
inaod568, line 40:
```suggestion
                    and project.current_deadline <= current_time()
```

dds_web/scheduled_tasks.py
```diff
@@ -1,24 +1,74 @@
...
 
     with scheduler.app.app_context():
-        # scheduler.app.logger.debug("This means that the app context works!")
-        test = models.Project.query.all()
-        # print(test, flush=True)
+        expire = ProjectStatus()
+
+        errors = {}
+
+        for unit in models.Unit.query.all():
```
inaod568, line 25: We also need to add a try-except around this for loop so that we can log the error
```
try:
    for unit.... etc and do stuff
except (both operational and sql):
    log the error 
    rollback 
    "raise"
```

dds_web/scheduled_tasks.py
valyo: 👍

dds_web/scheduled_tasks.py
inaod568: We can improve on it later maybe and have it as a decorator or something but just add this now so that we can catch potential query-issues

inaod568 (changes requested):

inaod568 (approved):
Looks good!

PR #1147, opened 2022-04-08
Delete Invite

### LINKED ISSUE #1137: Cronjob: Clean up invites which are not answered after a week
Go to the `scheduled_tasks.py` module and follow the structure in the other cron jobs.

1. Loop through the invites
2. Check when it was issued
3. Remove those entries that are older than 7 days
(Double check that this also removes the corresponding rows in the ProjectInviteKeys table)

zishanmirza:
This patch deletes invites that are older than a week.

Fixes #1137.

- [x] Tests passing
- [X] Black formatting
- [X] Migrations for any changes to the database schema
- [x] Note in the CHANGELOG

<a href="https://gitpod.io/#https://github.com/ScilifelabDataCentre/dds_web/pull/1147"></a>


dds_web/scheduled_tasks.py
```diff
@@ -167,3 +169,32 @@ def set_expired_to_archived():
...
+
+    with scheduler.app.app_context():
+        expiration: datetime.datetime = current_time() - timedelta(weeks=1)
+
+        try:
+            invites: list = db.session.query(models.Invite).all()
+            for invite in invites:
+                if invite.created_at < expiration:
+                    models.Invite.query.filter_by(id=invite.id).delete()
+                    models.ProjectInviteKeys.query.filter_by(invite_id=invite.id).delete()
```
inaod568, line 194: The db constraints actually handle the `ProjectInviteKeys` so all you need to do is this and both the invites and the rows in the `ProjectInviteKeys` connected to that invite should be deleted.
```suggestion
                    db.session.delete(invite)
```

inaod568 (changes requested):
Looks good, a few comments though. 
Apart from this, could you also keep track of the non-deleted invites in a similar way as the other scheduled tasks? So have a dict that you put the invite errors in and afterward print them out. If there's an issue with one of the invite deletions it should still be possible to delete the others. 

So there should be 2 try-except blocks, one outside of the for loop and one per commit.

zishanmirza:
I have updated the `if` statement, added a `try` statement, and included `errors`.

zishanmirza:
I have updated the `if` statement, added a `try` statement, and included `errors`.

zishanmirza:
I have updated the `if` statement, added a `try` statement, and included `errors`.

inaod568 (changes requested):

zishanmirza:
I have updated `expiration`.

zishanmirza:
I have updated `expiration`.

dds_web/scheduled_tasks.py
```diff
@@ -167,3 +169,39 @@ def set_expired_to_archived():
...
+                        error[invite.id] = str(invite)
+                        scheduler.app.logger.exception(err)
+                        db.session.rollback()
+                        raise
+        except (OperationalError, SQLAlchemyError) as err:
+            scheduler.app.logger.exception(err)
+            raise
+
+        for error in errors.items():
+            scheduler.app.logger.error(f"{errors[error]} not deleted.")
```
inaod568, line 207:
```suggestion
        for inv, error in errors.items():
            scheduler.app.logger.error(f"{inv} not deleted: {error}")
```

inaod568 (changes requested):
There was one tiny typo and also we need to save the error message and print that out. Also shouldn't raise in the inner try except, otherwise it will not continue with the next invite.

inaod568 (approved):
Looks good!

PR #1159, opened 2022-04-19
Rename unit.py

zishanmirza:
This patch renames "unit.py" to "superadmin_only.py".

Fixes DDS-1165.

- [X] Tests passing
- [X] Black formatting
- [x] Note in the CHANGELOG

<a href="https://gitpod.io/#https://github.com/ScilifelabDataCentre/dds_web/pull/1159"></a>


inaod568 (changes requested):
I think this was a good start and I like the idea of keeping is simple but I think might cause some confusion. Also, I see now that the `superadmin_resources` is not great. 

I think we should do the following changes: 
1. `superadmin_resources.py` --> `superadmin_only.py` 
2. Remove the `as unit` part and change the occurrences of `unit`

dds_web/api/__init__.py
```diff
@@ -72,8 +72,8 @@ def output_json(data, code, headers=None):
 
 # Units #################################################################################### Units #
```
inaod568, line 73:
```suggestion
# Super Admins ###################################################################### Super Admins #
```

inaod568 (approved):

PR #1163, opened 2022-05-06
partial fix for delete_invite error DDS-1264 

valyo:
Fixes DDS-1264 partially; we need to make sure that we avoid invites with strange created_at date, and to handle informing the user, perhaps. 

- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG

<a href="https://gitpod.io/#https://github.com/ScilifelabDataCentre/dds_web/pull/1163"></a>

### DIFF
--- a/dds_web/scheduled_tasks.py
+++ b/dds_web/scheduled_tasks.py
@@ -190,9 +190,7 @@ def delete_invite():
         try:
             invites: list = db.session.query(models.Invite).all()
             for invite in invites:
-                invalid_invite = False
-                if invite.created_at == "0000-00-00 00:00:00":
-                    invalid_invite = True
+                invalid_invite = invite.created_at == "0000-00-00 00:00:00"
                 if invalid_invite or (invite.created_at + timedelta(weeks=1)) < expiration:
                     try:
                         db.session.delete(invite)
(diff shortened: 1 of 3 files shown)


inaod568:
@valyo Could you add a little note in the PR description on why it's temporary?

dds_web/scheduled_tasks.py
```diff
@@ -190,11 +190,19 @@ def delete_invite():
         try:
             invites: list = db.session.query(models.Invite).all()
             for invite in invites:
-                if (invite.created_at + timedelta(weeks=1)) < expiration:
+                invalid_invite = False
+                if invite.created_at == "0000-00-00 00:00:00":
+                    invalid_invite = True
```
inaod568, line 195: This could also be replaced with simply
```suggestion
                invalid_invite = invite.created_at == "0000-00-00 00:00:00"
```

inaod568 (commented):
Small suggestion. Everything looks good though and it works.
Let me know if you commit the suggestion or if you want to keep it as it is now.

dds_web/scheduled_tasks.py
valyo: so this sets invalid_invite to True if equality check is true, and to False otherwise, nice 👍

inaod568 (approved):
Looks good!

PR #1187, opened 2022-05-23
Version Header

zishanmirza:
This patch validates the version header.

Fixes DDS-1260 and DDS-1261.

# Description

Please include the following in this section

- [x] Summary of the changes and the related issue
- [ ] Relevant motivation and context
- [ ] Any dependencies that are required for this change

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Checklist:

Please delete options that are not relevant.

- [X] Rebase/merge the branch which this PR is made to

## Formatting and documentation

- [ ] I have added a row in the [changelog](../CHANGELOG.md)
- [X] The code follows the style guidelines of this project: Black / Prettier formatting
- [X] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings

## Tests

- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

<a href="https://gitpod.io/#https://github.com/ScilifelabDataCentre/dds_web/pull/1187"></a>


dds_web/__init__.py
```diff
@@ -185,6 +185,20 @@ def create_app(testing=False, database_uri=None):
             # Simplifies testing as we don't test the session protection anyway
             login_manager.session_protection = "basic"
 
+        @app.after_request
```
inaod568, line 188: We want to check the version before the request and this checks it after the request has been made and the API is returning a response.

inaod568: So in this case something could be executed even though it shouldn't be because the version is incorrect

inaod568 (approved):

dds_web/__init__.py
inaod568, line 188: We want to check the version before any code is executed, so we should change this. Otherwise the request is made, something is executed and then when the response is being returned the version is checked

PR #1190, opened 2022-05-30
Save usage per project in db (new table added)

valyo:
New table "usage" for monthly records of the projects usage.
Added functionality to the cron job monthly_usage for making the db records. 

Please include the following in this section

- [x] Summary of the changes and the related issue
- [ ] Relevant motivation and context
- [ ] Any dependencies that are required for this change

Fixes # DDS-1288

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Checklist:

Please delete options that are not relevant.

- [ ] Any dependent changes have been merged and published in downstream modules
- [x] Rebase/merge the branch which this PR is made to
- [ ] Changes to the database schema: A new migration is included in the PR

## Formatting and documentation

- [ ] I have added a row in the [changelog](../CHANGELOG.md)
- [x] The code follows the style guidelines of this project: Black / Prettier formatting
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings

## Tests

- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

<a href="https://gitpod.io/#https://github.com/ScilifelabDataCentre/dds_web/pull/1190"></a>


dds_web/scheduled_tasks.py
```diff
@@ -242,8 +242,42 @@ def monthly_usage():
             scheduler.app.logger.info(usage)
 
         scheduler.app.logger.debug("Task: Projects usage from database")
+        # a mock for a unit with two projects with some usage in bhours
+        usage_data_mock = {
```
inaod568, line 246: Seems like this is unused - is this intentional?

dds_web/scheduled_tasks.py
```diff
@@ -242,8 +242,42 @@ def monthly_usage():
...
+                "Unit cost": 342,
+            },
+        }
+
+        # go through units and projects in the database and
+        # construct a usage and cost data dictionary
+        usage_data = {}
         try:
             for unit in db.session.query(models.Unit).with_for_update().all():
+                usage_data[unit.safespring_name] = {}
```
inaod568, line 278: Maybe we can remove the for loop above now and use that code in this for loop instead?
```suggestion
                # Get info from safespring / sunet (replace with e.g. api call later)
                safespring_data[unit.safespring_name] = {
                    "TotalBytes": 434595434499,
                    "TotalBytesRounded": 1434614451200,
                    "TotalEntries": 10333,
                }
                usage = f"Total usage in cloud for unit {unit.name} ({unit.safespring_name}): {safespring_data[unit.safespring_name]['TotalBytes']}"
                scheduler.app.logger.info(usage)

                # Calculate usage per project and total in db
                usage_data[unit.safespring_name] = {}
```

dds_web/scheduled_tasks.py
```diff
@@ -254,14 +288,41 @@ def monthly_usage():
                     )
                     .with_for_update()
                 ):
-                    proj_bhours, proj_cost = UserProjects.project_usage(project)
+                    usage_data[unit.safespring_name][project.id] = {}
+                    proj_bhours, proj_cost = UserProjects.project_usage(project=project)
```
inaod568, line 292:
```suggestion

                    # Calculate project usage
                    proj_bhours, proj_cost = UserProjects.project_usage(project=project)
```

dds_web/scheduled_tasks.py
```diff
@@ -254,14 +288,41 @@ def monthly_usage():
...
                     scheduler.app.logger.info(
                         "Current total usage for project %s is %s bhours, and total cost is %s kr",
-                        project.public_id,
+                        project.id,
                         proj_bhours,
                         proj_cost,
                     )
+                    usage_data[unit.safespring_name][project.id]["usage"] = proj_bhours
```
inaod568, line 299:
```suggestion

                    # Save project usage 
                    usage_data[unit.safespring_name][project.id]["usage"] = proj_bhours
```

dds_web/scheduled_tasks.py
```diff
@@ -254,14 +288,41 @@ def monthly_usage():
...
+                    usage_data[unit.safespring_name][project.id]["cost"] = proj_cost
+                    unit_usage += proj_bhours
```
inaod568, line 301:
```suggestion

                    # Increase total usage
                    unit_usage += proj_bhours
```

dds_web/scheduled_tasks.py
```diff
@@ -254,14 +288,41 @@ def monthly_usage():
...
+                    unit_cost += proj_cost
+                usage_data[unit.safespring_name]["Unit usage"] = unit_usage
```
inaod568, line 303:
```suggestion

                # Save total usage
                usage_data[unit.safespring_name]["Unit usage"] = unit_usage
```

dds_web/scheduled_tasks.py
```diff
@@ -254,14 +288,41 @@ def monthly_usage():
...
+            if data["Unit usage"] != 0:
+                scheduler.app.logger.info(f'Total usage for unit {unit} is: {data["Unit usage"]}')
+                for project, pdata in data.items():
+                    if isinstance(project, int):
+                        new_record = models.Usage(
+                            project_id=project,
+                            usage=pdata["usage"],
+                            cost=pdata["cost"],
+                            time_collected=current_time(),
+                        )
```
inaod568, line 321: I'm thinking we could move this to the previous for loop and just create the usage row after the calculation of the usage? Instead of doing another iteration through the items and saving to a dict

dds_web/scheduled_tasks.py
valyo: it is intended for easier testing by only changing line 311

valyo:
All the suggestions are good, but since I am working with the same code in the next task, we either have to make those changes here and merge, or I can implement them in the next one directly and then we can discard this PR(we mentioned this during the morning meeting). 
The second option will be less interaction/work for you, I think.

dds_web/scheduled_tasks.py
inaod568: Yeah, I agree

dds_web/scheduled_tasks.py
inaod568: Ah ok, fair enough. Could you comment it out and add a comment about this too? There's a risk that someone at some point removes it otherwise because the variable is unused.

dds_web/scheduled_tasks.py
```diff
@@ -225,7 +225,7 @@ def monthly_usage():
     from dds_web import db
     from dds_web.database import models
     from dds_web.api.project import UserProjects
-    from dds_web.utils import page_query
+    from dds_web.utils import current_time, page_query
 
     with scheduler.app.app_context():
         # a mock dict with data that should be obtained from Safesprig's API
```
inaod568, line 231: Leave the `safespring_data` variable as it is, but you can remove lines 233-242 since the code is present in the for loops below now. 
`unit.safespring_name` on line 247 also needs to change since it's outside of a loop and `unit` is undefined (not caught until the above is removed)

dds_web/scheduled_tasks.py
inaod568: Also, I think you can remove the `usage_data` parts now since you're saving the info to the db straight away anyway and increasing the total usage, so `usage_data` is not needed.

valyo: Yes, I was thinking about this already yesterday; it is used only for the percentages calculation, and this is something we can do by querying the DB in the billing job.

valyo:
I cleaned up all the unnecessary stuff so that this job does only what it is supposed to do.

inaod568:
Ok! I'll review again now 👍🏻

inaod568 (approved):
Looks good!

PR #1298, opened 2022-10-10
Update maintenance migration and change init-db command

inaod568:
https://github.com/ScilifelabDataCentre/dds_web/pull/1296

<a href="https://gitpod.io/#https://github.com/ScilifelabDataCentre/dds_web/pull/1298"></a>


inaod568:
@talavis It didn't work. Do I need to create a new migration? I thought I could just edit it but that doesn't seem to work. But having a migration with only adding a row to the maintenance table doesn't feel great.

PR #1345, opened 2023-01-03
Fix the path for dds-reporting.csv

valyo:
> **Before submitting the PR, please go through the sections below and fill in what you can. If there are any items that are irrelevant for the current PR, remove the row. If a relevant option is missing, please add it as an item and add a PR comment informing that the new option should be included into this template.**

> **All _relevant_ items should be ticked before the PR is merged**

Since the reporting cron job was failing to find the csv file, see https://scilifelab.zendesk.com/agent/tickets/2645, we change the path 

- [x] Summary of the changes and the related issue:
- [x] Motivation and context regarding why the change is needed:

## Type of change

- [x] Bug fix
  - [x] Non-breaking

_"Breaking": The change will cause existing functionality to not work as expected._

# Checklist:

## General

- [x] [Changelog](../CHANGELOG.md): New row added. Not needed when PR includes _only_ tests.
- [x] Code change
  - [x] Self-review of code done

## Checks

- [x] CodeQL passes
- [x] Formatting: Black & Prettier checks pass
- Tests
  - [x] The tests pass
- Trivy / Snyk:
  - [x] Security alerts have been dismissed: `setuptools` had a vulnerability, but the web does not use the cli in production, and setuptools is not included in the web outside of the cli

<a href="https://gitpod.io/#https://github.com/ScilifelabDataCentre/dds_web/pull/1345"></a>


i-oden (approved):
Looks good

PR #1480, opened 2023-10-10
Dds 1691 -- New API endpoint to extend the deadline: `ProjectStatus.patch`

rv0lt:

## 1. Description / Summary

Implemented the functionality in the API of extend the deadline for a project with a new endpoint:

The new endpoint is PATCH /proj/status 

Currently, it only supports the functionality of extending the deadline for which the following body is necessary:

```
{
"new_deadline_in": int
}

```
If, _new_deadline_in_  has not been provided then, nothing is executed. If it is provided, it needs to be an integer that defines the extension in days for the new deadline. These days while be added on top of the current day's left.

The workflow process that should be handled when **project status extend deadline** is called is as follows:

1.  CLI: call ProjectStatus.patch
2. API: ProjectStatus.patch returns project status and deadline (days left)
3. CLI: Print _project status, deadline_, and chosen _new deadline_. Ask for confirmation.
4. CLI: call ProjectStatus.patch with confirmed flag
5. API: ProjectStatus.patch verifies correct status, and ok deadline, and extends the deadline.
    
    
    
    

[_Link to the github issue or add the Jira task ID here._](https://scilifelab.atlassian.net/jira/software/projects/DDS/boards/13?isInsightsOpen=true&selectedIssue=DDS-1691)

## 3. Type of change

- [x] New feature
  - [x] Non-breaking 
- [ ] Bug fix 
- [ ] Security Alert fix 
- [ ] Documentation 
- [ ] Workflow 
- [ ] Tests **only** 

## 4. Additional information

- [ ] Blocking PRs 
  - [ ] Merged 

## 5. Actions / Scans

- [x] **Black**

- [x] **Prettier**

- [x] **Yamllint**

- [x] **Tests**

- [x] **CodeQL**

- [x] **Trivy**

- [x] **Snyk**

### DIFF
--- a/SPRINTLOG.md
+++ b/SPRINTLOG.md
@@ -314,3 +314,4 @@ _Nothing merged in CLI during this sprint_
 - Use full DDS name in MOTD email subject ([#1477](https://github.com/ScilifelabDataCentre/dds_web/pull/1477))
 - Add flag --verify-checksum to the comand in email template ([#1478])(https://github.com/ScilifelabDataCentre/dds_web/pull/1478)
 - Improved email layout; Highlighted information and commands when project is released ([#1479])(https://github.com/ScilifelabDataCentre/dds_web/pull/1479)
+- Added new API endpoint ProjectStatus.patch to extend the deadline ([#1480])(https://github.com/ScilifelabDataCentre/dds_web/pull/1480)
--- a/SPRINTLOG.md
+++ b/SPRINTLOG.md
@@ -314,4 +314,7 @@ _Nothing merged in CLI during this sprint_
 - Use full DDS name in MOTD email subject ([#1477](https://github.com/ScilifelabDataCentre/dds_web/pull/1477))
 - Add flag --verify-checksum to the comand in email template ([#1478])(https://github.com/ScilifelabDataCentre/dds_web/pull/1478)
 - Improved email layout; Highlighted information and commands when project is released ([#1479])(https://github.com/ScilifelabDataCentre/dds_web/pull/1479)
+
+# 2023-10-16 - 2023-10-27
+
 - Added new API endpoint ProjectStatus.patch to extend the deadline ([#1480])(https://github.com/ScilifelabDataCentre/dds_web/pull/1480)
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -271,7 +271,7 @@ class ProjectStatus(flask_restful.Resource):
                 # wont change again -> error
                 if project.times_expired >= 2:
                     raise DDSArgumentError(
-                        "Project availability limit: Project cannot be made Available any more times"
+                        "Project availability limit: The maximun number of changes in data availability has been reached."
                     )
                 try:
                     # add a fake expire status to mimick a re-release in order to have an udpated deadline
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -188,6 +188,102 @@ class ProjectStatus(flask_restful.Resource):
 
         return {"message": return_message}
 
+    @auth.login_required(role=["Unit Admin", "Unit Personnel"])
+    @logging_bind_request
+    @json_required
+    @handle_validation_errors
+    def patch(self):
+        """Partially update a the project status"""
+        # Get project ID, project and verify access
+        project_id = dds_web.utils.get_required_item(obj=flask.request.args, req="project")
+        project = dds_web.utils.collect_project(project_id=project_id)
+        dds_web.utils.verify_project_access(project=project)
+
+        # Cannot change project status if project is busy
+        if project.busy:
+            raise ProjectBusyError(
+                message=(
+                    f"The status for the project '{project_id}' is already in the process of being changed. "
+                    "Please try again later. \n\nIf you know the project is not busy, contact support."
+                )
+            )
+
+        self.set_busy(project=project, busy=True)
+
+        try:
+            # get atributes
+            json_input = flask.request.get_json(silent=True)  # Already checked by json_required
+            extend_deadline = json_input.get("extend_deadline", False)  # False by default
+
+            # some variable definition
+            curr_date = dds_web.utils.current_time()
+            send_email = False
+
+            if extend_deadline:
+                # deadline can only be extended from Available
+                if not project.current_status == "Available":
+                    raise DDSArgumentError("Can only extend deadline if the project is available")
+
+                new_deadline_in = json_input.get("new_deadline_in")
+                if not new_deadline_in:
+                    raise DDSArgumentError(
+                        message="No new deadline provived, cannot perform operation."
+                    )
+                if new_deadline_in > 90:
+                    raise DDSArgumentError(
+                        message="The deadline needs to be less than (or equal to) 90 days."
+                    )
+
+                if project.times_expired > 2:
+                    raise DDSArgumentError(
+                        "Project availability limit: Project cannot have an extended deadline any more times"
+                    )
+
+                try:
+                    # add a fake archived status to mimick a re-release in order to have an udpated deadline
+                    new_status_row = self.expire_project(
+                        project=project,
+                        current_time=curr_date,
+                        deadline_in=project.responsible_unit.days_in_expired,
+                    )
+                    project.project_statuses.append(new_status_row)
+                    db.session.commit()
+
+                    new_status_row = self.release_project(
+                        project=project, current_time=curr_date, deadline_in=new_deadline_in
+                    )
+                    project.project_statuses.append(new_status_row)
+
+                    project.busy = False  # return to not busy
+                    db.session.commit()
+
+                except (sqlalchemy.exc.OperationalError, sqlalchemy.exc.SQLAlchemyError) as err:
+                    flask.current_app.logger.exception(err)
+                    db.session.rollback()
+                    raise DatabaseError(
+                        message=str(err),
+                        alt_message=(
+                            "Status was not updated"
+                            + (
+                                ": Database malfunction."
+                                if isinstance(err, sqlalchemy.exc.OperationalError)
+                                else ": Server Error."
+                            )
+                        ),
+                    ) from err
+
+                return_message = f"{project.public_id} has been given a new deadline"
+
+                return_message += (
+                    f". An e-mail notification has{' not ' if not send_email else ' '}been sent."
+                )
+
+        except:
+            self.set_busy(project=project, busy=False)
+            raise
+
+        return {"message": return_message}
+
     @staticmethod
     @dbsession
     def set_busy(project: models.Project, busy: bool) -> None:
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -203,6 +203,13 @@ class ProjectStatus(flask_restful.Resource):
         # Get json input from request
         json_input = flask.request.get_json(silent=True)  # Already checked by json_required
 
+        # the status has changed at least two times,
+        # next time the project expires it wont change again -> error
+        if project.times_expired >= 2:
+            raise DDSArgumentError(
+                "Project availability limit: The maximun number of changes in data availability has been reached."
+            )
+
         # Operation must be confirmed by the user - False by default
         confirmed_operation = json_input.get("confirmed", False)
         if not isinstance(confirmed_operation, bool):
@@ -266,12 +273,6 @@ class ProjectStatus(flask_restful.Resource):
                 if new_deadline_in + current_deadline > 90:
                     raise DDSArgumentError(
                         message=f"You requested the deadline to be extended with {new_deadline_in} days (from {current_deadline}), giving a new total deadline of {new_deadline_in + current_deadline} days. The new deadline needs to be less than (or equal to) 90 days."
-                    )
-                # the dealine has changed at least two times, next time it expires
-                # wont change again -> error
-                if project.times_expired >= 2:
-                    raise DDSArgumentError(
-                        "Project availability limit: The maximun number of changes in data availability has been reached."
                     )
                 try:
                     # add a fake expire status to mimick a re-release in order to have an udpated deadline
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -247,7 +247,7 @@ class ProjectStatus(flask_restful.Resource):
             # some variable definition
             curr_date = dds_web.utils.current_time()
             send_email = False
-            default_unit_days = project.responsible_unit.days_in_expired
+            default_unit_days = project.responsible_unit.days_in_available
 
             # Update the deadline functionality
             if new_deadline_in:
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -228,10 +228,10 @@ class ProjectStatus(flask_restful.Resource):
                 )
             )
 
-        self.set_busy(project=project, busy=True)
-
         # Extend deadline
         try:
+            self.set_busy(project=project, busy=True)
+
             new_deadline_in = json_input.get(
                 "new_deadline_in", None
             )  # if not provided --> is None -> deadline is not updated
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -271,7 +271,7 @@ class ProjectStatus(flask_restful.Resource):
                 except (sqlalchemy.exc.OperationalError, sqlalchemy.exc.SQLAlchemyError) as err:
                     flask.current_app.logger.exception("Failed to extend deadline")
                     db.session.rollback()
-                    raise err
+                    raise
 
                 return_message = f"{project.public_id} has been given a new deadline"
 
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -209,6 +209,7 @@ class ProjectStatus(flask_restful.Resource):
             raise DDSArgumentError(message="`confirmed` is a boolean value: True or False.")
         if not confirmed_operation:
             warning_message = "Operation must be confirmed before proceding."
+            # When not confirmed, return information about the project
             project_info = ProjectInfo().get()
             project_status = self.get()
             json_returned = {
@@ -228,10 +229,10 @@ class ProjectStatus(flask_restful.Resource):
                 )
             )
 
+        self.set_busy(project=project, busy=True)
+
         # Extend deadline
         try:
-            self.set_busy(project=project, busy=True)
-
             new_deadline_in = json_input.get(
                 "new_deadline_in", None
             )  # if not provided --> is None -> deadline is not updated
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -208,7 +208,15 @@ class ProjectStatus(flask_restful.Resource):
             raise DDSArgumentError(message="`confirmed` is a boolean value: True or False.")
         if not confirmed_operation:
             warning_message = "Operation must be confirmed before proceding."
-            return {"warning": warning_message}
+            project_info = ProjectInfo().get()
+            project_status = self.get()
+            json_returned = {
+                **project_info,
+                "project_status": project_status,
+                "warning": warning_message,
+                "default_unit_days": project.responsible_unit.days_in_expired,
+            }
+            return json_returned
 
         # Cannot change project status if project is busy
         if project.busy:
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -225,13 +225,19 @@ class ProjectStatus(flask_restful.Resource):
                     raise DDSArgumentError("Can only extend deadline if the project is available")
 
                 new_deadline_in = json_input.get("new_deadline_in")
+                current_deadline = (project.current_deadline - curr_date).days
+
                 if not new_deadline_in:
                     raise DDSArgumentError(
                         message="No new deadline provived, cannot perform operation."
                     )
-                if new_deadline_in > 90:
+                if current_deadline > 10:
                     raise DDSArgumentError(
-                        message="The deadline needs to be less than (or equal to) 90 days."
+                        message=f"There are still {current_deadline} days left, it is not possible to extend deadline yet"
+                    )
+                if new_deadline_in + current_deadline > 90:
+                    raise DDSArgumentError(
+                        message="The new deadline needs to be less than (or equal to) 90 days."
                     )
 
                 if project.times_expired > 2:
@@ -250,7 +256,9 @@ class ProjectStatus(flask_restful.Resource):
                     db.session.commit()
 
                     new_status_row = self.release_project(
-                        project=project, current_time=curr_date, deadline_in=new_deadline_in
+                        project=project,
+                        current_time=curr_date,
+                        deadline_in=new_deadline_in + current_deadline,
                     )
                     project.project_statuses.append(new_status_row)
 
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -233,7 +233,7 @@ class ProjectStatus(flask_restful.Resource):
                 # deadline can only be extended from Available
                 if not project.current_status == "Available":
                     raise DDSArgumentError(
-                        "you can only extend the deadline for a project that has the status Available."
+                        "You can only extend the deadline for a project that has the status 'Available'."
                     )
 
                 new_deadline_in = json_input.get("new_deadline_in")
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -273,7 +273,7 @@ class ProjectStatus(flask_restful.Resource):
                     db.session.commit()
 
                 except (sqlalchemy.exc.OperationalError, sqlalchemy.exc.SQLAlchemyError) as err:
-                    flask.current_app.logger.exception(err)
+                    flask.current_app.logger.exception("Failed to extend deadline")
                     db.session.rollback()
                     raise err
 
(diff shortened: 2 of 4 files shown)


dds_web/api/project.py
```diff
@@ -188,6 +188,107 @@ def post(self):
...
+            # some variable definition
+            curr_date = dds_web.utils.current_time()
+            send_email = False
+
+            # Update the deadline
+            if extend_deadline:
+                # deadline can only be extended from Available
+                if not project.current_status == "Available":
+                    raise DDSArgumentError(
+                        "you can only extend the deadline for a project that has the status Available."
```
i-oden, line 236:
```suggestion
                        "You can only extend the deadline for a project that has the status 'Available'."
```

dds_web/api/project.py
```diff
@@ -188,6 +188,107 @@ def post(self):
...
+                current_deadline = (project.current_deadline - curr_date).days
+
+                if not new_deadline_in:
+                    raise DDSArgumentError(
+                        message="No new deadline provived, cannot perform operation."
+                    )
+                if current_deadline > 10:
+                    raise DDSArgumentError(
+                        message=f"There are still {current_deadline} days left, it is not possible to extend deadline yet."
+                    )
```
i-oden, line 249: They should still be able to extend the deadline, as long as they confirm it in the CLI. There's a risk that e.g. the unit doesn't have time to extend within those days. 
```suggestion
```

dds_web/api/project.py
```diff
@@ -188,6 +188,107 @@ def post(self):
...
+
+                try:
+                    # add a fake archived status to mimick a re-release in order to have an udpated deadline
+                    new_status_row = self.expire_project(
+                        project=project,
+                        current_time=curr_date,
+                        deadline_in=project.responsible_unit.days_in_expired,
+                    )
+                    project.project_statuses.append(new_status_row)
+                    db.session.commit()
```
i-oden, line 263: Do we need this first commit?

dds_web/api/project.py
```diff
@@ -188,6 +188,107 @@ def post(self):
...
+                    )
+                    project.project_statuses.append(new_status_row)
+
+                    project.busy = False  # return to not busy
+                    db.session.commit()
+
+                except (sqlalchemy.exc.OperationalError, sqlalchemy.exc.SQLAlchemyError) as err:
+                    flask.current_app.logger.exception("Failed to extend deadline")
+                    db.session.rollback()
+                    raise err
```
i-oden, line 278: Should just need `raise` 
```suggestion
                    raise
```

dds_web/api/project.py
```diff
@@ -188,6 +188,107 @@ def post(self):
...
+    @json_required
+    @handle_validation_errors
+    def patch(self):
+        """Partially update a the project status"""
+        # Get project ID, project and verify access
+        project_id = dds_web.utils.get_required_item(obj=flask.request.args, req="project")
+        project = dds_web.utils.collect_project(project_id=project_id)
+        dds_web.utils.verify_project_access(project=project)
+
+        # get atributes
```
i-oden, line 202:
```suggestion
        # Get json input from request
```

dds_web/api/project.py
```diff
@@ -188,6 +188,107 @@ def post(self):
...
+        json_input = flask.request.get_json(silent=True)  # Already checked by json_required
+
+        # false by default - operation must be confirmed by the user
```
i-oden, line 205:
```suggestion
        # Operation must be confirmed by the user - False by default
```

dds_web/api/project.py
```diff
@@ -188,6 +188,107 @@ def post(self):
...
+            raise ProjectBusyError(
+                message=(
+                    f"The deadline for the project '{project_id}' is already in the process of being changed. "
+                    "Please try again later. \n\nIf you know the project is not busy, contact support."
+                )
+            )
+
+        self.set_busy(project=project, busy=True)
+
+        try:
```
i-oden, line 224:
```suggestion
        # Extend deadline
        try:
```

dds_web/api/project.py
```diff
@@ -188,6 +188,107 @@ def post(self):
...
+            extend_deadline = json_input.get("extend_deadline", False)  # False by default
```
i-oden, line 225: This is a good check to have if the endpoint had more options, but since we're only allowing extending the deadline for now, it's better to leave this out and only add it if we need it in the future, to keep the code as simple as possible.

dds_web/api/project.py
```diff
@@ -188,6 +188,107 @@ def post(self):
...
+
+            # some variable definition
+            curr_date = dds_web.utils.current_time()
+            send_email = False
+
+            # Update the deadline
+            if extend_deadline:
```
i-oden, line 232: Connected to previous commend about `extend_deadline`

dds_web/api/project.py
```diff
@@ -188,6 +188,107 @@ def post(self):
...
+                        "you can only extend the deadline for a project that has the status Available."
+                    )
+
+                new_deadline_in = json_input.get("new_deadline_in")
+                current_deadline = (project.current_deadline - curr_date).days
+
+                if not new_deadline_in:
+                    raise DDSArgumentError(
+                        message="No new deadline provived, cannot perform operation."
+                    )
```
i-oden, line 245: If the user hasn't specified a deadline - use the `Unit.days_in_available` variable here

dds_web/api/project.py
```diff
@@ -188,6 +188,107 @@ def post(self):
...
+                if current_deadline > 10:
+                    raise DDSArgumentError(
+                        message=f"There are still {current_deadline} days left, it is not possible to extend deadline yet."
+                    )
+                if new_deadline_in + current_deadline > 90:
+                    raise DDSArgumentError(
+                        message="The new deadline needs to be less than (or equal to) 90 days."
+                    )
```
i-oden, line 253: Maybe add a bit of information, something like this? 
```suggestion
                    raise DDSArgumentError(
                        message=f"You requested the deadline to be extended with {new_deadline_in} days (from {current_deadline}), giving a new total deadline of {new_deadline_in + current_deadline} days. The new deadline needs to be less than (or equal to) 90 days."
                    )
```

i-oden (changes requested):
Am I understanding it correctly if I say you're aiming at this being the full flow for the CLI and API? 
1. CLI: call `ProjectStatus.get`
2. API: `ProjectStatus.get` returns project status and deadline
3. CLI: Print project status, deadline, and chosen new deadline. Ask for confirmation.
4. CLI: call `ProjectStatus.patch`
5. API: `ProjectStatus.patch` verifies correct status, and ok deadline, and extends the deadline. 

If so, also add this to the PR description so that it's easier to work on the CLI command part. If not add the correct imagined steps to the PR description 🙏🏻

dds_web/api/project.py
rv0lt: Oka, I will make then the check with:

```
            new_deadline_in = json_input.get(
                "new_deadline_in", None
            )  # if not provided is None -> deadlie is not updated
```

Then, we just need to add the atributes to update in the future if the function is expanded

i-oden (changes requested):
Change the title of the PR, e.g. match it to the sprintlog.

Use the `Unit.days_in_available` if the new deadline is not specified 
https://github.com/ScilifelabDataCentre/dds_web/pull/1480#discussion_r1357778177

dds_web/api/project.py
```diff
@@ -188,6 +188,107 @@ def post(self):
...
+                )
+            )
+
+        self.set_busy(project=project, busy=True)
+
+        # Extend deadline
+        try:
+            new_deadline_in = json_input.get(
+                "new_deadline_in", None
+            )  # if not provided is None -> deadlie is not updated
```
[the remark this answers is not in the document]
i-oden: This also means you can remove the `if new_deadline_in` and `else: return_message="Nothing to update"`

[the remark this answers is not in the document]
i-oden: So the idea is that the unit user shouldn't need to specify the `deadline` in the CLI command, and that if they don't specify it it goes to the default -- `Unit.days_in_available`. But it would be good to display that to the users when checking for confirmation. "You're attempting to extend the deadline with x days". The problem is atm though that `ProjectStatus.get` doesn't return that.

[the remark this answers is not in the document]
i-oden: So it seems that either we 
1. Keep it like it is now: `ProjectStatus.get` first and then `ProjectStatus.patch` to do the deadline extension, and in this case we force the user to specify the `deadline` in the CLI. 
2. Change the `ProjectStatus.patch` endpoint to check the project status, return the project status + what they're attempting to extend the deadline with, and then the CLI calls `ProjectStatus.patch` a second time which (if `confirmed=True` extends the deadline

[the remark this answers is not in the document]
rv0lt: But if in the future the function is extended to update only other parameters, then it will always fetch a new_deadline attribute, making it difficult to filter out.

Also, the default _days_in_available_ is 90. And the function is checking for _new_deadline_in + current_deadline > 90:_
This check will always fail when using new_deadline = 90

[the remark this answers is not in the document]
i-oden: Let me know if I'm not explaining it clearly etc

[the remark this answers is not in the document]
rv0lt: Yeah, the comment was written before refreshing to see the full feedback. But my point stands with the option 1. Because of what I mentioned. 

If we don't want to "force" them to write it in the command. Then it can be asked as a terminal input when project status extend is invoked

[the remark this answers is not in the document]
i-oden: Discussed via zoom. Agreed upon: 

```python
        if not confirmed_operation:
            warning_message = "Operation must be confirmed before proceding."
            return {"warning": warning_message}
```

Should return 
1. current project status
2. The default number of days_in_available (project.responsible_unit.days_in_available) 
3. Possibly the message

dds_web/api/project.py
```diff
@@ -188,6 +188,115 @@ def post(self):
...
```
i-oden, line 236:
```suggestion
            )  # if not provided --> is None -> deadline is not updated
```

dds_web/api/project.py
```diff
@@ -188,6 +188,115 @@ def post(self):
 
         return {"message": return_message}
 
+    @auth.login_required(role=["Unit Admin", "Unit Personnel"])
+    @logging_bind_request
+    @json_required
+    @handle_validation_errors
+    def patch(self):
```
i-oden, line 195: We don't have this on the other `ProjectStatus` endpoints but we should add the `handle_db_error` decorator so that the endpoint returns a `DatabaseError` instead of pure SQLAlchemy errors etc in the case of a database issue. E.g. if there's an error in the lines leading up to the `try` block. 
```suggestion
    @auth.login_required(role=["Unit Admin", "Unit Personnel"])
    @logging_bind_request
    @json_required
    @handle_validation_errors
    @handle_db_error
    def patch(self):
```

dds_web/api/project.py
```diff
@@ -188,6 +188,115 @@ def post(self):
...
+        # Cannot change project status if project is busy
+        if project.busy:
+            raise ProjectBusyError(
+                message=(
+                    f"The deadline for the project '{project_id}' is already in the process of being changed. "
+                    "Please try again later. \n\nIf you know the project is not busy, contact support."
+                )
+            )
+
+        self.set_busy(project=project, busy=True)
```
i-oden, line 230: So since `set_busy` doesn't contain a commit we'll need to add a try except to this and the end where busy is set to False. This is not done in the other `ProjectStatus` endpoints either (noticed now), but if we need to change those too then that's another task. Unless we could add a test (or test manually) somehow to verify that that the `busy=True` is actually noticed in other database sessions running in parallel

i-oden (changes requested):

dds_web/api/project.py
rv0lt: You mean inserting the set busy inside the try catch below?

Like

```
try:
            self.set_busy(project=project, busy=True)
            logic_of_the_code()
except:
            self.set_busy(project=project, busy=False)
            raise
return {"message": return_message}
```

i-oden: No we need to make sure that the `self.set_busy` is commited first and also that the last change with `set_busy` is commited.
```python
try:
            self.set_busy(project=project, busy=True)
            db.session.commit()
            logic_of_the_code()
except:
            try: 
                 self.set_busy(project=project, busy=False)
            except SQLAlchemyError, OperationalError as err: 
                 # log something 
                 raise
            raise
return {"message": return_message}
```

But also we don't want SQLAlchemyError etc to be raised, we want `DatabaseError` to be raised, e.g. via the decorator that I mentioned in another comment (haven't checked yet if that's changed now)

i-oden: @rv0lt forget this

i-oden: The busy status is actually commited, I forgot. You don't need to change it at all.

i-oden: In fact move the `self.set_busy(project=project, busy=True)` to where you had it before, outside the try except

rv0lt: Okay, yes I was getting confused with the _set busy_. Bc with the tests I discovered that mocking one commit actually mocked the set busy. Good to confirm that it actually commits

i-oden: Yeah it's the decorator on the function that does it, I was confusing myself too

SPRINTLOG.md
```diff
@@ -314,3 +314,4 @@ _Nothing merged in CLI during this sprint_
 - Use full DDS name in MOTD email subject ([#1477](https://github.com/ScilifelabDataCentre/dds_web/pull/1477))
 - Add flag --verify-checksum to the comand in email template ([#1478])(https://github.com/ScilifelabDataCentre/dds_web/pull/1478)
 - Improved email layout; Highlighted information and commands when project is released ([#1479])(https://github.com/ScilifelabDataCentre/dds_web/pull/1479)
+- Added new API endpoint ProjectStatus.patch to extend the deadline ([#1480])(https://github.com/ScilifelabDataCentre/dds_web/pull/1480)
```
i-oden, line 317: Also this should be in a new sprint
```suggestion

# 2023-10-16 - 2023-10-27

- Added new API endpoint ProjectStatus.patch to extend the deadline ([#1480])(https://github.com/ScilifelabDataCentre/dds_web/pull/1480)
```

i-oden (changes requested):
Looks very good! After this I don't have anymore comments for now, just waiting to test with the CLI, so I'll approve both at the same time probably.
Just a couple tiny things:
- The sprintlog
- Change the description since we're not going to use the `ProjectStatus.get` endpoint anymore

dds_web/api/project.py
```diff
@@ -188,6 +188,123 @@ def post(self):
...
+            if new_deadline_in:
+                # deadline can only be extended from Available
+                if not project.current_status == "Available":
+                    raise DDSArgumentError(
+                        "You can only extend the deadline for a project that has the status 'Available'."
+                    )
+
+                if type(new_deadline_in) is not int:
+                    raise DDSArgumentError(
+                        message="The deadline atribute passed should be of type Int (i.e a number)."
```
i-oden, line 255:
```suggestion
                        message="The deadline attribute passed should be of type Int (i.e a number)."
```

dds_web/api/project.py
```diff
@@ -188,6 +188,123 @@ def post(self):
...
+                "warning": warning_message,
+            }
+            return json_returned
+
+        # Cannot change project status if project is busy
+        if project.busy:
+            raise ProjectBusyError(
+                message=(
+                    f"The deadline for the project '{project_id}' is already in the process of being changed. "
+                    "Please try again later. \n\nIf you know the project is not busy, contact support."
```
i-oden, line 228:
```suggestion
                    "Please try again later. \n\nIf you know that the project is not busy, contact support."
```

dds_web/api/project.py
```diff
@@ -188,6 +188,123 @@ def post(self):
...
+                except (sqlalchemy.exc.OperationalError, sqlalchemy.exc.SQLAlchemyError) as err:
+                    flask.current_app.logger.exception("Failed to extend deadline")
+                    db.session.rollback()
+                    raise
+
+                return_message = f"{project.public_id} has been given a new deadline"
+
+                return_message += (
+                    f". An e-mail notification has{' not ' if not send_email else ' '}been sent."
+                )
```
i-oden, line 299:
```suggestion
                return_message = (
                    f"{project.public_id} has been given a new deadline. "
                    f"An e-mail notification has{' not ' if not send_email else ' '}been sent."
                )
```

i-oden (changes requested):
A few minor things

tests/api/test_project.py
```diff
@@ -1079,6 +1124,326 @@ def test_projectstatus_post_invalid_deadline_expire(module_client, boto3_session
...
+        tests.DDSEndpoint.PROJECT_STATUS,
+        headers=tests.UserAuth(tests.USER_CREDENTIALS["unitadmin"]).token(module_client),
+        query_string={"project": project_id},
+        json=extend_deadline_data_no_confirmed,
+    )
+    # status code is ok but no operation perform
+    assert response.status_code == http.HTTPStatus.OK
+    assert project.times_expired == 0
+
+    assert "Operation must be confirmed before proceding." in response.json["warning"]
```
i-oden, line 1169: Also verify that the project status and info etc are returned
```suggestion
    assert "Operation must be confirmed before proceding." in response.json["warning"]
    assert all(item in response.json for item in ["project_info", "project_status", "warning", "default_unit_days"])
```

dds_web/api/project.py
```diff
@@ -188,6 +188,123 @@ def post(self):
...
+
+        # Cannot change project status if project is busy
+        if project.busy:
+            raise ProjectBusyError(
+                message=(
+                    f"The deadline for the project '{project_id}' is already in the process of being changed. "
+                    "Please try again later. \n\nIf you know that the project is not busy, contact support."
+                )
+            )
+
```
i-oden, line 231: Your test `test_extend_deadline_maxium_number_available_exceded` reminded me that we don't have a check for this prior to attempting the extension of the deadline. We need to check `project.times_expired` before trying to expire again.

i-oden (changes requested):

dds_web/api/project.py
i-oden: (And also before setting the project as busy)

rv0lt: You mean do the check of times expired before the set busy?

i-oden: Yeah

rv0lt: but if the check raises a DDSArgumentError. It will be caught by the last _except_ anyway and change the state to not busy again

i-oden: Yeah, but it's unnecessary for the code to go through all the following checks if it's anyway not possible to extend the deadline. 
For example if a user inputs a deadline that will not be approved, the user would currently potentially be changing the values and spending time on running the command multiple times, even though it won't matter since they can't change the deadline anyway. 

So actually we should put the `times_expired` check before the `confirmed_operation` part. 
1. check number of times_expired and return ProjectStatusError or something like that
2. check if operation confirmed 
3. continue with rest

i-oden: Plus the message that the user gets in the end is "Project availability limit: The maximun number of changes in data availability has been reached." which will confuse the users since they're not running `dds project status release` and they'll be thinking "I'm not trying to make the project available though, I'm just extending the deadline"

i-oden: Actually ignore that, the error message is fine. But the placement of the check should be changed to before `confirmed` check

i-oden: Or at least in the `try` but before all the other checks

dds_web/api/project.py
```diff
@@ -188,6 +188,129 @@ def post(self):
...
+        dds_web.utils.verify_project_access(project=project)
+
+        # Get json input from request
+        json_input = flask.request.get_json(silent=True)  # Already checked by json_required
+
+        # the status has changed at least two times,
+        # next time the project expires it wont change again -> error
+        if project.times_expired >= 2:
+            raise DDSArgumentError(
+                "Project availability limit: The maximun number of changes in data availability has been reached."
```
i-oden, line 210: A little typo
```suggestion
                "Project availability limit: The maximum number of changes in data availability has been reached."
```

dds_web/api/project.py
```diff
@@ -188,6 +188,129 @@ def post(self):
...
+        if not confirmed_operation:
+            warning_message = "Operation must be confirmed before proceding."
+            # When not confirmed, return information about the project
+            project_info = ProjectInfo().get()
+            project_status = self.get()
+            json_returned = {
+                **project_info,
+                "project_status": project_status,
+                "warning": warning_message,
+                "default_unit_days": project.responsible_unit.days_in_expired,
```
i-oden, line 226: Missed this but this should be `days_in_available`
```suggestion
                "default_unit_days": project.responsible_unit.days_in_available,
```

i-oden (changes requested):

dds_web/api/project.py
```diff
@@ -188,6 +188,129 @@ def post(self):
...
+
+                if type(new_deadline_in) is not int:
+                    raise DDSArgumentError(
+                        message="The deadline attribute passed should be of type Int (i.e a number)."
+                    )
+
+                # New deadline shouldnt surpass the default unit days
+                if new_deadline_in > default_unit_days:
+                    raise DDSArgumentError(
+                        message=f"You requested the deadline to be extended {new_deadline_in}. The number of days has to be lower than the default deadline extension number of {default_unit_days} days"
```
i-oden, line 268:
```suggestion
                        message=f"You requested the deadline to be extended {new_deadline_in} days. The number of days has to be lower than the default deadline extension number of {default_unit_days} days"
```

dds_web/api/project.py
```diff
@@ -188,6 +188,129 @@ def post(self):
...
+                    project.busy = False  # return to not busy
+                    db.session.commit()
+
+                except (sqlalchemy.exc.OperationalError, sqlalchemy.exc.SQLAlchemyError) as err:
+                    flask.current_app.logger.exception("Failed to extend deadline")
+                    db.session.rollback()
+                    raise
+
+                return_message = (
+                    f"{project.public_id} has been given a new deadline. "
```
i-oden, line 302:
```suggestion
                    f"The project '{project.public_id}' has been given a new deadline. "
```

i-oden (changes requested):
Looking good, need some small adjustments though - e.g. change from `days_in_expired` to `days_in_available` and a change in an error message for the users

i-oden:
@rv0lt So since the new statuses added to the `projectstatuses` table have the exact same creation date, I'm wondering if there's a risk that the code thinks that the `Expired` status is added last and therefore the system thinks that that is the current status..? Maybe we should do for example

```python
                    new_status_row = self.expire_project(
                        project=project,
                        current_time=curr_date - <1 second here>, 
                        deadline_in=project.responsible_unit.days_in_expired,
                    )
```
? 

Or if we should add checks in the tests that you have added that verifies that `project.current_status` is `Available`.

Also, currently we set (example above) `deadline_in = project.responsible_unit.days_in_expired` but it feels like it could become confusing to us in the case that the deadline for the `Expired` status is after the deadline for the `Available` status that's added.

rv0lt:
> @rv0lt So since the new statuses added to the `projectstatuses` table have the exact same creation date and time, I'm wondering if there's a risk that the code thinks that the `Expired` status is added last and therefore the system thinks that that is the current status..? Maybe we should do for example
> 
> ```python
>                     new_status_row = self.expire_project(
>                         project=project,
>                         current_time=curr_date - <1 second here>, 
>                         deadline_in=project.responsible_unit.days_in_expired,
>                     )
> ```
> 
> ?
> 
> Or if we should add checks in the tests that you have added that verifies that `project.current_status` is `Available`.
> 
> Also, currently we set (example above) `deadline_in = project.responsible_unit.days_in_expired` but it feels like it could become confusing to us in the case that the deadline for the `Expired` status is after the deadline for the `Available` status that's added. So I'm wondering if we should set the `deadline_in` here as e.g. tomorrow.

1.  The double-checking that the project is kept as available sounds alright. Maybe set again the curr date before the call of the expire and release so it has different seconds
2. Yeah sounds reasonable that the deadilne here can be some dummy data to avoid confussion

i-oden:
> > @rv0lt So since the new statuses added to the `projectstatuses` table have the exact same creation date and time, I'm wondering if there's a risk that the code thinks that the `Expired` status is added last and therefore the system thinks that that is the current status..? Maybe we should do for example
> > ```python
> >                     new_status_row = self.expire_project(
> >                         project=project,
> >                         current_time=curr_date - <1 second here>, 
> >                         deadline_in=project.responsible_unit.days_in_expired,
> >                     )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> > Or if we should add checks in the tests that you have added that verifies that `project.current_status` is `Available`.
> > Also, currently we set (example above) `deadline_in = project.responsible_unit.days_in_expired` but it feels like it could become confusing to us in the case that the deadline for the `Expired` status is after the deadline for the `Available` status that's added. So I'm wondering if we should set the `deadline_in` here as e.g. tomorrow.
> 
>     1. The double-checking that the project is kept as available sounds alright. Maybe set again the curr date before the call of the expire and release so it has different seconds
> 
>     2. Yeah sounds reasonable that the deadilne here can be some dummy data to avoid confussion

Great. Do that. 
1. Expired should have one second before Available.
2. Add check that the status is available still
3. Change `deadline_in=project.responsible_unit.days_in_expired` to `deadline_in=1`

i-oden (approved):
If the tests pass now, this PR is approved and you can merge!

PR #1551, opened 2024-09-09
Flask command to change unit quotas

rv0lt:

## 1. Description / Summary

Added a new command to update the quotas of a unit

[_Link to the github issue or add the Jira task ID here._](https://scilifelab.atlassian.net/jira/software/projects/DDS/boards/13?selectedIssue=DDS-2080)

## 3. Type of change

- [X] New feature
  - [X] Non-breaking 
- [ ] Bug fix 
- [ ] Security Alert fix 
  - [ ] Package update 
    - [ ] Major version update 
- [ ] Documentation 
- [ ] Workflow 
- [ ] Tests **only** 

## 4. Additional information

- [X] [Sprintlog](../SPRINTLOG.md) 
- [ ] Blocking PRs 
  - [ ] Merged 

## 5. Actions / Scans

- [X] **Black**

- [X] **Prettier**

- [X] **Yamllint**

- [X] **Tests**

- [X] **CodeQL**

- [X] **Trivy**

- [X] **Snyk**

### DIFF
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -282,13 +282,15 @@ def create_app(testing=False, database_uri=None):
             send_usage,
             collect_stats,
             monitor_usage,
-            update_unit,
+            update_unit_sto4,
+            update_unit_quota,
         )
 
         # Add flask commands - general
         app.cli.add_command(fill_db_wrapper)
         app.cli.add_command(create_new_unit)
-        app.cli.add_command(update_unit)
+        app.cli.add_command(update_unit_sto4)
+        app.cli.add_command(update_unit_quota)
         app.cli.add_command(update_uploaded_file_with_log)
         app.cli.add_command(lost_files_s3_db)
 
--- a/dds_web/commands.py
+++ b/dds_web/commands.py
@@ -227,7 +227,9 @@ def update_unit_quota(unit_id, quota):
         f"You are about to update the quota for unit '{unit_id}' to {quota} bytes (approx {round(quota / 1000 ** 3,2)} GB). Are you sure? \n"
     )
     if not do_update:
-        flask.current_app.logger.info(f"Cancelling quota update for unit '{unit_id}'. The quota is still {unit.quota} bytes.")
+        flask.current_app.logger.info(
+            f"Cancelling quota update for unit '{unit_id}'. The quota is still {unit.quota} bytes."
+        )
         return
 
     # Set sto4 info
--- a/dds_web/commands.py
+++ b/dds_web/commands.py
@@ -157,7 +157,7 @@ def create_new_unit(
     gc.collect()
 
 
-@click.command("update-unit")
+@click.command("update-unit-sto4")
 @click.option("--unit-id", "-u", type=str, required=True)
 @click.option("--sto4-endpoint", "-se", type=str, required=True)
 @click.option("--sto4-name", "-sn", type=str, required=True)
@@ -205,12 +205,12 @@ def update_unit_sto4(unit_id, sto4_endpoint, sto4_name, sto4_access, sto4_secret
     gc.collect()
 
 
-@click.command("update-unit")
+@click.command("update-unit-quota")
 @click.option("--unit-id", "-u", type=str, required=True)
 @click.option("--quota", "-q", type=int, required=True)
 @flask.cli.with_appcontext
 def update_unit_quota(unit_id, quota):
-    """Update unit quota."""
+    """Update unit quota. The quota is in bytes."""
     # Imports
     from dds_web import db
     from dds_web.database import models
@@ -219,7 +219,7 @@ def update_unit_quota(unit_id, quota):
     unit: models.Unit = models.Unit.query.filter_by(public_id=unit_id).one_or_none()
     if not unit:
         flask.current_app.logger.error(f"There is no unit with the public ID '{unit_id}'.")
-        return
+        sys.exit(1)
 
     # Set sto4 info
     unit.quota = quota
--- a/dds_web/commands.py
+++ b/dds_web/commands.py
@@ -212,6 +212,7 @@ def update_unit_sto4(unit_id, sto4_endpoint, sto4_name, sto4_access, sto4_secret
 def update_unit_quota(unit_id, quota):
     """Update unit quota. The quota is in bytes."""
     # Imports
+    import rich.prompt
     from dds_web import db
     from dds_web.database import models
 
@@ -220,6 +221,15 @@ def update_unit_quota(unit_id, quota):
     if not unit:
         flask.current_app.logger.error(f"There is no unit with the public ID '{unit_id}'.")
         sys.exit(1)
+
+    # ask the user for confirmation
+    do_update = rich.prompt.Confirm.ask(
+        f"You are about to update the quota for unit '{unit_id}' to {quota}. Are you sure? \n"
+        f"Remember that the quota is in bytes. You have select about {round(quota / 1024 ** 3,2)} GB."
+    )
+    if not do_update:
+        flask.current_app.logger.info(f"Cancelling quota update for unit '{unit_id}'.")
+        return
 
     # Set sto4 info
     unit.quota = quota
--- a/dds_web/commands.py
+++ b/dds_web/commands.py
@@ -224,8 +224,7 @@ def update_unit_quota(unit_id, quota):
 
     # ask the user for confirmation
     do_update = rich.prompt.Confirm.ask(
-        f"You are about to update the quota for unit '{unit_id}' to {quota}. Are you sure? \n"
-        f"Remember that the quota is in bytes. You have select about {round(quota / 1024 ** 3,2)} GB."
+        f"You are about to update the quota for unit '{unit_id}' to {quota} bytes (approx {round(quota / 1000 ** 3,2)} GB). Are you sure? \n"
     )
     if not do_update:
         flask.current_app.logger.info(f"Cancelling quota update for unit '{unit_id}'.")
--- a/dds_web/commands.py
+++ b/dds_web/commands.py
@@ -164,8 +164,8 @@ def create_new_unit(
 @click.option("--sto4-access", "-sa", type=str, required=True)
 @click.option("--sto4-secret", "-ss", type=str, required=True)
 @flask.cli.with_appcontext
-def update_unit(unit_id, sto4_endpoint, sto4_name, sto4_access, sto4_secret):
-    """Update unit info."""
+def update_unit_sto4(unit_id, sto4_endpoint, sto4_name, sto4_access, sto4_secret):
+    """Update unit sto4 storage info."""
     # Imports
     import rich.prompt
     from dds_web import db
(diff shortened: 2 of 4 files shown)


dds_web/commands.py
```diff
@@ -205,6 +205,39 @@ def update_unit(unit_id, sto4_endpoint, sto4_name, sto4_access, sto4_secret):
...
+    # Get unit
+    unit: models.Unit = models.Unit.query.filter_by(public_id=unit_id).one_or_none()
+    if not unit:
+        flask.current_app.logger.error(f"There is no unit with the public ID '{unit_id}'.")
+        sys.exit(1)
+
+    # ask the user for confirmation
+    do_update = rich.prompt.Confirm.ask(
+        f"You are about to update the quota for unit '{unit_id}' to {quota}. Are you sure? \n"
+        f"Remember that the quota is in bytes. You have select about {round(quota / 1024 ** 3,2)} GB."
```
i-oden, line 228: Good idea on adding the confirmation. Just a little change though in my opinion
```suggestion
        f"You are about to update the quota for unit '{unit_id}' to {quota} bytes (approx {round(quota / 1024 ** 3,2)} GB). Are you sure? \n"
```

dds_web/commands.py
```diff
@@ -205,6 +205,39 @@ def update_unit(unit_id, sto4_endpoint, sto4_name, sto4_access, sto4_secret):
...
+    )
+    if not do_update:
+        flask.current_app.logger.info(f"Cancelling quota update for unit '{unit_id}'.")
```
i-oden, line 231:
```suggestion
        flask.current_app.logger.info(f"Cancelling quota update for unit '{unit_id}'. The quota is still {unit.quota} bytes.")
```

dds_web/commands.py
i-oden: Also, I like the use of the gibibyte calc, but in this case I think it'll be better to use gigabytes (1000 instead of 1024)
```suggestion 
        f"You are about to update the quota for unit '{unit_id}' to {quota} bytes (approx {round(quota / 1000 ** 3,2)} GB). Are you sure? \n"
```

tests/test_commands.py
```diff
@@ -443,6 +444,109 @@ def test_update_unit_sto4_start_time_exists_mock_prompt_True(
     ]
 
 
+# update_unit_quota
+
+
+def test_update_unit_no_such_unit(client, runner, capfd: LogCaptureFixture) -> None:
```
i-oden, line 450: We need to rename the tests, they are currently duplicates --> only one of the tests will run. 
The previous one --> `test_update_unit_sto4_no_such_unit`
The new one --> `test_update_unit_quota_no_such_unit`

And the same for the other tests below where there are "duplicates"

i-oden (changes requested):

dds_web/__init__.py
```diff
@@ -282,13 +282,15 @@ def load_user(user_id):
             send_usage,
             collect_stats,
             monitor_usage,
-            update_unit,
+            update_unit_sto4,
```
valyo, line 285: why do we need to be specific about the unit's s3 storage? 
If/when we change s3 provider, or if we use a particular s3 provider for a particular unit, we will need to change command name again or/and add a new command

valyo (changes requested):
I find both commands difficult to use if I don't see the current info/quota values (they can be seen by querying the DB, but that's obviously not optimal). I suggest that the commands can at least print the current values before prompting for new ones.

dds_web/commands.py
```diff
@@ -157,15 +157,15 @@ def create_new_unit(
...
 
 
-@click.command("update-unit")
+@click.command("update-unit-sto4")
 @click.option("--unit-id", "-u", type=str, required=True)
 @click.option("--sto4-endpoint", "-se", type=str, required=True)
 @click.option("--sto4-name", "-sn", type=str, required=True)
 @click.option("--sto4-access", "-sa", type=str, required=True)
 @click.option("--sto4-secret", "-ss", type=str, required=True)
 @flask.cli.with_appcontext
```
valyo, line 166: again, it doesn't make sense for me to have "sto4" in all of these, because it seems the command is supposed to update all the values, which intuitively means changing the s3 provider

dds_web/__init__.py
rv0lt: I chose that name because the command specific sets the sto4 fields in the DB, and it has harcoded "sto4" as a string as well, so I figured it made sense as it is right now

rv0lt:
> I find both commands difficult to use if I don't see the current info/quota values (they can be seen by querying the DB, but that's obviously not optimal). I suggest that the commands can at least print the current values before prompting for new ones.

That's a valid point! I can add another line to the confirmation prompt for the new command. However, for the old command I am not sure if it should belong to a different PR

valyo (changes requested):
Why don't we add possibility for change of "External Display Name", DIA and DIE, while we are dealing with this. 
Is there a reason that all options of `update-unit-sto4` are required, maybe @i-oden  knows?

valyo:
> However, for the old command I am not sure if it should belong here or to a different PR

Yes, maybe the old command should be considered in a different PR

dds_web/commands.py
```diff
@@ -205,6 +205,40 @@ def update_unit(unit_id, sto4_endpoint, sto4_name, sto4_access, sto4_secret):
...
+        f"You are about to update the quota for unit '{unit_id}' to {quota} bytes (approx {round(quota / 1000 ** 3,2)} GB). Are you sure? \n"
+    )
+    if not do_update:
+        flask.current_app.logger.info(
+            f"Cancelling quota update for unit '{unit_id}'. The quota is still {unit.quota} bytes."
+        )
+        return
+
+    # Set sto4 info
+    unit.quota = quota
```
valyo, line 236: Since the quota will always be a relatively round number, it is better to have the input (and any printed output) in GB

valyo (changes requested):

dds_web/__init__.py
i-oden: We created this command when we were migrating from sto2 to sto4 and I believe we had the intention from the beginning on using `update_unit` as a command that could be used to update other things (and not just s3 info) in the future. We have the different options as `xxx_sto4` to match the fields in the database as @rv0lt said. I think for now we can keep it like this, and potentially change it when we remove the `sto2` info from the database in coming sprints.

i-oden:
> > I find both commands difficult to use if I don't see the current info/quota values (they can be seen by querying the DB, but that's obviously not optimal). I suggest that the commands can at least print the current values before prompting for new ones.
> 
> That's a valid point! I can add another line to the confirmation prompt for the new command. However, for the old command I am not sure if it should belong here or to a different PR

@rv0lt for the `update_unit_sto4` command, do not change it to display the info first. Sounds like a good idea for this new command though.

i-oden:
> Why don't we add possibility for change of "External Display Name", DIA and DIE, while we are dealing with this. Is there a reason that all options of `update-unit-sto4` are required, maybe @i-oden knows?

@valyo We shouldn't mix tasks, there's always a risk of complication and also we need to add more tests, plus it's more difficult to review. We should keep the PRs as simple as possible and also finish the tasks planned for the sprint before adding new ones from the backlog.

dds_web/commands.py
i-oden: Re: that all options are required -- it was because we didn't want to risk anything being left out when we were making the sto2 --> sto4 migration. 

Re: that we have `sto4` in all of these, I agree, we should change in the future but if we change it now we risk confusing ourselves in the future again. Now we have `sto2_name`, `sto4_name` etc in the database, so if we change this option in this command to `name` for example, then it will become more confusing to us later. We don't want to keep the `sto2` info in the future, so until we have removed that, we should keep as much detail as possible.

dds_web/commands.py
i-oden: Sure, we can change to GB instead and multiply to bytes before saving in the database

dds_web/__init__.py
valyo: we agreed to take this in a separate task/PR

rv0lt: Okay, just one thought, tho. 

If the intention was for this command to be more general, then it needs some refactoring, that's why I opted for a new command, it seemed cleaner as it is right now

dds_web/api/superadmin_only.py
valyo, line 1: Shouldn't this be a separate PR?

dds_web/commands.py
```diff
@@ -205,6 +205,43 @@ def update_unit(unit_id, sto4_endpoint, sto4_name, sto4_access, sto4_secret):
...
+
+    # ask the user for confirmation
+    do_update = rich.prompt.Confirm.ask(
+        f"Current quota for unit '{unit_id}' is {round(unit.quota / 1000 ** 3,2)} GB. \n"
+        f"You are about to update the quota to {quota} GB (aprox {quota * 1000 ** 3} bytes). \n"
+        "Are you sure you want to continue?"
+    )
+    if not do_update:
+        flask.current_app.logger.info(
+            f"Cancelling quota update for unit '{unit_id}'. The quota is still {unit.quota} bytes."
```
valyo, line 233: should be displayed in GB too

dds_web/static/swagger.yaml
valyo, line 1: same a for superadmin_only.py

valyo (changes requested):

dds_web/api/superadmin_only.py
rv0lt: ay, I am stupid, forgot to change branch

tests/test_commands.py
```diff
@@ -443,6 +444,109 @@ def test_update_unit_sto4_start_time_exists_mock_prompt_True(
...
+
+    # Get logging
+    _, err = capfd.readouterr()
+
+    # Verify message
+    assert f"There is no unit with the public ID '{command_options[1]}'." in err
+
+
+def test_update_unit_quota_confirm_prompt_False(client, runner, capfd: LogCaptureFixture) -> None:
+    """Start time already recorded. Answer no to prompt about update anyway. No changes should be made."""
```
valyo, line 473: I don't completely understand, especially the first sentence. 
Maybe something like 
```suggestion
    """Unit quota should not be changed when answer to prompt is False."""
```

tests/test_commands.py
```diff
@@ -443,6 +444,109 @@ def test_update_unit_sto4_start_time_exists_mock_prompt_True(
...
+    assert f"Unit '{unit_id}' updated successfully" not in err
+
+    # Verify no change in unit
+    unit: models.Unit = models.Unit.query.filter_by(public_id=unit_id).first()
+    assert unit
+    assert unit.quota == quota_original
+
+
+def test_update_unit_quota_confirm_prompt_true(client, runner, capfd: LogCaptureFixture) -> None:
+    """Sucessfully update unit quota - The answer to the prompt is True."""
```
valyo, line 511:
```suggestion
    """Unit quota successfully updated when answer to the prompt is True."""
```

valyo (changes requested):

i-oden:
> > @rv0lt for the `update_unit_sto4` command, do not change it to display the info first.
> 
> @i-oden do you mean in this PR, or you think it should not display the info at all?

We shouldn't display them at all, at least not the keys. We can display the endpoint, like "sto2" or "sto4" or something like that but other than that, no.

i-oden (approved):

PR #1562, opened 2024-10-11
Alert unit email and us if storage threshold reached

rv0lt:

## 1. Description / Summary

Modify the monitor usage command to send an email to both the unit affected and our team in case the warning level is surpassed

[_Link to the github issue or add the Jira task ID here._](https://scilifelab.atlassian.net/jira/software/projects/DDS/boards/13?selectedIssue=DDS-1675)

## 3. Type of change

- [x] New feature
  - [x] Non-breaking 
- [x] Database change: _Remember the to include a new migration version, **or** explain here why it's not needed._ 
- [ ] Bug fix 
- [ ] Security Alert fix 
  - [ ] Package update 
    - [ ] Major version update 
- [ ] Documentation 
- [ ] Workflow 
- [ ] Tests **only** 

## 4. Additional information

- [ ] Blocking PRs 
  - [ ] Merged 

## 5. Actions / Scans

- [x] **Black**

- [x] **Prettier**

- [x] **Yamllint**

- [x] **Tests**

- [x] **CodeQL**

- [x] **Trivy**

- [x] **Snyk**


i-oden (changes requested):
Just a quick note -- add a description to the PR

valyo:
Also, the title says "instead", and it should be "Alert unit and us..." or something like that

SPRINTLOG.md
```diff
@@ -444,3 +444,7 @@ _Nothing merged during this sprint_
 - Modify the invoicing commands to send the instance name in the emails([#1561](https://github.com/ScilifelabDataCentre/dds_web/pull/1561))
 - Fix the MOTD endpoint according to post merge review([#1564](https://github.com/ScilifelabDataCentre/dds_web/pull/1564))
 - New version & changelog([#1565](https://github.com/ScilifelabDataCentre/dds_web/pull/1565))
+
+# 2024-10-23 - 2024-11-01
+
+- Modify the monitor usage command to send warning to the affected unit instead of Data Centre([#1562](https://github.com/ScilifelabDataCentre/dds_web/pull/1562))
```
i-oden, line 450: Not instead, both

valyo (approved):

dds_web/commands.py
```diff
@@ -1305,15 +1305,16 @@ def monitor_usage():
         # Email if the unit is using more
         if perc_used_decimal > warn_after:
             # Email settings
+            unit_contact: str = unit.contact_email
             message: str = (
-                "A SciLifeLab Unit is approaching the allocated data quota.\n"
-                f"Affected unit: {unit.name}\n"
+                "Your unit is approaching the allocated data quota.\n"
```
i-oden, line 1310:
```suggestion
                "Your unit is approaching the allocated data quota (see details below).\n\n"
                f"NB! If you would like to increase or decrease the allocated quota ('Quota') or the level after which you receive this email ('Warning level'), the technical contact person for your unit must send a request to {dds_contact}.\n"
```

i-oden (changes requested):
Aside from my comments, we need to have some way of not spamming the units with these emails. Should we have a new table keeping track of if this email has been sent and in that case don't send this again (could probably add a reminder though, not in this PR though)? Any other / better ideas for keeping track of this? 
@ScilifelabDataCentre/teamhermes

valyo:
> Aside from my comments, we need to have some way of not spamming the units with these emails. Should we have a new table keeping track of if this email has been sent and in that case don't send this again (could probably add a reminder though, not in this PR though)? Any other / better ideas for keeping track of this? @ScilifelabDataCentre/teamhermes

I am not sure I understand the question, what should we keep track of?

valyo:
Wouldn't it be sent once a day, when the daily cronjob runs? In such case, if nothing has changed, i.e. threshold hasn't been changed and or usage hasn't gone down, then the email should be send again to all concerned parties.

i-oden:
> Wouldn't it be sent once a day, when the daily cronjob runs? In such case, if nothing has changed, i.e. threshold hasn't been changed and or usage hasn't gone down, then the email should be send again to all concerned parties.

But the units have a threshold of 80%. I don't think it's reasonable to ask them to immediately message us to ask for a change in warning level, or to expect that we would be able to change the warning level for them immediately (worst case multiple units on the same day), and even if they reach the threshold they may not want to increase their quota because they may know that they won't be uploading that much now or that there are some projects that will be archived any second. We shouldn't force the units to change anything, and in the case that they would want to change the quota: We don't do that within a day, there are steps, partly within database but also to check with SUNET/Safespring that there is enough space, and to have them change settings, and to inform e.g. Jonas or other DCMG about the change / sometimes get approval. 

My first though was to  
- create a table `WarningEmails` (or something) with the columns `unit`, `date` and `active` 
- before sending this email, check if there's a row in the table 
- if there is a row in the table --> check when the row was added (when the email was sent) 
    - if the row was added less than a week ago --> don't send another email 
    - if the row was added at least a week ago --> send a reminder with 1. updated information, 2. info that they will keep getting this reminder (or something like this)
- if there's not a row in the table --> send email and add row to table 

Now thinking about it, I have another couple of options which might be better:
**Option 1** 
- May need a new table here too to keep track of which warning has been sent
- Change the `warning_level` column to a list or something if possible, to contain multiple levels, e.g. 80, 90, 95. 
- Cronjob does normal usage calculation and sees if any of the warning levels have been reached 
- In that case checks if the warning email has already been sent 
    - If the warning has already been sent --> check which level it was sent on and possibly send a new warning if new warning level has been passed 
    - If warning has not been sent --> send email
- If warning level has not been passed and there is a row in table --> remove rows from table 

**Option 2**
- Add columns to `Unit` table: `warnings_active` (bool) and `warnings_deactivated` (date and time) 
- When cronjob is run, check if `warnings_active == True` and in that case send the warning
    - Also include in warning email that if they do not message us to say they don't want warnings then they will keep getting them every day
- If `warnings_active == False` --> check if the warning has been deactivated for x number of days 
    - If so --> set `warnings_active` to True again and null `warnings_deactivated`
    - If not --> no warning sent 

@ScilifelabDataCentre/teamhermes These are just options and thoughts on my end and some are probably a bit too complicated / too much work, I'm completely open to suggestions.

valyo:
My opinion is that such kind of warning should be as straightforward as possible. Trying to implement some logic in order to spare the units email notification when they are above the threshold is not going to do any good to them and/or to us. On the contrary, it will potentially lead to more work for us trying to handle it if they "dangerously" approach the quota, plus that it will complicate the system even more. 
I think that improving the system stability, which will allow us to onboard more units is far more important.

rv0lt:
My commnets:

1. Agree about the contact email thing, didn't realise it could be sent to null, and that does not make sense. So agree it should be non nullable

2. About the emails, maybe we can discuss the periodicity. But I definitely think that this "spamming" is not bad. Most other cloud providers will also do that (attach some monthly emails I was getting from AWS). And actually, the thing is that if they get issues with the storage they will also complain, it is a way to cover ourselves: "Hey, you were getting alerts quite often, you should have done something". 
I don't see it as spam, as they agree to get alerts in that threshold. They don't want it? Okay, then change the **warning_level**


image

i-oden:
> My opinion is that such kind of warning should be as straightforward as possible. Trying to implement some logic in order to spare the units email notification when they are above the threshold is not going to do any good to them and/or to us. On the contrary, it will potentially lead to more work for us trying to handle it if they "dangerously" approach the quota, plus that it will complicate the system even more. I think that improving the system stability, which will allow us to onboard more units is far more important.

@valyo
- Re: as straight forward as possible -- I completely agree. My thoughts / options were really just that, I agree that they would take too much time and effort atm. If we do fix something then it should be a simple fix.
- Re: that improving the system stability is far more important, ofc, I agree.

@rv0lt
- Re: 1 -- :+1::skin-tone-2:
- Re: your second point, yes, I agree, fair enough.

@ScilifelabDataCentre/teamhermes
My point was mainly: Yes, spamming them with these warnings is ok, until we know that they have the information. What if they tell us "yes, we want to change the quota", but we can't do that in a day or maybe even a week, due to contracts or general negotiations -- Do they (and we) just continue getting these warning emails every day until it's potentially fixed? This will lead us to get both the cronjob emails, and heaps of emails from the units about confusion "I thought we were fixing this already" and "please stop with these emails now" which will also take up our time.
I do agree though that they could just change the warning level if they don't want the warnings, BUT then we need to add this to the email. So then the message should be this in that case.
```python
message: str = (
    "Your unit is approaching the allocated data quota (see details below).\n\n"
    "NB! You will continue to get this email until you adjust your allocated quota ('Quota') "
    "or / and the level after which you receive this email ('Warning level'). "
    f"In order to change these settings, the technical contact person for your unit must send a request to {dds_contact}.\n"
    f"Unit name: {unit.name}"
    f"{info_string}"
)
```

migrations/versions/e02fe8fde71e_unit_contact_email_non_nullable.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = "e02fe8fde71e"
+down_revision = None
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
i-oden, line 20: In the upgrade, also add values to rows where the contact_email is not filled in. See other migrations where similar things have been done.

i-oden (changes requested):

i-oden (changes requested):

migrations/README
```diff
@@ -1 +1,39 @@
-Single-database configuration for Flask.
+# Database Migration Guide
```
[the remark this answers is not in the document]
rv0lt: The link you mentioned in the reply is just a typo right?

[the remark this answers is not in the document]
rv0lt: Actually I didn't know about this section in the root, I thought there was no info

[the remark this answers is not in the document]
i-oden: > The link you mentioned in the reply is just a typo right?

Oops, now the link is correct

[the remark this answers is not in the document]
i-oden: > Actually I didn't know about this section in the root, I thought there was no info

Aha, sorry about that, I didn't get that you didn't know about that section.

rv0lt:
Checked the two instances and there is no current unit that has a null contact_email

i-oden (changes requested):
Even though we currently don't have any values that are null, we still need to include the possibility in the migration. An example is seen in `666003748d14_change_active_nullable_and_default` for example, but there are others too (not regarding nullable=False though I don't think), add something similar.

migrations/versions/0cd0a3b251e0_unit_contact_email_non_nullable.py
```diff
@@ -0,0 +1,40 @@
...
+down_revision = "3d610b382383"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute(
+        "UPDATE units SET contact_email = 'delivery@scilifelab.se' WHERE contact_email IS NULL"
+    )
```
i-oden, line 24: Currently other migrations e.g. `666003748d14_change_active_nullable_and_default.py` are using another method of executing these things, and I don't want to introduce different methods, so my suggestion is to change to the following
```suggestion
    unit_table = sa.sql.table("units", sa.sql.column("contact_email", mysql.TINYINT(display_width=1)))
    op.execute(unit_table.update().where(unit_table.c.contact_email == None).values(contact_email="delivery@scilifelab.se"))
```
If there rest of the team @ScilifelabDataCentre/teamhermes disagrees and thinks we should have SQL instead, then I want to change the other migrations to have this structure too.

i-oden (changes requested):

migrations/versions/0cd0a3b251e0_unit_contact_email_non_nullable.py
rv0lt: I can change, it is esentially the same so i dont mind

valyo: I think it makes sense to be consistent, i.e. to use the `sa` (despite of SQL being easier for me to read because I am not that used to `sa` :) )

i-oden (approved):
OK to merge **once you have tested it locally** (using `flask db downgrade` and `upgrade` etc)

rv0lt:
> OK to merge **once you have tested it locally** (using `flask db downgrade` and `upgrade` etc)

Tested manually

PR #1574, opened 2024-11-19
Log which user reset password

reutenauer:

## 1. Description / Summary

The task was to add in logging which user reset the password. The relevant function seemed to be password_reset_completed, that already had logging, using the decorator @logging_bind_request. However, since the user is not logged in at any point during the password resetting process, the decorator only logged the IP address the request came from. The user can be deduced from the value of the reset token, but it didn’t seem appropriate to modify the decorator for such a specific case so I just removed it and called the action logger from password_reset_completed explicitly.

GitHub [#1574](https://github.com/ScilifelabDataCentre/dds_web/pull/1574), JIRA [DDS-2178](https://scilifelab.atlassian.net/browse/DDS-2178)

## 3. Type of change

- [X] New feature
  - [X] Non-breaking 
- [ ] Bug fix 
- [ ] Security Alert fix 
  - [ ] Package update 
    - [ ] Major version update 
- [ ] Documentation 
- [ ] Workflow 
- [ ] Tests **only** 

## 4. Additional information

- [X] [Sprintlog](../SPRINTLOG.md) 
- [ ] Blocking PRs 
  - [ ] Merged 

## 5. Actions / Scans

- [x] **Black**

- [x] **Prettier**

- [x] **Yamllint**

- [x] **Tests**

- [x] **CodeQL**

- [x] **Trivy**

- [x] **Snyk**

[DDS-2178]: https://scilifelab.atlassian.net/browse/DDS-2178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reutenauer:
The task was to add in logging which user reset the password.  The relevant function seemed to be `password_reset_completed`, that already had logging, using the decorator `@logging_bind_request`.  However, since the user is not logged in at any point during the password resetting process, the decorator only logged the IP address the request came from.  The user can be deduced from the value of the reset token, but it didn’t seem appropriate to modify the decorator for such a specific case so I just removed it and called the action logger from `password_reset_completed` explicitly.

dds_web/web/user.py
```diff
@@ -604,6 +604,8 @@ def password_reset_completed():
             for unit_admin in unit_admins_to_contact:
                 dds_web.utils.send_project_access_reset_email(unit_admin, email[0], token)
 
+    action_logger = structlog.getLogger("actions")
+    action_logger.info(f"{flask.request.endpoint}", user=user.name, resource=flask.request.path)
```
valyo, line 608: We definitely want to see the `username` in the log, but maybe having the name is also good, so something like 
```suggestion
    action_logger.info(f"{flask.request.endpoint}", user=user.name, username=user.username, resource=flask.request.path)
```

I was wondering about `flask.request.endpoint`, which in this case results in the same thing as the `flask.request.path`, but maybe it's good to have in case the URL path of the request is different from the view function handling the request

valyo (changes requested):

dds_web/web/user.py
reutenauer: I’ve done that now, thanks for the discussion!

valyo (approved):
Maybe your comment explaining the issue and the solution can be pasted in the description part, otherwise looks good.

rv0lt:
I don't how this action logger works, but I am not unable to reproduce the logs locally. 

I am able, however, by using the flask logs. Which is used in most parts of the code, it seems a bit weird that we have two differents ways to call the logs in the code, although this structlogs is used in very few and specific places.
https://flask.palletsprojects.com/en/stable/logging/


Something like:
```
flask.current_app.logger.info(
        f"{flask.request.endpoint} {user.username} "
    )
```

(But a bit more explicative)



IN myopinion, I prefer using the flask logs, as it is more natural than manually leading with the struclog, and it is used in most other parts of the code

rv0lt (changes requested):
Add a small description of the task

valyo:
> I don't how this action logger works, but I am not unable to reproduce the logs locally.
> 
> I am able, however, by using the flask logs. Which is used in most parts of the code, it seems a bit weird that we have two differents ways to call the logs in the code, although this structlogs is used in very few and specific places. https://flask.palletsprojects.com/en/stable/logging/
> 
> Something like:
> 
> ```
> flask.current_app.logger.info(
>         f"{flask.request.endpoint} {user.username} "
>     )
> ```
> 
> (But a bit more explicative)
> 
> IN myopinion, I prefer using the flask logs, as it is more natural than manually leading with the struclog, and it is used in most other parts of the code

@rv0lt the action logger outputs in /dds_web/logs/actions.log and you can see the result of this change there. Your comment is a good one and I think we should discuss it. Maybe here you point is mainly from the coding point of view, but I've always been bothered by the effect of this on the troubleshooting process. What I mean: when seeing errors in the container log, which is basically the flask logs, it is not straightforward to figure out what action caused it, which project, user, etc, because one needs to look at a separate log file. 
I think we should look if there was a good reson for this approach, and consider changing it.

valyo:
well, they are actually three - there's also /dds_web/logs/dds.log 😸

i-oden (commented):
Review from me not needed **but** remember to add PR description.

PR #1575, opened 2024-11-26
[DDS-2077] Quick and dirty fix for systematic 500 error on dds ls --tree.

reutenauer:
The immediate cause of the error is obvious, and shown prominently in the logs: we’re calling `.rstrip` on a variable whose value is `None` at that point.  This has an equally obvious fix, which I implemented in this pull request.  I didn’t go much deeper in the code to see why the value is `None` since I understand we want to refactor anyway; I only note that the 500 error was systematic, so that we might want to get right of that line entirely; and that the method name of `old_get` doesn’t inspire a great deal of confidence ;-)

## 1. Description / Summary

Quick and dirty fix for the fact that `dds ls --tree` systematically failed.

https://scilifelab.atlassian.net/browse/DDS-2077

## 3. Type of change

- [ ] New feature
  - [ ] Non-breaking 
- [x] Bug fix 
- [ ] Security Alert fix 
  - [ ] Package update 
    - [ ] Major version update 
- [ ] Documentation 
- [ ] Workflow 
- [ ] Tests **only** 

## 4. Additional information

- [ ] Blocking PRs 
  - [ ] Merged 

## 5. Actions / Scans

- [x] **Black**

- [x] **Prettier**

- [x] **Yamllint**

- [x] **Tests**

- [x] **CodeQL**

- [x] **Trivy**

- [x] **Snyk**


valyo (commented):
Seems to work, but I am not sure (without being digging into the code too much) what the idea of `subpath` was from the beginning, I guess that's why you label it with "Quick and dirty fix". 
Question is should we address it properly here, or accept this fix and create a separate task for dealing with `subpath`
Another note is that the fix should also be done in the "new" get, since it will be the one used with api/v3

rv0lt:
So the explanation is as follows:

When the client is called with the --tree option, it makes a recursive request to the list files endpoint. The API has no option to return as a tree, that logic is handled directly by the client. You can take a more interactive look at the endpoint here **https://delivery.scilifelab.se/:5000/api/documentation/v3/#/files/listFiles**

Going back to this, because of how the python module handles things, when an option is None in a body parameter, it is translated as null in the generated request. You can manually reproduce the API call with:

```
curl -X 'GET' \
  'localhost:5000/api/v1/files/list?project=project_1' \
  -H 'accept: application/json' \
  -H 'X-CLI-Version: 2.0.1' \
  -H 'Authorization: Bearer eyJhbGciOiJBMjU2S1ciLCJlbmMiOiJBMjU2R0NNIiwiZXhwIjoiMjAyNC0xMS0yOVQwODo1NToxNy40MDg2MzYifQ.jGJ2P9qDcEQtspW84caPr5tU0hlIKiQrQ5lqZUJ0njJW8c5w2TqqPA.zIAZazYvfpVPR6Is.sHmrxtwN_b3LIIJZm6zjrcIIA57oGWv7m-RKfqlml0_lkXYf0sHpXryNINxaCXy7aXmbM7ocRIm5_SNDlLsYMfzF_wPREvVb1O3kqm6gTUeQFAb-lhHrmxM4UA-BaXJikSAS3iRaP0u70axjyDly7oWLjDm5hsr6Gaml_prXNZEuIr32QdTd2G0P_QzA3VRHvvUGPwgU6beGWVEaYPHQZmubPax8bu1WF5dGLBxTZfbYDYZNtFLMHHOPRT9VkJQp2dQcHVGTWx8SFd4oyUk9fo4YmrhGcaYxljqVA9IvuXOpsdOJxGsRYs9YvFHFHPvRHLaW3nBDoLYyKl9F80JaebBMDpFMYZS5BXj1GUdHf-coCLvKDxEYUVitoFcu9W2aHtb0MClAb4xWEew.Xy0yd3OCbcqq36_35Apoew' \
  -H 'Content-Type: application/json' \
  -d '{
  "show_size": "false",
  "subpath": null
}'
```

Because of this, extra_args.get("subpath", ".") → returns None.


Why, it does not happen in the other method? i.e the one that is not under old_get. Because in this case, it expects the data to come from the request. And in this case: flask.request.args.get("subpath", default=".") is able to identify that it is null and replace it with the "."

Which should be the default option.


The two ways to handle this would be to do a deep check of the client, to replace the API calls to the backend, or this change, albeit with some small modification that I will propose now.


And as I wrote when I reported this task, I strongly suggest adding some tests. This error stepped under our noses when there should have been a test to detect it

dds_web/api/files.py
```diff
@@ -324,7 +324,7 @@ def old_get(self, project):
         show_size = extra_args.get("show_size")
 
         # Check if to get from root or folder
-        subpath = extra_args.get("subpath", ".").rstrip(os.sep)
+        subpath = (extra_args.get("subpath", ".") or "").rstrip(os.sep)
```
rv0lt, line 327:
```suggestion
        subpath = (extra_args.get("subpath") or ".").rstrip(os.sep)
```

rv0lt (changes requested):
Default should be "."

valyo:
ok, but could `subpath` possibly be?

i-oden (changes requested):
Want to review.

SPRINTLOG.md
```diff
@@ -454,3 +454,7 @@ _Nothing merged during this sprint_
 # 2024-11-04 - 2024-11-15
 
 - Removed exception for invalid token to simplify logging and reduce unnecessary error entries ([#1572](https://github.com/ScilifelabDataCentre/dds_web/pull/1572))
+
+# 2024-11-18 – 2024-11-29
+
+- Bugfix: Quick and dirty change to prevent `dds ls --tree` from failing systematically ([#1575](https://github.com/ScilifelabDataCentre/dds_web/pull/1575).
```
i-oden, line 460: Move to new sprint.

i-oden (changes requested):
I agree with Alvaros suggestion

reutenauer:
On Mon, Dec 02, 2024 at 02:19:18AM -0800, Ina Odén Österbo wrote:
> I agree with Alvaros suggestion 

  Adopted!

SPRINTLOG.md
```diff
@@ -458,3 +458,4 @@ _Nothing merged during this sprint_
 # 2024-11-18 – 2024-11-29
 
 - Logging: Add which user name reset password ([https://github.com/ScilifelabDataCentre/dds_web/pull/1574](https://github.com/ScilifelabDataCentre/dds_web/pull/1574))
+- Bugfix: Quick and dirty change to prevent `dds ls --tree` from failing systematically ([#1575](https://github.com/ScilifelabDataCentre/dds_web/pull/1575).
```
i-oden, line 461: Small thing but just because the final fixes and merge is done in this sprint.
```suggestion

# 2024-12-02 - 2024-12-13
- Bugfix: Quick and dirty change to prevent `dds ls --tree` from failing systematically ([#1575](https://github.com/ScilifelabDataCentre/dds_web/pull/1575).
```

reutenauer:
> @reutenauer just checking that you understood that I meant this suggestion and not the other one that you've already fixed 

  Sorry – I had misunderstood “move to new sprint”, indeed; but it’s changed now.

i-oden:
> @reutenauer just checking that you understood that I meant this suggestion and not the other one that you've already fixed
> Sorry – I had misunderstood “move to new sprint”, indeed; but it’s changed now.

@reutenauer No worries! This is ok to merge!

i-oden:
> @reutenauer just checking that you understood that I meant this suggestion and not the other one that you've already fixed
> Sorry – I had misunderstood “move to new sprint”, indeed; but it’s changed now.

@reutenauer No worries! When Alvaro has approved you can merge.
@rv0lt Can you have a look and approve?

PR #1591, opened 2025-02-04
Implement Redis Queue for Async Requests & Project Deletion

rv0lt:

## 1. Description / Summary

This PR contains a first implementation for a Redis Queue to process some of the requests.
 - There is a new dashboard where you can check the status of the requests at **http://localhost:5000/rq/**
 - For tests, there is a new fixture to bypass the queue and execute the function. New tests for endpoints that use the queues can invoke it
 - The implementation of delete projects contents has been done as POC
 - For motd there is a template and suggestions to keep working, but it needs a larger change due to how rq handles the jobs

**To test the PR**:

Run the containers in full-dev mode 
`docker-compose --profile full-dev up`

- Log in the client container
Authenticate with any Unit Admin or Unit Personal.
Create a new project and upload some data
Remove all the data
`dds data rm -p someunit00001 -a`

- Check the dashboard at **http://localhost:5000/rq/** to see the status of the job, when it is finished it means it is done
- Check the minio at **http://localhost:9001/buckets** to ensure that all files have been deleted

[ Jira task ID](https://scilifelab.atlassian.net/jira/software/projects/HMS/boards/13?selectedIssue=HMS-2297&sprintStarted=true)

- New feature
- Bug fix --> label: `bug` 
- Security Alert fix 
    - Major version update 

## 4. Additional information

## 5. Actions / Scans

- **Black**

- **Prettier**

- **Yamllint**

- **Tests**

- **CodeQL**

- **Trivy**

- **Snyk**

### DIFF
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -297,18 +297,6 @@ def create_app(testing=False, database_uri=None):
             client_kwargs={"scope": "openid profile email"},
         )
 
-        # Redis Worker needs to run as its own process, we initialize it here.
-        # If some worker was already running, it will not be started again.
-        redis_url = app.config.get("REDIS_URL")
-        redis_connection = Redis.from_url(redis_url)
-
-        workers = Worker.all(redis_connection)
-        if not workers:
-            worker = Worker(["default"], connection=redis_connection)
-            worker.log = app.logger
-            p = multiprocessing.Process(target=worker.work, daemon=True)
-            p.start()
-
         # Import flask commands - all
         from dds_web.commands import (
             fill_db_wrapper,
@@ -361,6 +349,19 @@ def create_app(testing=False, database_uri=None):
             from dds_web.web.root import pages
             from dds_web.web.user import auth_blueprint
             from flask_swagger_ui import get_swaggerui_blueprint
+
+            # Redis Worker needs to run as its own process, we initialize it here.
+            # If some worker was already running, it will not be started again.
+            redis_url = app.config.get("REDIS_URL")
+            redis_connection = Redis.from_url(redis_url)
+
+            workers = Worker.all(redis_connection)
+            if not workers:
+                worker = Worker(["default"], connection=redis_connection)
+                worker.log = app.logger
+                p = multiprocessing.Process(target=worker.work, daemon=True)
+                p.start()
+
 
             # base url for the api documentation
             SWAGGER_URL_1 = "/api/documentation/v1"
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -362,7 +362,6 @@ def create_app(testing=False, database_uri=None):
                 p = multiprocessing.Process(target=worker.work, daemon=True)
                 p.start()
 
-
             # base url for the api documentation
             SWAGGER_URL_1 = "/api/documentation/v1"
             SWAGGER_URL_3 = "/api/documentation/v3"
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -9,6 +9,7 @@ import logging
 import pathlib
 import sys
 import os
+import multiprocessing
 
 # Installed
 import flask
@@ -21,6 +22,8 @@ import flask_mail
 import flask_login
 import flask_migrate
 import rq_dashboard
+from redis import Redis
+from rq import Worker
 
 # import flask_qrcode
 from werkzeug.middleware.proxy_fix import ProxyFix
@@ -294,6 +297,18 @@ def create_app(testing=False, database_uri=None):
             client_kwargs={"scope": "openid profile email"},
         )
 
+        # Redis Worker needs to run as its own process, we initialize it here.
+        # If some worker was already running, it will not be started again.
+        redis_url = app.config.get("REDIS_URL")
+        redis_connection = Redis.from_url(redis_url)
+
+        workers = Worker.all(redis_connection)
+        if not workers:
+            worker = Worker(["default"], connection=redis_connection)
+            worker.log = app.logger
+            p = multiprocessing.Process(target=worker.work, daemon=True)
+            p.start()
+
         # Import flask commands - all
         from dds_web.commands import (
             fill_db_wrapper,
@@ -309,7 +324,7 @@ def create_app(testing=False, database_uri=None):
             monitor_usage,
             update_unit_sto4,
             update_unit_quota,
-            run_redis_worker,
+            restart_redis_worker,
         )
 
         # Add flask commands - general
@@ -319,7 +334,7 @@ def create_app(testing=False, database_uri=None):
         app.cli.add_command(update_unit_quota)
         app.cli.add_command(update_uploaded_file_with_log)
         app.cli.add_command(lost_files_s3_db)
-        app.cli.add_command(run_redis_worker)
+        app.cli.add_command(restart_redis_worker)
 
         # Add flask commands - cronjobs
         app.cli.add_command(set_available_to_expired)
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -309,7 +309,7 @@ def create_app(testing=False, database_uri=None):
             monitor_usage,
             update_unit_sto4,
             update_unit_quota,
-            run_worker,
+            run_redis_worker,
         )
 
         # Add flask commands - general
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -20,7 +20,7 @@ from flask_httpauth import HTTPBasicAuth, HTTPTokenAuth
 import flask_mail
 import flask_login
 import flask_migrate
-
+import rq_dashboard
 
 # import flask_qrcode
 from werkzeug.middleware.proxy_fix import ProxyFix
@@ -198,6 +198,10 @@ def create_app(testing=False, database_uri=None):
         # User config file, if e.g. using in production
         app.config.from_envvar("DDS_APP_CONFIG", silent=True)
 
+        # redis dashboard
+        app.config.from_object(rq_dashboard.default_settings)
+        rq_dashboard.web.setup_rq_connection(app)
+
         # Test related configs
         if database_uri is not None:
             app.config["SQLALCHEMY_DATABASE_URI"] = database_uri
@@ -305,6 +309,7 @@ def create_app(testing=False, database_uri=None):
             monitor_usage,
             update_unit_sto4,
             update_unit_quota,
+            run_worker,
         )
 
         # Add flask commands - general
@@ -323,6 +328,7 @@ def create_app(testing=False, database_uri=None):
         app.cli.add_command(send_usage)
         app.cli.add_command(collect_stats)
         app.cli.add_command(monitor_usage)
+        app.cli.add_command(run_worker)
 
         # Make version available inside jinja templates:
         @app.template_filter("dds_version")
@@ -374,6 +380,8 @@ def create_app(testing=False, database_uri=None):
             app.register_blueprint(swagger_ui_blueprint_v3, url_prefix=SWAGGER_URL_3, name="v3")
             app.register_blueprint(api_blueprint, url_prefix="/api/v1")
             app.register_blueprint(api_blueprint_v3, url_prefix="/api/v3")
+
+            app.register_blueprint(rq_dashboard.blueprint, url_prefix="/rq")
             app.register_blueprint(pages, url_prefix="")
             app.register_blueprint(auth_blueprint, url_prefix="")
 
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -319,6 +319,7 @@ def create_app(testing=False, database_uri=None):
         app.cli.add_command(update_unit_quota)
         app.cli.add_command(update_uploaded_file_with_log)
         app.cli.add_command(lost_files_s3_db)
+        app.cli.add_command(run_redis_worker)
 
         # Add flask commands - cronjobs
         app.cli.add_command(set_available_to_expired)
@@ -328,7 +329,6 @@ def create_app(testing=False, database_uri=None):
         app.cli.add_command(send_usage)
         app.cli.add_command(collect_stats)
         app.cli.add_command(monitor_usage)
-        app.cli.add_command(run_redis_worker)
 
         # Make version available inside jinja templates:
         @app.template_filter("dds_version")
--- a/dds_web/__init__.py
+++ b/dds_web/__init__.py
@@ -328,7 +328,7 @@ def create_app(testing=False, database_uri=None):
         app.cli.add_command(send_usage)
         app.cli.add_command(collect_stats)
         app.cli.add_command(monitor_usage)
-        app.cli.add_command(run_worker)
+        app.cli.add_command(run_redis_worker)
 
         # Make version available inside jinja templates:
         @app.template_filter("dds_version")
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -15,6 +15,8 @@ import sqlalchemy
 import datetime
 import botocore
 import marshmallow
+from rq import Queue
+from redis import Redis
 
 # Own modules
 import dds_web.utils
@@ -479,7 +481,9 @@ class ProjectStatus(flask_restful.Resource):
 
         try:
             # Deletes files (also commits session in the function - possibly refactor later)
-            RemoveContents().delete_project_contents(project=project, delete_bucket=True)
+            RemoveContents().delete_project_contents(
+                project_id=project.public_id, delete_bucket=True
+            )
             self.rm_project_user_keys(project=project)
 
             # Delete metadata from project row
@@ -519,7 +523,9 @@ class ProjectStatus(flask_restful.Resource):
 
         try:
             # Deletes files (also commits session in the function - possibly refactor later)
-            RemoveContents().delete_project_contents(project=project, delete_bucket=True)
+            RemoveContents().delete_project_contents(
+                project_id=project.public_id, delete_bucket=True
+            )
             delete_message = f"\nAll files in {project.public_id} deleted"
             self.rm_project_user_keys(project=project)
 
@@ -765,13 +771,27 @@ class RemoveContents(flask_restful.Resource):
             )
 
         # Delete project contents from db and cloud
-        self.delete_project_contents(project=project)
 
+        # Get redis connection to add a job to delete project contents
+        redis_url = flask.current_app.config.get("REDIS_URL")
+        r = Redis.from_url(redis_url)
+        q = Queue(connection=r)
+
+        # Enqueue job to delete project contents
+        job = q.enqueue(self.delete_project_contents, project.public_id)
+
+        # OLD
+        # self.delete_project_contents(project=project)
+
+        # TODO - return job id to client to check status of deletion
         return {"removed": True}
 
     @staticmethod
-    def delete_project_contents(project, delete_bucket=False):
+    def delete_project_contents(project_id, delete_bucket=False):
         """Remove project contents"""
+        # Get project
+        project: models.Project = models.Project.query.filter_by(public_id=project_id).one_or_none()
+
         # Delete from cloud
         with ApiS3Connector(project=project) as s3conn:
             try:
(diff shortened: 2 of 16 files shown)


dds_web/commands.py
```diff
@@ -1319,3 +1321,25 @@ def monitor_usage():
                 body=message,
             )
             dds_web.utils.send_email_with_retry(msg=msg)
+
+
+@click.command("run-worker")
```
reutenauer, line 1326: Since “worker” is rather a generic name, I think we should call this action something else, perhaps `run-broker`.  We don’t name to change the names workers in the code below.

rv0lt: Good idea, maybe run-flask-worker ? I still want to keep that name because it is how the documentation calls it, for consistency shake

dds_web/config.py
```diff
@@ -64,6 +64,9 @@ class Config(object):
         "RATELIMIT_STORAGE_URI", "memory://"
     )  # Use in devel only! Use Redis or memcached in prod
 
+    # Redis
+    REDIS_URL = "redis://dds_redis"  # In production the URL also has a password
```
valyo, line 68: The Redis SVC is `dds-redis`
```suggestion
    REDIS_URL = "redis://dds-redis"  # In production the URL also has a password
```

@rv0lt can you make this change so that we can test on dev instance

valyo (changes requested):

dds_web/config.py
valyo: Ok, I realise it shouldn't be `dds-redis`, but since we cannot have the same URL in all environments, it needs to handled via environment variable

rv0lt: But for dev I think it is better to recreate the secret of the config? It has to be the same URL as RATELIMIT_STORAGE_URI uses there. And last time I checked it contains password details for the redis connection.

This one I setted it up as default due to being the one docker-compose uses, otherwise we would need to make modifications for local testing every time as well

valyo:
I tested on the dev instance, and seemingly things work without errors, but the RQ job remains in Status: queued forever. 
I ran:
`dds data rm -p someunit00040 -a` 
and after ~35 min I can still see the project 
`dds ls -p someunit00040
     ︵
 ︵ (  )   ︵
(  ) ) (  (  )   SciLifeLab Data Delivery System
 ︶  (  ) ) (    https://dev.delivery.scilifelab.se/
      ︶ (  )    CLI Version 2.9.0
          ︶
Current user: unitadmin_1
INFO     Listing files for project 'someunit00040'

 Files / directories in project: someunit00040
 └── test-p8w1t8-17-fpt47.1736531492/
`
and RQ-dashboard shows this:
Screenshot 2025-02-11 at 11 53 23

rv0lt:
hmmmm interesting

> I tested on the dev instance, and seemingly things work without errors, but the RQ job remains in Status: queued forever. I ran: `dds data rm -p someunit00040 -a`
> 
> and after ~35 min I can still see the project
> 
> `dds ls -p someunit00040 ︵ ︵ ( ) ︵ ( ) ) ( ( ) SciLifeLab Data Delivery System ︶ ( ) ) ( https://dev.delivery.scilifelab.se/ ︶ ( ) CLI Version 2.9.0 ︶ Current user: unitadmin_1 INFO Listing files for project 'someunit00040'
> 
> Files / directories in project: someunit00040 └── test-p8w1t8-17-fpt47.1736531492/ `
> 
> and RQ-dashboard shows this:
> Screenshot 2025-02-11 at 11 53 23

valyo:
Now I made another test locally (docker setup, no k8s involved) and I get slightly different, but still unexpected result: 
the job in RQ finishes quite quickly:

Screenshot 2025-02-11 at 14 57 33

but the data is not really deleted:
Screenshot 2025-02-11 at 14 58 49

here's the output from `flask fun-redis-worker`:
```
/code $ flask run-redis-worker
/usr/local/lib/python3.12/site-packages/rq_dashboard/legacy_config.py:29: UserWarning: Configuration option REDIS_URL is deprecated and will be removed in future versions. Please use RQ_DASHBOARD_REDIS_URL instead.
  warnings.warn(
[2025-02-11 13:29:36,731] __init__ [INFO] Logging initiated.
13:29:37 Worker rq:worker:redis-queue started with PID 1860, version 2.1.0
13:29:37 Subscribing to channel rq:pubsub:redis-queue
13:29:37 *** Listening on default...
13:29:37 Cleaning registries for queue: default
13:43:07 Cleaning registries for queue: default
13:54:28 default: dds_web.api.project.RemoveContents.delete_project_contents('someunit00001') (d61f271a-4850-4505-a6c6-abcb5e6ef88e)
[2025-02-11 13:54:28,069] utils [INFO] Safespring location for project 'someunit00001': sto2
13:54:29 Successfully completed dds_web.api.project.RemoveContents.delete_project_contents('someunit00001') job in 0:00:01.911512s on worker redis-queue
13:54:29 default: Job OK (d61f271a-4850-4505-a6c6-abcb5e6ef88e)
13:54:29 Result is kept for 500 seconds
13:54:29 Cleaning registries for queue: default

```

i-oden:
> Ok, we need to add a session.commit. I guess the API call does it somewhere, but being an independent function it wasn't doing it
> 
> > > hm, the bucket is actually empty, but DB is not updated
> > 
> > 
> > indeed, missed that. I was only checking the buckets the whole time



> Ok, we need to add a session.commit. I guess the API call does it somewhere, but being an independent function it wasn't doing it
> 
> > > hm, the bucket is actually empty, but DB is not updated
> > 
> > 
> > indeed, missed that. I was only checking the buckets the whole time

The `db_session` decorator does this

dds_web/api/superadmin_only.py
```diff
@@ -14,6 +14,8 @@
 import flask
 import structlog
 import flask_mail
+from rq import Queue
+from redis import Redis
```
i-oden, line 18: Remove this for now then?

dds_web/api/superadmin_only.py
```diff
@@ -149,6 +151,12 @@ class SendMOTD(flask_restful.Resource):
     @handle_db_error
     def post(self):
         """Send MOTD as email to users."""
+
```
i-oden, line 154: And this too

i-oden (changes requested):
@rv0lt Could you remove the changes relating to the MOTD in this PR? Make a new branch with those changes if you want so that we have it when we start looking at that again

rv0lt:
Ahm, then we could probably add this decorator to the new function instead. 
Preparing a commit with this changes and the other comments

> > > Ok, we need to add a session.commit. I guess the API call does it somewhere, but being an independent function it wasn't doing it
> > > > > hm, the bucket is actually empty, but DB is not updated
> > > > 
> > > > 
> > > > indeed, missed that. I was only checking the buckets the whole time
> > 
> > 
> > > Ok, we need to add a session.commit. I guess the API call does it somewhere, but being an independent function it wasn't doing it
> > > > > hm, the bucket is actually empty, but DB is not updated
> > > > 
> > > > 
> > > > indeed, missed that. I was only checking the buckets the whole time
> > 
> > 
> > The `db_session` decorator does this
> 
> > > Ok, we need to add a session.commit. I guess the API call does it somewhere, but being an independent function it wasn't doing it
> > > > > hm, the bucket is actually empty, but DB is not updated
> > > > 
> > > > 
> > > > indeed, missed that. I was only checking the buckets the whole time
> > 
> > 
> > > Ok, we need to add a session.commit. I guess the API call does it somewhere, but being an independent function it wasn't doing it
> > > > > hm, the bucket is actually empty, but DB is not updated
> > > > 
> > > > 
> > > > indeed, missed that. I was only checking the buckets the whole time
> > 
> > 
> > The `db_session` decorator does this
> 
> So I guess we should remove the decorator from the `RemoveContents.delete` endpoint? The point of it is to catch db exceptions

rv0lt:
> @rv0lt Could you remove the changes relating to the MOTD in this PR? Make a new branch with those changes if you want so that we have it when we start looking at that again

Yeah, I wasn't sure if I should remove it or not, because I started trying to do it with MOTD, and then I changed so kind of forgot about it.  But will move it then!!

dds_web/__init__.py
```diff
@@ -323,6 +328,7 @@ def load_user(user_id):
         app.cli.add_command(send_usage)
         app.cli.add_command(collect_stats)
         app.cli.add_command(monitor_usage)
+        app.cli.add_command(run_redis_worker)
```
valyo, line 331: this should probably not be under "Add flask commands - cronjobs" but under "Add flask commands - general"

valyo (changes requested):
Unless the aim of this PR is purely POC, i.e. to show that it can work with RQ by manually starting a worker, then next thing is to figure out how to start a worker in each container upon deployment

rv0lt:
> Unless the aim of this PR is purely POC, i.e. to show that it can work with RQ by manually starting a worker, then next thing is to figure out how to start a worker in each container upon deployment

Yeah, I am also trying to think about the best way to do that

rv0lt:
> @rv0lt Could you remove the changes relating to the MOTD in this PR? Make a new branch with those changes if you want so that we have it when we start looking at that again

@i-oden I also just realized that I made two tests for **send_email_with_retry** because it was not covered, and I was touching those functions. Should I remove that also?

dds_web/api/project.py
```diff
@@ -746,7 +752,6 @@ class RemoveContents(flask_restful.Resource):
 
     @auth.login_required(role=["Unit Admin", "Unit Personnel"])
     @logging_bind_request
-    @dbsession
```
i-oden, line 749: This also handles any of the db errors in the `delete` endpoint,e.g. `project.current_status`, `project.has_been_available` etc, if we move this decorator from this enpoint, we also need to add a try except:

try: before `project = project_schemas.ProjectRequiredSchema [....]`
except, including everything caught in the decorator: after `if not project.files:`

i-oden (changes requested):

i-oden:
> > @rv0lt Could you remove the changes relating to the MOTD in this PR? Make a new branch with those changes if you want so that we have it when we start looking at that again
> 
> @i-oden I also just realized that I made two tests for **send_email_with_retry** because it was not covered, and I was touching those functions. Should I remove that also?

Leave it for now, I haven't looked at the tests yet

i-oden:
> Unless the aim of this PR is purely POC, i.e. to show that it can work with RQ by manually starting a worker, then next thing is to figure out how to start a worker in each container upon deployment

Imo the POC is that we can get this working in general, so personally I'd see this as a part of the POC.

tests/conftest.py
```diff
@@ -576,3 +577,43 @@ def cli_test_app():
 @pytest.fixture()
 def cli_runner(cli_test_app):
     return cli_test_app.test_cli_runner()
+
+
+@pytest.fixture(scope="session", autouse=True)
+def mock_enqueue():
```
i-oden, line 583: Very nice!

tests/test_utils.py
```diff
@@ -757,6 +757,36 @@ def test_send_project_access_reset_email(client):
...
+
+    # Get email row
+    email_row: models.Email = db.session.query(models.Email).first()
+
+    # Run function
+    with patch("dds_web.utils.mail.send", side_effect=side_effect):
+        response = utils.send_email_with_retry(msg=email_row)
+    assert response is None
+
+
```
i-oden, line 789: Answer to your previous question: Yes, remove this too. BUT put it in another branch, it's great that you've added these tests, we definitely want to keep them, just not in this PR.

i-oden (changes requested):

dds_web/config.py
i-oden, line 68: https://github.com/ScilifelabDataCentre/dds_web/pull/1591#discussion_r1951121103
```suggestion
    RQ_DASHBOARD_REDIS_URL = "redis://dds_redis"  # In production the URL also has a password
```

i-oden (changes requested):

rv0lt:
Okay I have been thinking, and I think the best approach to start this worker in our cluster would be:

- Add a new container in the deployment.yaml (i.e same pod) that has a readiness probe that will check that Flask has been initialized (we could even add easily some health endpoints).
After the app has been initialized, the readiness probe will pass and this new container will run the redis worker.

I think this can easily solve the issue of running this command only after the main app logic, and it won't add unnecessary complexity like the other solutions I was reading about

Thoughts: @valyo @i-oden ?

rv0lt:
Sorry, I meant liveness probe not readiness. So that if the check is not passed, the command will not be executed. My second solution would be to add the code to start the worker in the __init__.py, there is a flag that can be passed to the function to start it in background mode so it won't block the app


> Okay I have been thinking, and I think the best approach to start this worker in our cluster would be:
> 
>     * Add a new container in the deployment.yaml (i.e same pod) that has a readiness probe that will check that Flask has been initialized (we could even add easily some health endpoints).
>       After the app has been initialized, the readiness probe will pass and this new container will run the redis worker.
> 
> 
> I think this can easily solve the issue of running this command only after the main app logic, and it won't add unnecessary complexity like the other solutions I was reading about
> 
> Thoughts: @valyo @i-oden ?

i-oden:
> Sorry, I meant liveness probe not readiness. So that if the check is not passed, the command will not be executed. My second solution would be to add the code to start the worker in the **init**.py, there is a flag that can be passed to the function to start it in background mode so it won't block the app
> 
> > Okay I have been thinking, and I think the best approach to start this worker in our cluster would be:
> > ```
> > * Add a new container in the deployment.yaml (i.e same pod) that has a readiness probe that will check that Flask has been initialized (we could even add easily some health endpoints).
> >   After the app has been initialized, the readiness probe will pass and this new container will run the redis worker.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think this can easily solve the issue of running this command only after the main app logic, and it won't add unnecessary complexity like the other solutions I was reading about
> > Thoughts: @valyo @i-oden ?

As discussed on Slack, we'll talk about this in a bit during the sprint review and see what the others think. Imo though, we should go for initializing the redis worker during the app initialization.

dds_web/__init__.py
```diff
@@ -198,6 +201,10 @@ def create_app(testing=False, database_uri=None):
         # User config file, if e.g. using in production
         app.config.from_envvar("DDS_APP_CONFIG", silent=True)
 
+        # redis dashboard
+        app.config.from_object(rq_dashboard.default_settings)
+        rq_dashboard.web.setup_rq_connection(app)
```
i-oden, line 206: What happens if you move this row and rows 300-311 to within the `with app.app_context():`? 

```suggestion
```

rv0lt: Yeah, moving the ones below work. These ones only init the dashboard

dds_web/api/project.py
```diff
@@ -765,13 +771,25 @@ def delete(self):
...
+
+        # Enqueue job to delete project contents
+        job = q.enqueue(self.delete_project_contents, project.public_id)
+
+        # TODO - return job id to client to check status of deletion
         return {"removed": True}
 
     @staticmethod
-    def delete_project_contents(project, delete_bucket=False):
+    @dbsession
```
valyo, line 787: I am curious how/why this was before working without the decorator

rv0lt: The dbsession ? It doesn't commit the changes if not. So the bucket is empty but the db is not

i-oden:
@rv0lt I think we should have the job status changes in a new PR

rv0lt:
> @rv0lt I think we should have the job status changes in a new PR

Should we create a new subtask then?

dds_web/commands.py
```diff
@@ -1319,3 +1322,30 @@ def monitor_usage():
                 body=message,
             )
             dds_web.utils.send_email_with_retry(msg=msg)
+
+
+@click.command("restart-redis-worker")
```
i-oden, line 1327: I like this.
Question:  What happens to potential queued tasks?

i-oden (changes requested):
This looks really good. Have tried it out, things are working. 
1. Move the job ID code to a new PR 
2. What do we need to change / add in the test/dev/prod environments to get this working?
3. The RQ dashboard -- how will this be protected? Will we (the team) be able to see queued tasks somewhere? We don't want everyone to see everything

dds_web/api/project.py
valyo: Now I see that I was confused by its removal and re-adding for the delete() method. But it doesn't seem that it was present for the delete_project_contents() before this PR. So my question remains for this method

dds_web/commands.py
rv0lt: If there is no worker, they will stay enqueded but unprocessed

valyo: I hope they will be picked up when new worker(s) appear. 
Question: why would we want to restart existing worker(s)?
Question: is there going to be one worker per API/backend replica?

rv0lt:
I am trying a way to have a username and password, the doc is not very complete on that but there is an option. 
> This looks really good. Have tried it out, things are working.
> 
>     1. Move the job ID code to a new PR
> 
>     2. What do we need to change / add in the test/dev/prod environments to get this working?
> 
>     3. The RQ dashboard -- how will this be protected? Will we (the team) be able to see queued tasks somewhere? We don't want everyone to see everything

valyo:
> 3. The RQ dashboard -- how will this be protected? Will we (the team) be able to see queued tasks somewhere? We don't want everyone to see everything
I don't think it needs to be exposed to the users, I mean that it is of any interest for them provided that we handle everything in the client in an user friendly way. 
For us it will be always good to be able to take a look for monitoring/troubleshooting purposes.
The simplest way to protect it is to add IP whitelisting in the ingress we create for it. One can also set up GitHub login (OAuth2) using https://github.com/oauth2-proxy/oauth2-proxy, for example, if it doesn't feel like overkill.

rv0lt:
> > 3. The RQ dashboard -- how will this be protected? Will we (the team) be able to see queued tasks somewhere? We don't want everyone to see everything
> >    I don't think it needs to be exposed to the users, I mean that it is of any interest for them provided that we handle everything in the client in an user friendly way.
> >    For us it will be always good to be able to take a look for monitoring/troubleshooting purposes.
> >    The simplest way to protect it is to add IP whitelisting in the ingress we create for it. One can also set up GitHub login (OAuth2) using https://github.com/oauth2-proxy/oauth2-proxy, for example, if it doesn't feel like overkill.

Is it possible to do whitelist and block for everyone even thought it lives under a subdomain of the main app?

i.e  delivery.scilifelab.se/rq

valyo:
> Is it possible to do whitelist and block for everyone even thought it lives under a subdomain of the main app?
> 
> i.e delivery.scilifelab.se/rq

I believe yes, ingress rules have the `path` element, which we have set to "/" for the main app domain:
```
...
  rules:
  - host: delivery.scilifelab.se
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: dds-backend
            port:
              number: 5000
```

rv0lt:
> > Is it possible to do whitelist and block for everyone even thought it lives under a subdomain of the main app?
> > i.e delivery.scilifelab.se/rq
> 
> I believe yes, ingress rules have the `path` element, which we have set to "/" for the main app domain:
> 
> ```
> ...
>   rules:
>   - host: delivery.scilifelab.se
>     http:
>       paths:
>       - path: /
>         pathType: Prefix
>         backend:
>           service:
>             name: dds-backend
>             port:
>               number: 5000
> ```

then all good!

The changes in the cluster, as per now, are creating this two variables 

    REDIS_URL
    RQ_DASHBOARD_REDIS_URL

They need to have the same value, wich is also the same value as RATELIMIT_STORAGE_URI.  Which is the one that has the username and password of redis

i-oden (dismissed):

valyo (commented):
I don't have any changes to suggest without clarity on my 2 questions, mainly the last one bout the command

dds_web/commands.py
rv0lt: Q1: Only locally when development I can think of, because if you make changes to the code it is not picked up unless restarted

Q2: I haven't tested in a cluster. But I am almost sure that, how it is written right now, it will only register one worker. I can agree that at least one worker per replica would be better

PR #1594, opened 2025-03-17
Enqueue Message Of the Day

rv0lt:

## 1. Description / Summary

Added the queuing method for MOTD. 

The call to the function is almost the same. But now during the handling of the function, the connection object is removed. That is safe because the function to send the email is able to retrieve it if it is not provided.

Adapted tests for the changes

**Is this from a Jira task?** --> https://scilifelab.atlassian.net/jira/software/projects/HMS/boards/13?selectedIssue=HMS-2298&sprintStarted=true

- New feature
- Bug fix --> label: `bug` 
- Security Alert fix 
    - Major version update 

## 4. Additional information

## 5. Actions / Scans

- **Black**

- **Prettier**

- **Yamllint**

- **Tests**

- **CodeQL**

- **Trivy**

- **Snyk**

### DIFF
--- a/SPRINTLOG.md
+++ b/SPRINTLOG.md
@@ -486,4 +486,7 @@ _Nothing merged during this sprint_
 # 2025-03-31 - 2025-04-11
 
 - New version: 2.10.0 ([#1597](https://github.com/ScilifelabDataCentre/dds_web/pull/1597))
+
+# 2025-04-14 - 2025-04-25
+
 - Adapted Message of the Day function to work asynchronous using the Redis Queue. ([#1594](https://github.com/ScilifelabDataCentre/dds_web/pull/1594))
--- a/SPRINTLOG.md
+++ b/SPRINTLOG.md
@@ -489,5 +489,5 @@ _Nothing merged during this sprint_
 
 # 2025-04-14 - 2025-04-25
 
-- Adapted Message of the Day function to work asynchronous using the Redis Queue. ([#1594](https://github.com/ScilifelabDataCentre/dds_web/pull/1594))
+- Adapted Message of the Day function to work asynchronous using the Redis Queue. Set MOTD as a background task([#1594](https://github.com/ScilifelabDataCentre/dds_web/pull/1594))
 - Remove docker-compose.yml version (obsolete) ([#1598](https://github.com/ScilifelabDataCentre/dds_web/pull/1598))
--- a/SPRINTLOG.md
+++ b/SPRINTLOG.md
@@ -482,3 +482,7 @@ _Nothing merged during this sprint_
 # 2024-02-03 - 2025-02-14
 
 - Implement Redis Queue to process some requests asynschronusly and avoid timeouts. Set project deletion as a background task. ([#1591](https://github.com/ScilifelabDataCentre/dds_web/pull/1591))
+
+# 2025-03-17 – 2025-02-28
+
+- Adapted Message of the Day function to work asynchronous using the Redis Queue. ([#1594](https://github.com/ScilifelabDataCentre/dds_web/pull/1594))
--- a/SPRINTLOG.md
+++ b/SPRINTLOG.md
@@ -491,4 +491,3 @@ _Nothing merged during this sprint_
 
 - Adapted Message of the Day function to work asynchronous using the Redis Queue. ([#1594](https://github.com/ScilifelabDataCentre/dds_web/pull/1594))
 - Remove docker-compose.yml version (obsolete) ([#1598](https://github.com/ScilifelabDataCentre/dds_web/pull/1598))
-
--- a/SPRINTLOG.md
+++ b/SPRINTLOG.md
@@ -481,7 +481,7 @@ _Nothing merged during this sprint_
 
 # 2025-02-03 - 2025-02-14
 
-- Implement Redis Queue to process some requests asynschronusly and avoid timeouts. Set project deletion as a background task. ([#1591](https://github.com/ScilifelabDataCentre/dds_web/pull/1591)
+- Implement Redis Queue to process some requests asynschronusly and avoid timeouts. Set project deletion as a background task. ([#1591](https://github.com/ScilifelabDataCentre/dds_web/pull/1591))
 
 # 2025-03-31 - 2025-04-11
 
--- a/SPRINTLOG.md
+++ b/SPRINTLOG.md
@@ -489,5 +489,5 @@ _Nothing merged during this sprint_
 
 # 2025-04-14 - 2025-04-25
 
-- Adapted Message of the Day function to work asynchronous using the Redis Queue. Set MOTD as a background task([#1594](https://github.com/ScilifelabDataCentre/dds_web/pull/1594))
+- Set MOTD as a background task([#1594](https://github.com/ScilifelabDataCentre/dds_web/pull/1594))
 - Remove docker-compose.yml version (obsolete) ([#1598](https://github.com/ScilifelabDataCentre/dds_web/pull/1598))
--- a/dds_web/api/superadmin_only.py
+++ b/dds_web/api/superadmin_only.py
@@ -173,7 +173,10 @@ class SendMOTD(flask_restful.Resource):
 
         BaseUser = models.UnitUser if unit_only else models.User
         users_to_send = (
-            db.session.query(BaseUser.username, models.Email.email).outerjoin(models.Email).all()
+            db.session.query(BaseUser.username, models.Email.email)
+            .outerjoin(models.Email)
+            .filter(models.Email.primary == True)
+            .all()
         )
 
         # Get redis connection to add a job to delete project contents
--- a/dds_web/api/superadmin_only.py
+++ b/dds_web/api/superadmin_only.py
@@ -173,7 +173,7 @@ class SendMOTD(flask_restful.Resource):
 
         BaseUser = models.UnitUser if unit_only else models.User
         users_to_send = (
-            db.session.query(BaseUser.username, models.Email.email).join(models.Email).all()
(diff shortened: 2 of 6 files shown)


i-oden:
@rv0lt will have a look and review but change the label to one of the labels mentioned in section 3 in the description 🙏🏻

valyo (changes requested):

i-oden:
@rv0lt There should also be one of the following labels: 
- `breaking`
- `feature`
- `bug`
- `dependency`
- `skip-changelog`

dds_web/api/superadmin_only.py
```diff
@@ -179,40 +181,44 @@ def post(self):
...
-                    headers=[
-                        ["Content-ID", "<Logo>"],
-                    ],
+        # Get redis connection to add a job to delete project contents
+        redis_url = flask.current_app.config.get("REDIS_URL")
+        r = Redis.from_url(redis_url)
+        q = Queue(connection=r)
+
+        # Email users
+        for user in utils.page_query(users_to_send):
```
i-oden, line 190: Question: Is there a maximum number of "tasks" that can be placed in a queue? Or will this be fine even with hundreds or thousands of emails?

dds_web/api/superadmin_only.py
```diff
@@ -179,40 +181,44 @@ def post(self):
...
+                "inline",
+                headers=[
+                    ["Content-ID", "<Logo>"],
+                ],
+            )
 
-                # This funcion cannot be enqued because the connection object is not pickable
-                utils.send_email_with_retry(msg=msg, obj=conn)
+            # Enqueue function to send the email
+            job = q.enqueue(utils.send_email_with_retry, msg)
```
i-oden, line 214: The `with mail.connect() as conn:` part that we had previously makes sure to only set up the connection once. I don't know if that was the best way to do it but are we sure that this way is not "spamming" the email server or something?

dds_web/api/superadmin_only.py
```diff
@@ -179,40 +181,44 @@ def post(self):
...
-        if unit_only:
-            return_msg += "sent to unit personnel only."
-        else:
-            return_msg += "sent to all users."
+            if unit_only:
+                recipients = "unit personnel"
+            else:
+                recipients = "all users"
+
+        return_msg = f"Email notification #{motd_id} is scheduled for delivery to {recipients}"
```
i-oden, line 221: "Email notification #1" made me think that it was the first email that had been sent etc, so I think we should change it to say that the "MOTD #1" (for example) has been scheduled
```suggestion
        return_msg = f"MOTD #{motd_id} is scheduled for delivery to {recipients}"
```

i-oden (changes requested):

dds_web/api/superadmin_only.py
rv0lt: The documentation does not refer to a maximum. But some old PRs in their repo talk about being done performance testing. 

Unprocessed tasks will be enqueded until they are picked up by some worker

dds_web/api/superadmin_only.py
rv0lt: mmm Maybe. 
The current implementation send mails 1 by 1. And every job is a subprocess at the system level. So it would have to create a new connection anyway.

Another alt would be to have batches of emails by every job but that would be substancially more changes

i-oden: @valyo thoughts on this?

i-oden:
@rv0lt is this actually breaking? Breaking means that the CLI and the API will stop working together (new version of API doesn't work with old version of CLI, old version of API doesn't work with new version of CLI, etc), which is not the case, right? This should probably be `feature`?

rv0lt:
I see, I was considering all the Queuing tasks like that because of the system working in a very different way. But it's true that in this case it should be backwards compatbible 

> @rv0lt is this actually breaking? Breaking means that the CLI and the API will stop working together (new version of API doesn't work with old version of CLI, old version of API doesn't work with new version of CLI, etc), which is not the case, right? This should probably be `feature`?

i-oden (changes requested):
Decision from sprint review: 
1. Check if new connections for each email is a problem
2. If you find information that specifically says that it's no problem --> we can keep it like this for now (maybe, let's discuss)
3. If you cannot find info regarding this or that it is a problem --> move code to a new function called e.g. "email all users" and start the email connection in that function (because the conn object cannot be passed to the queue) and then queue the new function instead

valyo (approved):
works as expected and looks good

i-oden (changes requested):
want to review

SPRINTLOG.md
```diff
@@ -481,8 +481,12 @@ _Nothing merged during this sprint_
 
 # 2025-02-03 - 2025-02-14
 
-- Implement Redis Queue to process some requests asynschronusly and avoid timeouts. Set project deletion as a background task. ([#1591](https://github.com/ScilifelabDataCentre/dds_web/pull/1591))
+- Implement Redis Queue to process some requests asynschronusly and avoid timeouts. Set project deletion as a background task. ([#1591](https://github.com/ScilifelabDataCentre/dds_web/pull/1591)
```
i-oden, line 484: This parenthesis should still be there
```suggestion
- Implement Redis Queue to process some requests asynschronusly and avoid timeouts. Set project deletion as a background task. ([#1591](https://github.com/ScilifelabDataCentre/dds_web/pull/1591))
```

dds_web/api/superadmin_only.py
```diff
@@ -168,51 +170,32 @@ def post(self):
         unit_only: bool = request_json.get("unit_only", False)
         if not isinstance(unit_only, bool):
             raise ddserr.DDSArgumentError(message="The 'unit_only' argument must be a boolean.")
-        if unit_only:
-            users_to_send = db.session.query(models.UnitUser)
-        else:
-            users_to_send = db.session.query(models.User)
+
+        BaseUser = models.UnitUser if unit_only else models.User
+        users_to_send = (
```
i-oden, line 175: This will send to all emails, not only the primary one. Atm we only have one email per person, but it is technically possible to have multiple emails for one account, so there's a risk that we'd forget that this code existed and then started spamming all emails 

EDIT: I saw the code sending the actual emails and you do check that it's the primary email there.

i-oden: Also, what happens if this variable contains thousands of users, will it handle that? I mean partly the variable but also the query itself.

dds_web/api/superadmin_only.py
```diff
@@ -168,51 +170,32 @@ def post(self):
...
+        # Possible improvement: Divide the users into smaller chunks to avoid sending too many emails at once
+        job = q.enqueue(utils.send_motd_to_user_list, users_to_send, subject, body, html)
 
-                # This funcion cannot be enqued because the connection object is not pickable
-                utils.send_email_with_retry(msg=msg, obj=conn)
-
-        return_msg = f"MOTD '{motd_id}' has been "
         if unit_only:
-            return_msg += "sent to unit personnel only."
+            recipients = "unit personnel"
```
i-oden, line 194: I like that you changed the previous if statement above, so I think we should do that here too:

```
recipients = "unit personnel" if unit_only else "all users"
```

dds_web/utils.py
```diff
@@ -501,6 +501,36 @@ def send_email_with_retry(msg, times_retried=0, obj=None):
             send_email_with_retry(msg, times_retried=retry, obj=obj)
 
 
+def send_motd_to_user_list(users_to_send, subject, body, html):
+    """Send MOTD email to a list of users."""
+
+    with mail.connect() as conn:
+        for user in users_to_send:  # ('unituser_1', 'unituser1@mailtrap.io')
+            username = user[0]
+            primary_email = user[1]
```
i-oden, line 510: Are we sure that the primary email is the second item? What happens if there are multiple emails for one user, is this item the primary?

dds_web/utils.py
```diff
@@ -501,6 +501,36 @@ def send_email_with_retry(msg, times_retried=0, obj=None):
             send_email_with_retry(msg, times_retried=retry, obj=obj)
 
 
+def send_motd_to_user_list(users_to_send, subject, body, html):
```
i-oden, line 504: Nice, I like this implementation

i-oden (changes requested):
Nice, I like this, and it works as expected. Have a few questions / suggestions though

dds_web/api/superadmin_only.py
rv0lt: So, the context is:

The previous query object cannot be used in the RQ because it requires an active request context, that's why the only solution is to fetch the email's before processing. That's the tradeoff, I guess.

dds_web/utils.py
rv0lt: Yeah, I just tested locally, and it will send 1 email for every email, even the no primary. But they can be filtered out if you want thought.

dds_web/api/superadmin_only.py
i-oden: Yeah, understand this part, I was mostly wondering if it's necessary to do the querying of the users and scheduling in "chunks" or if getting all users in on single query and sending it to the queuing is fine.

dds_web/utils.py
i-oden: Yeah let's filter out the non-primary emails.

dds_web/api/superadmin_only.py
rv0lt: Yeah it could be done, I am not sure either what's the limit

i-oden: Let's just keep it as it is for now and change it if we encounter problems

i-oden (approved):
Looks good and works as expected

PR #1595, opened 2025-03-27
Set project archivation as a background task

rv0lt:

## 1. Description / Summary

Implement archiving operation as a queue for both the endpoint and the flask command.

The tests are adapted, and some new tests have been added.

A flag is_queue_operation is added to the main function in order to generate the return messages. 

Two new functions are added that perform the 1) Deleting project contents and 2) Updating DB status
These functions are called by jobs to be executed in the queue.

**Is this a GitHub issue?** --> Add the link to the github issue

**Is this from a Jira task?** --> If your branch does not contain info regarding the Jira task ID, put it here.

- New feature
- Bug fix --> label: `bug` 
- Security Alert fix 
    - Major version update 

## 4. Additional information

## 5. Actions / Scans

- **Black**

- **Prettier**

- **Yamllint**

- **Tests**

- **CodeQL**

- **Trivy**

- **Snyk**

### DIFF
--- a/SPRINTLOG.md
+++ b/SPRINTLOG.md
@@ -499,8 +499,8 @@ _Nothing merged during this sprint_
 
 - Bump jinja2 from 3.1.3 to 3.1.6 ([#1606](https://github.com/ScilifelabDataCentre/dds_web/pull/1606))
 - Bump werkzeug from 3.0.3 to 3.0.6 ([#1607](https://github.com/ScilifelabDataCentre/dds_web/pull/1607))
-- Add queuing for archiving projects([#1595](https://github.com/ScilifelabDataCentre/dds_web/pull/1595))
 
 # Summer 2025
 
 - Rename 'doc' directory to 'docs' for consistency ([#1614](https://github.com/ScilifelabDataCentre/dds_web/pull/1614))
+- Add queuing for archiving projects([#1595](https://github.com/ScilifelabDataCentre/dds_web/pull/1595))


i-oden (changes requested):
Want to have  look

i-oden (changes requested):

dds_web/api/project.py
```diff
@@ -204,9 +190,16 @@ def post(self):
...
+            # Return messages to user
+            if is_queue_operation:
+                return_message = f"{project.public_id} has started, in the background, the process of being updated to {new_status}"
+                return_message += " (Aborted)" if new_status == "Archived" and is_aborted else ""
+                return_message += ". It may take some time to complete"
+            else:
+                return_message = f"{project.public_id} updated to status {new_status}"
+                return_message += " (Aborted)" if new_status == "Archived" and is_aborted else ""
+
+            return_message += " (Aborted)" if new_status == "Archived" and is_aborted else ""
```
i-oden, line 202: this could be moved to the `if not is_queue_operation` on line 183 I believe, so that we don't need another if statement. My suggestion is to put the "contents" of `else` under `if not is_queue_operation` and the "contents" of `if is_queue_operaton` under a new `else` under the if statement. So that it would become like this from line 183:

```python
           # If operations was not handled by a queue, commit the changes
            if not is_queue_operation:
                self.update_status_project(project=project, new_status_row=new_status_row)
                return_message = f"{project.public_id} updated to status {new_status}"
                return_message += " (Aborted)" if new_status == "Archived" and is_aborted else
           else:
                return_message = f"{project.public_id} has started, in the background, the process of being updated to {new_status}"
                return_message += " (Aborted)" if new_status == "Archived" and is_aborted else ""
                return_message += ". It may take some time to complete"
```

Please let me know if I'm missing something here though

i-oden (changes requested):

dds_web/api/project.py
rv0lt: Yeah you are right, I didn't notice I already had the branch-if before

dds_web/api/project.py
```diff
@@ -522,20 +518,47 @@ def archive_project(
...
+            clear_proj_info=aborted,  # If aborted, clear project info
+        )
+
+        job_update_db = q.enqueue(
+            self.queue_helper_fun_update_db,
+            project_id=project.public_id,
+            current_time=current_time,
+            new_status="Archived",
+            aborted=aborted,
+            depends_on=job_delete_contents,  # This job is only executed after success of delete contents
```
i-oden, line 538: Nice

dds_web/api/project.py
```diff
@@ -522,20 +518,47 @@ def archive_project(
...
             )
-            delete_message = f"\nAll files in {project.public_id} deleted"
             self.rm_project_user_keys(project=project)
 
             # Only mark as inactive after all deletion operations succeed
             project.is_active = False
             # Delete metadata from project row
-            if aborted:
+            if clear_proj_info:
                 project = self.delete_project_info(project)
```
i-oden, line 561: This is up for discussion but I feel like this should be included in the database update function instead? If the  `queue_helper_fun_update_db` should only be for the status change then we should change that name to something else

i-oden (changes requested):

tests/api/test_project.py
```diff
@@ -536,8 +536,53 @@ def test_projectstatus_archived_project(module_client, boto3_session):
...
+        response = module_client.post(
+            tests.DDSEndpoint.PROJECT_STATUS,
+            headers=token,
+            query_string={"project": project_id},
+            json=new_status,
+        )
+        assert response.status_code == http.HTTPStatus.INTERNAL_SERVER_ERROR
+
+    _, err = capfd.readouterr()
+    assert "DatabaseError" in err
```
i-oden, line 581: I'd like to have a more detailed check here for the error message we're expecting.

rv0lt: Yeah, but that is because we write the custom error message in **delete_project_contents** 

In this case, it fails during commit in **dbsession** function which does not have a custom message. 

I could change it and write a try catch so we can raise a custom DeletionError

dds_web/api/project.py
```diff
@@ -169,33 +170,27 @@ def post(self):
...
+            if not is_queue_operation:
+                # If operations was not handled by a queue, commit the changes
+                self.update_project_status(project=project, new_status_row=new_status_row)
+                return_message += f"Project {project.public_id} updated to status {new_status}"
+            else:
+                return_message += (
+                    f"Project {project.public_id} has started, in the background, the process of being updated to {new_status}"
+                    + (" (Aborted)" if is_aborted else "")
+                    + ". It may take some time to complete"
                 )
```
i-oden, line 193: I'd like to change the message to this, it makes it clearer. Also, the aborted part is removed here since aborted is always a substatus of archived and archived is always queued operation.
```suggestion
                return_message += (
                    f"The status of project {project.public_id} is being updated to {new_status}. The DDS is handling this in the background.  It may take some time to complete."
                )
```

i-oden (changes requested):
I'm pretty close to approving this now I think, just testing it out a bit etc

i-oden (commented):
Note to self: There are new tests, but the old ones are not changed (apart from a new argument that is required) and all still pass.

tests/api/test_project.py
i-oden: I think we can have a think about this but I see what you mean. We can leave it as it is for now.

dds_web/api/project.py
```diff
@@ -169,33 +170,23 @@ def post(self):
...
-                    ),
-                ) from err
+            return_message = ""
+
+            if not is_queue_operation:
+                # If operations was not handled by a queue, commit the changes
+                self.update_project_status(project=project, new_status_row=new_status_row)
+                return_message += f"Project {project.public_id} updated to status {new_status}"
+            else:
+                return_message += f"The status of project {project.public_id} is being updated to {new_status}. The DDS is handling this in the background.  It may take some time to complete."
```
i-oden, line 189: I don't believe this line is tested, that we know it does print this, right?

i-oden: We have a bunch of tests for when errors happen, but I'd like to have a test that makes sure that specific status changes returns one thing and doesn't get added to a queue and that archiving returns another thing and does get added to the queue

i-oden: During live review: 
- The queuing etc is already indirectly tested, which is good, but I'd like us to add specific assertions to existing tests to verify that they are added to the queue / not added to the queue etc. Each project status transition should be tested like this.
- I'd like us to verify that this specific return_message is printed / returned, and also the same with the change in the `commands.py`

SPRINTLOG.md
```diff
@@ -499,6 +499,7 @@ _Nothing merged during this sprint_
 
 - Bump jinja2 from 3.1.3 to 3.1.6 ([#1606](https://github.com/ScilifelabDataCentre/dds_web/pull/1606))
 - Bump werkzeug from 3.0.3 to 3.0.6 ([#1607](https://github.com/ScilifelabDataCentre/dds_web/pull/1607))
+- Add queuing for archiving projects([#1595](https://github.com/ScilifelabDataCentre/dds_web/pull/1595))
```
i-oden, line 502: Could you move this to the bottom of the file? 🙏🏻 Under Summer 2025

rv0lt: done!

PR #1605, opened 2025-05-06
Move project inactivation to after data-removal in order to ensure projects cannot be inactive while still containing data

valyo:

## 1. Description / Summary

**Add a summary here**: These changes should address the issue of failed project content deletion but project being marked as  inactive. 
The `is_active = False` transaction should be rolled back if an exception gets thrown even if that line is before the try-catch; the only possible scenarion for not being rolled back is when the exception gets thrown by the S3 related code but `db.session.rollback()` fails due to DB issue happening exactly at that time.

**Is this from a Jira task?** --> HMS-2391

- New feature
- Bug fix --> label: `bug` 
- Security Alert fix 
    - Major version update 

## 4. Additional information

## 5. Actions / Scans

- **Black**

- **Prettier**

- **Yamllint**

- **Tests**

- **CodeQL**

- **Trivy**

- **Snyk**

### DIFF
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -485,11 +485,11 @@ class ProjectStatus(flask_restful.Resource):
             )
             self.rm_project_user_keys(project=project)
 
+            # Only mark as inactive after all deletion operations succeed
+            project.is_active = False
             # Delete metadata from project row
             self.delete_project_info(proj=project)
 
-            # Only mark as inactive after all deletion operations succeed
-            project.is_active = False
         except (TypeError, DatabaseError, DeletionError, BucketNotFoundError) as err:
             flask.current_app.logger.exception(err)
             db.session.rollback()
@@ -530,13 +530,13 @@ class ProjectStatus(flask_restful.Resource):
             delete_message = f"\nAll files in {project.public_id} deleted"
             self.rm_project_user_keys(project=project)
 
+            # Only mark as inactive after all deletion operations succeed
+            project.is_active = False
             # Delete metadata from project row
             if aborted:
                 project = self.delete_project_info(project)
                 delete_message += " and project info cleared"
 
-            # Only mark as inactive after all deletion operations succeed
-            project.is_active = False
         except (TypeError, DatabaseError, DeletionError, BucketNotFoundError) as err:
             flask.current_app.logger.exception(err)
             db.session.rollback()
--- a/dds_web/api/project.py
+++ b/dds_web/api/project.py
@@ -477,7 +477,6 @@ class ProjectStatus(flask_restful.Resource):
                 "You cannot delete a project that has been made available previously. "
                 "Please abort the project if you wish to proceed."
             )
-        project.is_active = False
 
         try:
             # Deletes files (also commits session in the function - possibly refactor later)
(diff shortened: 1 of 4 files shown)


dds_web/api/project.py
```diff
@@ -486,8 +485,11 @@ def delete_project(self, project: models.Project, current_time: datetime.datetim
             )
```
i-oden, line 485: Side note: I started looking at delete_project_contents which then led me to the `remove_bucket_contents` in `api_s3_connector.py`. I checked the `bucket.objects.all().delete()` documentation and it says this regarding `delete()`

> The request can contain a list of up to 1,000 keys that you want to delete.

Since `all()` can definitely return more than 1000 objects in some cases (although the risk isn't as likely atm since most are zipping / taring the files) we need to add a new task in the backlog for improving this so that it can handle any number of objects. Also, I'm currently not sure what happens if the number of objects is more than 1000, and if this could potentially be causing the timeouts / issues in general as well.

i-oden: Possibly adding https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/bucket/objects.html#page_size would do

dds_web/api/project.py
```diff
@@ -486,8 +485,11 @@ def delete_project(self, project: models.Project, current_time: datetime.datetim
...
             self.rm_project_user_keys(project=project)
 
+            # Only mark as inactive after all deletion operations succeed
+            project.is_active = False
```
i-oden, line 489: Would be good to add a test that mocks `delete_project_contents`, and one that mocks `rm_project_user_keys`; They should fail and we want to test that the `project.is_active` is still `True`

i-oden (changes requested):
Looks good, nice with a simple change. Would like to add tests for this though, see comment

dds_web/api/project.py
i-oden: https://scilifelab.atlassian.net/browse/HMS-2396

tests/api/test_project.py
```diff
@@ -1538,6 +1541,92 @@ def mock_bucketnotfounderror():
...
+    response = module_client.post(
+        tests.DDSEndpoint.PROJECT_CREATE,
+        headers=tests.UserAuth(tests.USER_CREDENTIALS["unituser"]).token(module_client),
+        json=proj_data,
+    )
+    assert response.status_code == http.HTTPStatus.OK
+    project_id = response.json.get("project_id")
+    project = project_row(project_id=project_id)
+
+    def mock_deletionerror():
```
i-oden, line 1564: Would be good if we could put these outside of the test, at the top of the file, so that more tests could use the same thing. But I remember that being a problem sometimes due to circular imports or something so if it doesn't work then it doesn't. Also I think some of these errors are already implemented somewhere in the test files, but maybe you encountered the circular imports there too

i-oden (changes requested):

i-oden:
Don't forget the sprintlog

i-oden:
@valyo Re: the `skip-changelog` label, could you explain? I think it should be `type: bug` but I may be missing something

valyo:
> @valyo Re: the `skip-changelog` label, could you explain? I think it should be `type: bug` but I may be missing something

That's right, I don't know what I was thinking

PR #1813, opened 2026-03-20
Reduce DB contention: skip project UPDATE when registering new files

valyo:
### Before Marking as Ready for Review

- [x] Add relevant information to the sections below ([Summary](#summary) etc)
- [x] Rebase or merge the latest `dev` (or other targeted branch)
- [x] Update documentation if needed
- [x] Add an entry to the [`SPRINTLOG.md`](https://github.com/ScilifelabDataCentre/dds_web/blob/dev/SPRINTLOG.md) if needed
- [x] Choose an appropriate label. See [here](https://github.com/ScilifelabDataCentre/dds_web/blob/dev/docs/procedures/labelling_a_pull_request.md) for information on the labelling options
- [x] The code follows the [style guidelines](https://github.com/ScilifelabDataCentre/dds_web/blob/dev/docs/procedures/style_guidelines.md)
- [x] Perform a self-review: read the diff as if reviewing someone else's code
- [x] I have commented the code, particularly in hard-to-understand areas
- [x] Verify that all checks and tests have passed

**If the target branch is `master`**:

- [ ] Read and follow [the release instructions](https://github.com/ScilifelabDataCentre/dds_web/blob/dev/docs/procedures/new_release.md)

### Summary

Removes the `UPDATE projects` row write that previously fired on every
`POST /file/new`, eliminating the per-file lock contention that
serialized concurrent file inserts on the same project.

#### Problem

`POST /file/new` registers one file in the database. Internally, the
`NewFileSchema` attached the new file to the project via collection
appends:

```python
project.file_versions.append(new_version)
project.files.append(new_file)
```

SQLAlchemy treats those appends as mutations of the parent `Project`
row, so each call issued an `UPDATE projects ...` alongside the
`INSERT files` and `INSERT versions`. With many parallel uploads in a
single `dds data put` run, every `POST /file/new` had to acquire a
write lock on the same project row, serializing what should have been
independent inserts and slowing large uploads noticeably.

#### Fix

Set the foreign keys directly instead of going through the parent's
collection:

```python
new_file.project_id = project.id
new_version.project_id = project.id
new_file.versions.append(new_version)
```

The project row is no longer dirtied, so no `UPDATE projects` is
emitted and concurrent file inserts no longer contend on its row lock.

Because the file is no longer attached to the session via the
project-cascade, `api/files.py` adds it explicitly:

```python
db.session.add(new_file)
```

The version is still attached via `new_file.versions.append(new_version)`
through the File→Version cascade.

#### New endpoint: `POST /proj/upload/complete`

The project row's `date_updated` (and `last_updated_by`) used to be
refreshed as a side-effect of the per-file `UPDATE projects`. With
that gone, project metadata would otherwise stop reflecting upload
activity. To preserve that signal — but at one write per batch instead
of one per file — this PR adds:

`POST /proj/upload/complete?project=<public_id>`

The CLI calls it once at the end of `dds data put` when at least one
file was registered. Auth: `Unit Admin` and `Unit Personnel` (matching
the rest of the upload surface — Project Owners are not upload-eligible).

#### Backwards compatibility

Older CLIs continue to work unchanged; they just no longer call
`/proj/upload/complete`, which means `project.date_updated` will not
tick during their upload runs until they upgrade. Once the matching
CLI version is published, all timestamps catch up automatically.

#### Tests

- `test_proj_upload_complete_updates_timestamp` (v1 and v3): happy-path
  test that the new endpoint refreshes `date_updated` and
  `last_updated_by`.
- `test_init.py`: `/proj/upload/complete` blocked under maintenance,
  and 403 for super-admin-without-role coverage.
- Existing `test_new_file` continues to verify the schema change
  end-to-end.

#### What to look at in review

- `dds_web/api/schemas/file_schemas.py` — the schema mutation
  is the heart of the change.
- `dds_web/api/files.py` — the explicit `db.session.add(new_file)`
  must be in place; without it, removing the collection append would
  silently drop the file.
- `dds_web/api/project.py` — the new resource class.

### Related Issue/Ticket

[dds_cli #942](https://github.com/ScilifelabDataCentre/dds_cli/pull/942)

### Testing

_If applicable: How did you verify the change? Include commands, data, or screenshots._

### Reviewer Notes

_Anything that helps reviewers (e.g. areas needing close attention)._

---

Once all boxes are checked, mark the PR as **Ready for Review** and tag at least one team member as the initial reviewer.

### DIFF
--- a/dds_web/api/files.py
+++ b/dds_web/api/files.py
@@ -87,10 +87,11 @@ class NewFile(flask_restful.Resource):
         # Verify that project has correct status for upload
         check_eligibility_for_upload(status=project.current_status)
 
-        # Create new files
+        # Create new files (new_file is not attached via project.files so add explicitly)
         new_file = file_schemas.NewFileSchema().load(
             {**flask.request.get_json(silent=True), "project": project.public_id}
         )
+        db.session.add(new_file)
 
         try:
             db.session.commit()
--- a/dds_web/api/files.py
+++ b/dds_web/api/files.py
@@ -177,11 +177,12 @@ class NewFile(flask_restful.Resource):
                 size_stored=file_info.get("size_processed"),
                 time_uploaded=new_timestamp,
                 active_file=existing_file.id,
-                project_id=project,
+                project_id=project.id,  # explicit integer FK, not the ORM object
             )
 
-            # Update foreign keys and relationships
-            project.file_versions.append(new_version)
+            # Set FK directly so we do not modify the project row (avoids UPDATE on projects
+            # and reduces lock contention during concurrent PUT /file/new).
+            new_version.project_id = project.id
             existing_file.versions.append(new_version)
 
             db.session.add(new_version)
--- a/dds_web/api/files.py
+++ b/dds_web/api/files.py
@@ -177,7 +177,6 @@ class NewFile(flask_restful.Resource):
                 size_stored=file_info.get("size_processed"),
                 time_uploaded=new_timestamp,
                 active_file=existing_file.id,
-                project_id=project,
             )
 
             # Set FK directly so we do not modify the project row (avoids UPDATE on projects
--- a/dds_web/api/files.py
+++ b/dds_web/api/files.py
@@ -177,7 +177,7 @@ class NewFile(flask_restful.Resource):
                 size_stored=file_info.get("size_processed"),
                 time_uploaded=new_timestamp,
                 active_file=existing_file.id,
-                project_id=project.id,  # explicit integer FK, not the ORM object
+                project_id=project,
             )
 
             # Set FK directly so we do not modify the project row (avoids UPDATE on projects
--- a/dds_web/api/schemas/file_schemas.py
+++ b/dds_web/api/schemas/file_schemas.py
@@ -140,9 +140,10 @@ class NewFileSchema(project_schemas.ProjectRequiredSchema):
         )
 
         project = data.get("project_row")
-        # Update foreign keys
-        project.file_versions.append(new_version)
-        project.files.append(new_file)
+        # Set FKs directly so we do not modify the project row (avoids UPDATE on projects
+        # and reduces lock contention during concurrent POST /file/new).
+        new_file.project_id = project.id
+        new_version.project_id = project.id
         new_file.versions.append(new_version)
 
         return new_file
--- a/tests/tests_v3/test_files_new.py
+++ b/tests/tests_v3/test_files_new.py
@@ -392,6 +392,28 @@ def test_new_file_checksum_none(client):
     )
 
 
+def test_proj_upload_complete_updates_timestamp(client):
+    """POST /proj/upload/complete refreshes date_updated and last_updated_by."""
+    project_1 = project_row(project_id="file_testing_project")
+    assert project_1
+    db.session.refresh(project_1)
+    before = project_1.date_updated
+
+    response = client.post(
+        tests.DDSEndpoint.PROJ_UPLOAD_COMPLETE,
+        headers=tests.UserAuth(tests.USER_CREDENTIALS["unitadmin"]).token(client),
+        query_string={"project": "file_testing_project"},
+    )
+    assert response.status_code == http.HTTPStatus.OK
+    assert response.json.get("message") == "Project upload timestamp updated."
+
+    db.session.refresh(project_1)
+    assert project_1.date_updated is not None
+    if before is not None:
+        assert project_1.date_updated >= before
+    assert project_1.last_updated_by == "unitadmin"
+
+
 def test_new_file(client):
     """Add and overwrite file to database."""
 
--- a/tests/tests_v3/test_files_new.py
+++ b/tests/tests_v3/test_files_new.py
@@ -465,6 +465,44 @@ def test_new_file(client):
     assert f"File '{updated_file['name']}' updated in db." in response.json["message"]
 
 
+def test_new_file_put_version_linked_to_file_and_project(client):
+    """PUT /file/new creates a Version row tied to both the file and the project."""
+    project_1 = project_row(project_id="file_testing_project")
+    assert project_1
+
+    # First add the file
+    response = client.post(
+        tests.DDSEndpoint.FILE_NEW,
+        headers=tests.UserAuth(tests.USER_CREDENTIALS["unitadmin"]).token(client),
+        query_string={"project": "file_testing_project"},
+        json=FIRST_NEW_FILE,
+    )
+    assert response.status_code == http.HTTPStatus.OK
+
+    # Overwrite the file via PUT
+    updated_file = {**FIRST_NEW_FILE, "size": 1200, "size_processed": 600}
+    response = client.put(
+        tests.DDSEndpoint.FILE_NEW,
+        headers=tests.UserAuth(tests.USER_CREDENTIALS["unitadmin"]).token(client),
+        query_string={"project": "file_testing_project"},
+        json=updated_file,
+    )
+    assert response.status_code == http.HTTPStatus.OK
+
+    # Fetch the file row and verify the new version is linked correctly
+    file_row = models.File.query.filter_by(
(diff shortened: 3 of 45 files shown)


i-oden:
Looking at this now (sorry it's taking time) and so far I don't see a big issue and I think you have found smart solutions. I will have a more thorough look though. 

One thing I'd like though first is for a more detailed description of the problem. So what is the PR trying to do and why? I know you have mentioned it a bit but could you write it a bit more clearly in case someone aside from us reads this at some point? Should basically just need a little restructuring and rephrasing.

valyo:
> Looking at this now (sorry it's taking time) and so far I don't see a big issue and I think you have found smart solutions. I will have a more thorough look though.
> 
> One thing I'd like though first is for a more detailed description of the problem. So what is the PR trying to do and why? I know you have mentioned it a bit but could you write it a bit more clearly in case someone aside from us reads this at some point? Should basically just need a little restructuring and rephrasing.

big thanks!
I have updated the summary now

dds_web/api/schemas/file_schemas.py
```diff
@@ -140,9 +140,10 @@ def return_items(self, data, **kwargs):
         )
 
         project = data.get("project_row")
-        # Update foreign keys
-        project.file_versions.append(new_version)
-        project.files.append(new_file)
+        # Set FKs directly so we do not modify the project row (avoids UPDATE on projects
+        # and reduces lock contention during concurrent POST /file/new).
+        new_file.project_id = project.id
+        new_version.project_id = project.id
```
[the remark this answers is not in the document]
valyo: To clarify, the race actually existed pre-PR too: the UPDATE projects from the cascade serializes commits but not the SELECT-vs-INSERT ordering, so two workers can both pass verify_file_not_exists and both insert under REPEATABLE READ. This PR widens the race rather than introducing it.
Either way, the right fix is a DB-level UNIQUE (project_id, name) on files, with the schema-level precheck kept as a fast path and a try/except IntegrityError → FileExistsError at commit. Happy to do this in a follow-up PR or fold it in here — let me know which is preferred. Need to confirm "delete" semantics first (hard vs soft) since that affects whether a partial constraint is needed.

i-oden (commented):
Sorry for the long wait again. I've looked again now and I've tested it out a bit. Most of it acted like I was expecting but one thing didn't. 

I first tested `dds_web/dev` with `dds_cli/dev` (the baseline) and the `projects.last_updated` was updated when there's a file upload and a file update. 

>>>
Then I tested this branch with `dds_cli/dev` and the `projects.last_updated` was **not** updated when there's a file upload (as expected) but **it's updated when there's a file update i.e. when e.g. `overwrite` flag is used**
<<<

Then I also ofc tested this branch with the `dds_cli/improve-project-updating...` branch and that seems to work regarding updating the `projects.last_updated` at the end. 

Only using the specific column here because that's where I think we really notice it at all. 

--- 

So what I'm wondering is if there should be some change in the `--overwrite` process here too?

dds_web/api/files.py
```diff
@@ -179,8 +180,9 @@ def put(self):
                 project_id=project,
             )
 
-            # Update foreign keys and relationships
-            project.file_versions.append(new_version)
+            # Set FK directly so we do not modify the project row (avoids UPDATE on projects
+            # and reduces lock contention during concurrent PUT /file/new).
+            new_version.project_id = project.id
```
i-oden, line 184: Not a blocker: 

We don't need both 
```python
            new_version = models.Version(
               [...]
                project_id=project,
            )
```
above and 
```python
            new_version.project_id = project.id
```

Not a huge deal, but might be confusing in the future. 

There are 2 alternatives I think (I tested locally) work the same way without us repeating the project connection: 
1. remove the `new_version.project_id = project.id` row and change `project` to `project.id` in the version:
    ```python
            # New version
            new_version = models.Version(
                size_stored=file_info.get("size_processed"),
                time_uploaded=new_timestamp,
                active_file=existing_file.id,
                project_id=project.id,
            )

            existing_file.versions.append(new_version)
    ```
2. Remove the `project_id=project` in the version definition and keep the `new_version.project_id = project.id`.
    ```python
            # New version
            new_version = models.Version(
                size_stored=file_info.get("size_processed"),
                time_uploaded=new_timestamp,
                active_file=existing_file.id,
            )

            # Set FK directly so we do not modify the project row (avoids UPDATE on projects
            # and reduces lock contention during concurrent PUT /file/new).
            new_version.project_id = project.id
            existing_file.versions.append(new_version)
    ```

i-oden (commented):
### A new review round

As I've said I think you've come up with a really good solution and it all looks very good, just some questions and notes, and potentially a few more tests to verify the behavior. 

#### Questions:
- Should the tests for the new endpoint be in the `test_files_new.py`? 

#### Edge cases:
- Project status changes during upload means the `date_updated` will not be updated -- this is already a potential issue during upload so nothing that really applies to this specific change. May need a solution at some point but not in this task. 

#### Potential additional tests:
- `ProjectUploadComplete` does not update if the project is not In Progress
- `ProjectUploadComplete` handling of db failures
- Missing `project` in query to `ProjectUploadComplete`
- `NewFile` `POST` no longer updates `project.date_updated` / `last_updated_by`
- `NewFile` `PUT` still creates a Version tied to both the file and the project

dds_web/api/files.py
valyo: Thanks @i-oden , all your comments are addressed now

tests/tests_v3/test_files_new.py
```diff
@@ -436,6 +465,44 @@ def test_new_file(client):
     assert f"File '{updated_file['name']}' updated in db." in response.json["message"]
 
 
+def test_new_file_put_version_linked_to_file_and_project(client):
```
i-oden, line 506: Maybe we should add the `date_updated == original_date_updated` + `last_updated_by == original_last_updated_by` to this test as well, same as for `POST`?

i-oden (commented):
Nice job! Heaps covered now. Just a couple comments below but only one minor suggestion and one thing that has already been kind of covered in previous comments
 
I'll test it out next week! 

> [!NOTE]
> Theoretical review, no testing this time

### Suggested fixes in this PR

- Potentially double check that the tests for POST and PUT check the same variables (see other comment) 
 
### Future tasks (not this PR)

This PR is big enough as it is, so I don't think we should introduce more changes unless absolutely necessary. 

- (This is related to the race comments made previously in this PR, e.g. by codex) Add uniqueness constraint in the database model. Atm technically two requests can pass validation at the same time and create duplications in the database. Probably not very likely, but could happen. Should be a uniqueness constraint for (project, file) and maybe some way to handle it if that does happen, that doesn't involve an error.

---

tests/tests_v3/test_files_new.py
valyo: Good point!
I added it as a separate test (test_new_file_put_does_not_update_project_timestamp) mirroring the POST one, rather than extending this test — keeps version-linkage and timestamp-preservation as separate concerns instead of one test checking both.

i-oden (approved):
I have now tested this out (this branch against `dds_cli/dev` + this branch against `dds_cli/improve-project-updating-during-upload`) after the changes and from what I can tell it all looks great and works completely as expected 🙌🏻 `dds data put` and `dds data put --overwrite` both **do not** update the `Project.date_updated` (etc) column with the `dds_cli/dev` branch, but all other updates, inserts and FKs work as before. The same commands from the matching branch updates the `Project.date_updated` after the uploads are completed. 

Good job!

valyo:
> I have now tested this out (this branch against `dds_cli/dev` + this branch against `dds_cli/improve-project-updating-during-upload`) after the changes and from what I can tell it all looks great and works completely as expected 🙌🏻 `dds data put` and `dds data put --overwrite` both **do not** update the `Project.date_updated` (etc) column with the `dds_cli/dev` branch, but all other updates, inserts and FKs work as before. The same commands from the matching branch updates the `Project.date_updated` after the uploads are completed.
> 
> Good job!

Thank you again for looking into this!

